PR #3, opened 2024-05-24
Add request and response structs for wallet API

sree-revoori1:


r-n-o:
@sree-revoori1 yep, I feel you. Here's a draft to address this: https://github.com/tkhq/rust-sdk/pull/4. Hoping having something like this would help? Disclaimer: I haven't tried to use the generated code yet. Just verified that the generation works!

sree-revoori1:
@r-n-o Thank you so much for the quick turnaround on this! This is super helpful, we will build off of the protos instead now.

PR #4, opened 2024-05-24
Add Turnkey API protos and generate code

r-n-o:
This PR is a draft for codegen in the Rust SDK: internally Turnkey uses proto3 to define all types and interfaces so it seems like a waste to force our users to manually write types. We can generate it all!

The `proto` folder now contains the source protos, and `src/gen` contains generated code. I haven't yet tried to use this with a Rust project hence why this is still a draft PR.


src/lib.rs
```diff
@@ -1,122 +1,2 @@
...
-        signature: hex::encode(sig.to_der()),
-        scheme: "SIGNATURE_SCHEME_TK_API_P256".to_string(),
-    };
-
-    let json_stamp = serde_json::to_string(&stamp).unwrap();
-
-    Ok(BASE64_URL_SAFE_NO_PAD.encode(json_stamp.as_bytes()))
-}
-
-#[cfg(test)]
```
andrewkmin, line 47: would be nice to retain some semblance of a test suite with the rewrite :)

src/client.rs
```diff
@@ -0,0 +1,267 @@
...
+                if let Ok(x_stamp_header) =
+                    HeaderValue::try_from(sealed_request_output.x_stamp.clone())
+                {
+                    headers.insert("X-Stamp", x_stamp_header);
+                }
+                match request_input.headers {
+                    Some(input_headers) => {
+                        for (_, value) in input_headers.iter() {
+                            if let Ok(val_header) = HeaderValue::try_from(value) {
+                                headers.insert("test keyname", val_header);
```
andrewkmin, line 211: do we need this in the case that there are other headers the user would like to include? or purely for testing?

src/client.rs
```diff
@@ -0,0 +1,267 @@
...
+        &self,
+        uri: String,
+        query: QueryShape,
+        substitution: SubstitutionShape,
+    ) -> TurnkeyResult<String> {
+        let path = Turnkey::substitute_path(uri, substitution)?;
+        // TODO: Might need to check if path starts with "/", not sure how its called
+        let mut url = format!("{}{}", self.base_url, path);
+
+        // TODO: This can all be cleaned up/simplified with Url crate if needed
```
andrewkmin, line 155: I'm not opposed to using a URL crate if it simplifies thing. any potential reason it'd be a bad idea?

andrewkmin (commented):
🏅 looking good!! just some minor comments to address

src/client.rs
andrewkmin: cc @joshhchun

mountainpath9:
If this is intended to be a general purpose SDK, I'd suggest that accessing environment variables directly (and the use of .env files) should be up to the SDK user, and not baked into the SDK itself. My personal use case is likely to see these values coming from other rust code.

If you really want this for convenience, then env variable access should be optional, ie something like this this: 

```
impl Turnkey {
    pub fn new(base_url: String, stamper: TurnkeyApiKey) -> Self {
        Self {
            base_url,
            stamper,
            client: Client::new(),
        })
    }

    pub fn new_from_env() -> TurnkeyResult<Self>  {
        Self::new(
           base_url: env::var("TURNKEY_BASE_URL")
                .map_err(|e| TurnkeyError::OtherError(e.to_string()))?,
            stamper: TurnkeyApiKey::new_from_env()?,
        )
    }
```

src/gen/external.activity.v1.rs
ruseinov, line 1: The first thing that stands out when looking at the this PR is that the generated code is committed to the repo. Is there a reason why? If not - I'd say that the approach with generating them via `buld.rs` into `OUT_DIR` along with `cargo build` might be a nice alternative. The files could still be inspected in the IDE, without polluting the history should `proto` files change. It also eliminates a manual generation step.

src/errors.rs
```diff
@@ -0,0 +1,36 @@
+use reqwest::Error as ReqwestError;
+use thiserror::Error;
+
+#[derive(Debug)]
+pub enum TurnkeyError {
+    MethodError(TurnkeyResponseError),
+    StampError(StampError),
+    HttpError(ReqwestError),
+    OtherError(String),
```
ruseinov, line 9: This could likely be replaced by `anyhow` to avoid the boilerplate related to string conversions. It would then also allow for specifying the context. `e.g. result.context("failed to do x")`.

Cargo.toml
```diff
@@ -11,3 +17,11 @@ p256 = { version = "0.13.2", default-features = false, features = ["ecdsa"] }
 serde = { version = "1.0.197", default-features = false, features = ["std", "derive"] }
 serde_json = { version = "1.0.115", default-features = false, features = ["std"] }
 thiserror = { version = "1.0.58", default-features = false }
+bytes = { version = "1.6.0", default-features = false }
+prost = { version = "0.12.6", default-features = false }
+prost-build = { version = "0.12.6", default-features = false, optional = true }
+reqwest = { version = "0.12.4", default-features = false, features = ["json"] }
+dotenv = "0.15.0"
```
ruseinov, line 24: https://github.com/allan2/dotenvy

src/gen/external.activity.v1.rs
r-n-o: Generated files being checked in is something we do for security reasons: it's convenient to see exactly what changes at PR time and avoids surprises. We basically treat it as first-party code (hence: should be reviewed and scrutinized just the same)

It comes with trade-offs: bigger diffs and manual generation step. But I think this makes sense given our release flow: Turnkey API releases a new version => a human has to update protos and make a PR, running a generation step.

Cargo.toml
r-n-o: Ah yes. Thank you for pointing this out!

src/gen/external.activity.v1.rs
ruseinov: Roger that. In theory this should not be needed if a version of the library you use to generate the code has been audited and locked. But I get where you are coming from, it's a bit of an overhead that comes with it's benefits.

r-n-o:
Closing this PR, it's now moot given #8 is merged! Onwards 🚀

PR #5, opened 2024-06-17
Implement a rust client for the turnkey API

mountainpath9:
This includes the 63 api endpoints in `public_api.proto`, though only a few have been tested. 

It builds atop the work already done in https://github.com/tkhq/rust-sdk/pull/4


frankli-dev:
`cargo run --example create_sub_organization hello`

```
thread 'main' panicked at examples/create_sub_organization.rs:51:63:
called `Result::unwrap()` on an `Err` value: OtherError("failed to parse response: error decoding response body")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Looks like the response type is changed.

PR #8, opened 2025-04-25
Rust SDK client: codegen and batteries included 🚀 

r-n-o:
## High-level

This pull request contains 2 new crates: `tkhq_codegen` and `tkhq_client` which contain, respectively:
* code to generate the repetitive portions of `tkhq_client`
* the actual client, to make requests to the Turnkey API

## Why this was hard

There are no strong Swagger/OpenAPI parsers, and I wanted to do _better_ than what our current SDKs do by having a type-safe interface for activities. The main issue is that the Turnkey API for activities is **polymorphic** and uses semantic level mapping (instead of structural mapping) to map an activity's result type to the intent that generated it.

More concretely, a request looks like this:
```
{
  "type": "ACTIVITY_TYPE_FOO",
  "timestamp": 1680000000,
  "parameters": { ...variant-specific payload... }
}
```
And the response looks like a **generic** activity response. You have to look at the original request (url, activity type) to know that the result should be of type "Foo" rather than "Bar".
```
{
  "activity" {
    "id": "some-id",
    "result": { ...variant-specific shape... }
  }
}
```

In other words, the design goals were
- Expose **type-safe request/response interfaces** (`FooParams` → `FooResult`)
- Avoid forcing consumers to manually match on enums or `Option` types
- Take care of polling automatically to avoid cases where the activity comes back PENDING

### New custom activity/url/intent/result mapping
To inject and force the mapping between activity type, intent type, and result type I had to manually specify these in a new `activities.json` file, used by codegen. This is admittedly not the prettiest solution, but it gets the job done. I'm also thinking that this file should be maintained in our monorepo and can be used to drive (for example) the golang logic which validates intent/result types (today this is manually done)

The big advantage: consumers do not have to guess the `result` type, and mismatches are simply _not possible_ anymore.

### Error handling

Settled on `thiserror` for convenience; the SDK client wraps the underlying errors such as:
- `reqwest::Error`
- encodingdecoding failures (serder)
- custom logic errors (e.g. missing result, activity needs consensus)

### Polling abstraction

Decided to bake this in and make polling optional, with a `RetryConfig`. The default retry config retries (duh), but `RetryConfig::none()` gives a "no-op" retry config if users want to opt out.

This uses configurable exponential backoff with jitter, max attempts, etc.

### Codegen...and rewrite

The `main.rs` file in `codegen` is ~500 LOC which is a lot more than expected, but I think it's worth it: I had to work through pretty tricky problems:
* `prost` generates "bad" enums which have `i32` instead of properly typed enums
* `prost` generates nested `Inner` structs/enums to express "oneof". This isn't compatible with serde / JSON serialization. Had to insert `#[serde(flatten)]` where necessary
* Had to insert "Default" where necessary to allow for clean serde parsing to go through when Vecs or Options aren't in the JSON
* had to write a custom proto parser to parse grpc annotations (which define routes and methods!) currently tonic/prost can't do this
* had to programmatically use `#[serde(rename = "...")]` to match the "SCREAMING_SNAKE_CASE" convention we have for enum values in JSON (e.g. "ACTIVITY_TYPE_FOO")

## Testing

Probably the weak point of this PR in its current state. The basic examples work, which is nice, but I need to do a lot more to test this new code thoroughly. Top of mind:
* RetryConfig behavior: test for correctness with unit tests
* Parsing: test for correctness and error handling with manually crafted payloads
* e2e: run the "examples" in CI, and maybe a few more activity types


codegen/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    // We manually have to specify the mapping between activity types, route, intent and result type through a file
+    // Unfortunately this information isn't available in proto directly because this mapping is semantic, not structural.
+    let activities_mapping_data =
+        fs::read_to_string(ACTIVITIES_MAPPING_PATH).expect("cannot read activities.json");
+    let parsed_activities: ActivitiesFile =
+        serde_json::from_str(&activities_mapping_data).expect("cannot parse activities.json");
+
+    for service_caps in service_re.captures_iter(&proto) {
+        let service_body = &service_caps[2];
```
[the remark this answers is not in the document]
zkharit: Kinda separate, but it took me a moment to understand that the whole input is the [0]th item of the array, I was thinking the captures groups were 1 indexed for some reason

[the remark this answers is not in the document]
zkharit: Just to clarify the first comment, I think itd be really helpful to breakdown what is captured at each step. Starts with services_caps[0]
```
service PublicApiService {
 rpc DeletePolicy(external.activity.v1.DeletePolicyRequest) returns (ActivityResponse) {
  option (google.api.http) = {
      post: "/public/v1/submit/delete_policy"
      body: "*"
    };
  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
      description: "Delete an existing Policy"
      summary: "Delete Policy"
      tags: "Policies"
    };
 }
 
 <rest of rpc methods>
}
```
Then goes to service_body[0]:
```
rpc DeletePolicy(external.activity.v1.DeletePolicyRequest) returns (ActivityResponse) {
  option (google.api.http) = {
      post: "/public/v1/submit/delete_policy"
      body: "*"
    };
  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
      description: "Delete an existing Policy"
      summary: "Delete Policy"
      tags: "Policies"
    };
}
```

And then youve already further broken it down which I really liked!

codegen/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+            //    };
+            let http_opts = &rpc_caps[4];
+            if http_opts
+                .contains("option (google.api.method_visibility).restriction = \"INTERNAL\"")
+            {
+                // Skip internal-only endpoints
+                continue;
+            }
+
+            let http_re = Regex::new(r#"post\s*:\s*\"([^\"]+)\""#).unwrap();
```
zkharit, line 102: A breakdown of this regex and its captures would be great too :)

codegen/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+    fs::write(generated_client_path, output).expect("Failed to write generated file");
+
+    // Now add client.rs to the generated/mod.rs file
+    let mut mod_rs = fs::OpenOptions::new()
+        .append(true)
+        .open(out_dir.join("mod.rs"))
+        .unwrap();
+
+    // This will append a new line to the end of the file
+    writeln!(mod_rs, "\n// Added by tkhq_codegen").unwrap();
```
zkharit, line 220: It doesnt seem like this comment is in the mod.rs file, and the order of the `pub use` statements below doesnt seem to be preserved either for some reason?

I tried running `make generate` and see the same output, not a blocker or anything but just a bit confused

codegen/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+    // ...and prost doesn't support JSON serialization...
+    // So here we are, doing some post-processing on the generated code, to manually switch enum fields from
+    // the generic "i32" type to their _actual_ type (e.g. super::super::common::v1::ApiKeyCurve)
+    // files that we need to add custom JSON logic to
+    for entry in WalkDir::new(GENERATED_CLIENT_DIR)
+        .into_iter()
+        .filter_map(Result::ok)
+        .filter(|e| {
+            e.path().extension().map_or(false, |ext| ext == "rs")
+                && !e.file_name().to_string_lossy().contains("google.api")
```
zkharit, line 240: There are two vendored google files: `google.api.rs` and `google.rpc.rs`, is it intentional that this statement is only excluding `google.api.rs`

zkharit: image

zkharit (commented):
This is super cool

zkharit:
Also a README refresh could be in order

codegen/src/activities.json
```diff
@@ -0,0 +1,269 @@
+{
```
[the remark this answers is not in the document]
r-n-o: It is possible but you have to manually construct the the body. I think it's nice to keep parity with docs and other SDKs, which are all broken down by URLs. The one-to-many mapping, while "real" from our server's point of view, isn't meant to be exported to users, who should see a 1-1 mapping. Fewer option, less confusion. If someone wants to use an old activity version they can either downgrade the SDK to a previous version, or maybe do this manual trick. I'm hesistant to document it and "promote" it because this 1-to-many mapping is a tool for us to offer backwards compatibility, rather than something users should be exposed to if they're looking at a single SDK version.

Actually, maybe a good place to document this would be in the docstring of the client's "process_activity". Hidden enough, but documented :)

codegen/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+        .field_attribute("google.rpc.Status.details", "#[serde(skip)]")
+        .type_attribute("google.rpc", SERDE_CAMEL_CASE)
+        .compile_protos(&[PUBLIC_API_PROTO_PATH], &[INCLUDE_PROTO_PATH])
+        .unwrap();
+
+    let proto = fs::read_to_string(PUBLIC_API_PROTO_PATH).expect("Failed to read proto file");
+
+    // Capture the start of "service... {" until a single "}" is encountered on its own line without indentation.
+    // That's just a simple alternative to writing a nesting-aware parser...
+    let service_re = Regex::new(r"(?ms)^service\s+(\w+)\s*\{\n(.*?)^\}").unwrap();
```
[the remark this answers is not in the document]
r-n-o: Added!

codegen/src/activities.json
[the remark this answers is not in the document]
r-n-o: Done!

codegen/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    let rpc_re = Regex::new(
```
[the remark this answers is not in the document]
r-n-o: Yep, good call. Added.

codegen/src/activities.json
```diff
@@ -0,0 +1,269 @@
...
+    "activities": {
```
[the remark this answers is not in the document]
r-n-o: I think this should be coming from mono actually. My plan is to check it into mono, and `make sync` will simply copy over the protos and this file. No manual shenanigans hopefully!

client/src/retry.rs
```diff
@@ -0,0 +1,122 @@
...
+            max_delay: Duration::from_secs(0),
+            max_retries: 0,
+        }
+    }
+
+    /// Computes the delay to wait after N attempts have been made.
+    /// `attempt_count` starts at 1 because it never makes sense to wait before the first try.
+    pub fn compute_delay(&self, attempt_count: usize) -> Duration {
+        if attempt_count == 0 {
+            panic!("attempt_count cannot be 0")
```
zkharit, line 45: Do we really want to panic!() here? Maybe this can return a Result/Option and be caught by the caller

client/src/retry.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Maximum number of retries.
+    pub max_retries: usize,
+}
+
+impl Default for RetryConfig {
+    fn default() -> Self {
+        RetryConfig {
+            initial_delay: Duration::from_millis(500),
+            multiplier: 2.0,
+            max_delay: Duration::from_secs(10),
```
zkharit, line 22: nit: the default config will never reach max_delay

client/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+
+    #[error("Maximum number of attempts reached (after {0} retries)")]
+    ExceededRetries(usize),
+}
+
+pub struct TurnkeyClient {
+    http: reqwest::Client,
+    base_url: String,
+    api_key: TurnkeyApiKey,
+    retry_config: RetryConfig,
```
zkharit, line 59: nit: this could be an Option, and use the default if not provided

examples/src/bin/wallet.rs
```diff
@@ -0,0 +1,113 @@
...
+        env::var("TURNKEY_API_PUBLIC_KEY").expect("cannot load TURNKEY_API_PUBLIC_KEY");
+    let api_private_key =
+        env::var("TURNKEY_API_PRIVATE_KEY").expect("cannot load TURNKEY_API_PRIVATE_KEY");
+
+    let api_key = TurnkeyApiKey {
+        private_key_hex: api_private_key,
+        public_key_hex: api_public_key.clone(),
+    };
+
+    // Create the request body for our create_sub_organization request
```
zkharit, line 40: nit: CreateWallet request

examples/src/bin/create_sub_organization.rs
```diff
@@ -39,73 +48,52 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-        },
-    })
-    .to_string();
 
-    // Sign the request with a stamp
-    let mut headers = HeaderMap::new();
-    let stamp = stamp(body.clone(), &api_key)?;
-    headers.insert("X-Stamp", HeaderValue::from_str(&stamp)?);
+    let client = tkhq_client::TurnkeyClient::new(TURNKEY_API_HOST, api_key, RetryConfig::none());
+    let intent = CreateSubOrganizationIntentV7 {
```
zkharit, line 53: So much cleaner, and safer!!

examples/src/bin/create_sub_organization.rs
```diff
@@ -39,73 +48,52 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-    // Send the request out!
-    let client = reqwest::Client::new();
-    println!("Sending request body:\n{}\n", body.clone());
-    println!("Associated X-Stamp header:\n{}\n", stamp.clone());
-    let response = client
-        .post(format!("{TURNKEY_API_HOST}{CREATE_SUB_ORGANIZATION_PATH}"))
-        .headers(headers)
-        .body(body)
-        .send()
+    let res = client
```
zkharit, line 86: 😮‍💨

examples/src/bin/whoami.rs
```diff
@@ -1,18 +1,23 @@
...
 // See <https://docs.turnkey.com/api-reference/sessions/who-am-i> for documentation
 const TURNKEY_API_HOST: &str = "https://api.turnkey.com";
-const WHOAMI_PATH: &str = "/public/v1/query/whoami";
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
     // Load .env file from the example folder (examples/.env)
     let current_dir = env::current_dir()?; // should be the workspace root
-    dotenv::from_path(current_dir.join("examples").join(".env"))?;
+    let env_path = current_dir.join("examples").join(".env");
```
zkharit, line 13: nit: Could maybe create a function to avoid the repetition for further examples of getting the env file

zkharit:
Improvement suggestion: GenerateApiKey helper function (not the CreateApiKey activity). Would be of great use here: https://github.com/tkhq/rust-sdk/blob/f33439e872ef02b830db563782aa2d3b85e90ae3/examples/src/bin/create_sub_organization.rs#L39-L40

codegen/src/main.rs
r-n-o: Ah yes. I'm not sure why I kept this processed. I don't need it processed! Fixed.

codegen/src/main.rs
r-n-o: Yeah this is because I'm post-processing mod.rs, which strips comments. I've fixed this now.

client/src/retry.rs
r-n-o: I think we want to `panic` here because this is an incorrect use of the API. I debated adding a Result but I think it's better if the API stays super simple.

client/src/retry.rs
r-n-o: Good catch! I'll modify this.

client/src/lib.rs
r-n-o: Great idea!

examples/src/bin/whoami.rs
r-n-o: Yep. I think I'll abstract this out. Good idea.

README.md
```diff
@@ -1,31 +1,52 @@
...
 To make a request to Turnkey:
-* use any HTTP library ([`reqwest`](https://docs.rs/reqwest/latest/reqwest/) is a popular choice)
-* look at our [API reference](https://docs.turnkey.com/api-reference/overview) to learn about the JSON schema of our requests and responses
-* create a POST request body (all Turnkey API requests are POST requests). You can use [`serde_json`](https://docs.rs/serde_json/latest/serde_json/) to help with JSON serialization.
-* use the the provided `stamp` function to sign your request body with a `TurnkeyApiKey`
-* add the stamp value (bytes) inside of a new `X-Stamp` header, and attach it to the request
-* POST the request, read the response, and use the activity result
+* Load an API key:
+  ```rust
+  let api_key
```
zkharit, line 14: Did you mean to include more here? Or is this intentionally blank until the gen/import api key function

README.md
```diff
@@ -1,31 +1,52 @@
...
 ## Examples
 
-For fully working examples, see our ['examples/' folder](./examples/).
+For fully working examples, see our ['examples/' folder](./examples/README.md).
+
+## Development
+
+This project uses `make` to encapsulate common tasks:
+* `make lint` will run rust fmt and clippy
+* `make test` runs the unit tests
```
zkharit, line 42: nit: A note about neeeding valid .env's for tests could be good

zkharit (approved):
🚀

README.md
r-n-o: errrr...yes 😅

README.md
r-n-o: tests do not require a valid env, but examples do. Good call!

PR #11, opened 2025-05-02
New crate: `enclave_encrypt` to interact with Auth, Export, and Import bundles

### LINKED ISSUE #9: Primitives to encrypt and decrypt bundles (auth, export, import)
Internally at Turnkey we have a crate to encrypt and decrypt bundles using HPKE ([RFC 9180](https://datatracker.ietf.org/doc/rfc9180/)).

There's been interest (@frankli-dev & @sevazhidkov) to access this crate so that auth or export bundles can be decrypted. Right now this is very hard to do for external users because the standard we follow, while an open standard, has some configuration choices which aren't obvious.

Need to get this committed to the Rust SDK repo. I'm envisioning a new `tkhq_enclave_encrypt` crate which can be published and used.

r-n-o:
## What is this change about?
This PR introduces a new `enclave_encrypt` crate to facilitate interaction with Auth, Export, and Import bundles. This resolves #9.

This crate was not written from scratch. Most of it was written (and is currently in use) in Turnkey's monorepo. The commit where the code is ported over is the first one, https://github.com/tkhq/rust-sdk/commit/d4538c54d5eade1405463da2986a799315c937cd.

The new abstractions which are more user-friendly (`AuthenticationClient`, `ExportClient`, `ImportClient`) were introduced in https://github.com/tkhq/rust-sdk/commit/4266a5a6f4b9c1bacb2286cf50a9625ed1dd3550.

## Testing
- [x] added a lot of unit tests to `enclave_encrypt`. See https://github.com/tkhq/rust-sdk/commit/4266a5a6f4b9c1bacb2286cf50a9625ed1dd3550 for the modifications I made on top of the version we've always had internally
- [x] added an example to the `examples` folder to make sure it works end-to-end

### DIFF
--- a/enclave_encrypt/README.md
+++ b/enclave_encrypt/README.md
@@ -1,0 +1,101 @@
+# tkhq_enclave_encrypt
+
+This repository contains primitives to encrypt and decrypt data, sent to and from Turnkey secure enclaves (see Enclave to [end-user secure channels](https://docs.turnkey.com/security/enclave-secure-channels)).
+
+Encryption and decryption are "one-shot" using the HPKE standard ([RFC 9180](https://datatracker.ietf.org/doc/rfc9180/)).  Neither the client or the server should ever be reused to send/receive more than one message. We want to avoid the recipient target key being used more then once in order to improve forward secrecy; see [security profile](#security-profile) section for important details and caveats.
+
+The flows where encryption and decryption are relevant are:
+* Authentication: [Email](https://docs.turnkey.com/authentication/email), [SMS](https://docs.turnkey.com/authentication/sms), or [Social](https://docs.turnkey.com/authentication/social-logins) logins return encrypted authentication bundles produced by Turnkey secure enclaves
+* Key or Wallet export: key material (private key or wallet mnemonic phase) are encrypted by Turnkey enclaves to end-user public keys. See [Export Wallets and Keys](https://docs.turnkey.com/wallets/export-wallets#export-wallets-and-keys).
+* Key or Wallet import: Turnkey enclaves send a bundle containing a signed public key, to which end-users encrypt their key material (private key or wallet mnemonic phrase). See [Import Wallets and Keys](https://docs.turnkey.com/wallets/import-wallets).
+
+## Usage
+
+Decrypting a bundle:
+
+```rust
+let client_ikm = hex::decode("...private bytes from client...").expect("cannot decode client secret bytes");
+let (client_private_key, client_public_key) = Kem::derive_keypair(&client_ikm);
+
+// Create a new `EnclaveEncryptClient` bound to the client's private/public key
+let mut client = EnclaveEncryptClient::from_enclave_auth_key_and_target_key(
+    // required to construct `EnclaveEncryptClient`, but unused when decrypting auth bundles.
+    *SigningKey::random(&mut OsRng).verifying_key(), 
+    // client public key (to which the bundle is encrypted)
+    client_public_key,
+    // client private key, used for decryption of the auth bundle
+    client_private_key,
+);
+
+let bundle = "<auth bundle goes here, it's a base58-encoded string>";
+let decrypted = client.auth_decrypt(bundle)
+```
+
+## Running test
+
+```
+cargo test
+```
+
+## HPKE Protocol Details
+
+### Terms
+
+- Encapsulated ("Encapped") Key - the public key of the sender used for ECDH.
+- Target Key Pair - the key pair of the receiver that the sender encrypts to the public key of. Only one message should ever be encrypted to the public key.
+- Server - a server inside of the enclave; normally an enclave application.
+- Client - a client outside of the enclave; normally a turnkey end user.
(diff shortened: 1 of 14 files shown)


enclave_encrypt/README.md
```diff
@@ -11,24 +11,64 @@ The flows where encryption and decryption are relevant are:
...
+
+let bundle = "<auth bundle>";
+
+let client.decrypt(bundle).expect("decryption should succeed");
+```
+
+If you persist client key material between initiation and decryption you may use `dangerous_from_bytes` to create an `AuthenticationClient`. Keep in mind the one-shot encryption semantics, you may not use the same client IKM to decrypt many different bundles.
 
+```rust
+let client_ikm = hex::decode("...private bytes from client...").expect("cannot decode client secret bytes");
```
omkarshanbhag, line 28: nit: the value uses "private bytes" but the error statement uses "secret bytes". Could we keep these consistent to avoid confusion (unless they are indeed different things.

enclave_encrypt/src/client.rs
```diff
@@ -1,17 +1,285 @@
...
+    }
+}
+
+/// Abstraction over `EnclaveEncryptClient` for private key or wallet import flows.
+pub struct ImportClient {
+    encrypt_client: EnclaveEncryptClient,
+}
+
+impl ImportClient {
+    /// Creates a new Export client. Takes in a Quorum public key (use `QuorumPublicKey::production_signer()`)
```
omkarshanbhag, line 175: Export --> Import

enclave_encrypt/src/errors.rs
```diff
@@ -1,71 +1,102 @@
...
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[allow(missing_docs)]
+#[derive(Debug, Error, Clone, PartialEq, Eq)]
 pub enum EnclaveEncryptError {
     /// Receiver or Encap key is likely wrong.
+    #[error("Failed to set up receiver context")]
     ReceiveCtxSetupFail,
+
     /// Something is likely wrong with the receiver public key.
+    #[error("Failed to set up receiver context")]
```
omkarshanbhag, line 13: should this be "Failed to setup send context" instead of receiver context?

enclave_encrypt/src/errors.rs
```diff
@@ -1,71 +1,102 @@
...
+
+    #[error("Error while deserializing the server encapsulated key")]
     InvalidSeverEncappedKeySignature,
-    /// Could not deserialize signature over server target key.
+
+    #[error("Error while deserializing signature over the server target key")]
     InvalidServerTargetSignature,
-    /// Could not use secret as signing key.
+
+    #[error("Count not use quorum secret as a valid signing key")]
```
omkarshanbhag, line 49: Count --> Could

enclave_encrypt/src/errors.rs
```diff
@@ -1,71 +1,102 @@
...
+
+    #[error("Could not deserialize P-256 public key: invalid SEC1 encoding")]
     InvalidP256PublicKeySec1Encoding(String),
-    /// Failed to decode a base58-encoded string.
+
+    #[error("Invalid base58 encoding")]
     FailedToBase58Decode(String),
-    /// Email recovery payload was shorter then expected.
+
+    #[error("Email recovery payload is shorter than")]
```
omkarshanbhag, line 70: missing the word "expected"

enclave_encrypt/src/errors.rs
```diff
@@ -1,71 +1,102 @@
...
+
+    #[error("Provided public key bytes are not sized correctly for a Quorum public key")]
     IncorrectQuorumPublicKeyBytesLength(usize),
-    /// Error while decoding hex bytes
+
+    #[error("Error while decoding hex-encoded string: {0}")]
     HexDecode(String),
-    /// Invalid bytes were used to create a `VerifyingKey`
+
+    #[error("Cannot create a verifying key from invalid")]
```
omkarshanbhag, line 91: invalid ... bytes?

examples/src/bin/wallet.rs
```diff
@@ -89,3 +117,10 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
     Ok(())
 }
+
+// Simple convenience function to display the first and last word of a mnemonic phrase
+// Technically we could just print the whole mnemonic out...but something about that feels wrong. I'm coy.
```
omkarshanbhag, line 122: ❤️

omkarshanbhag (approved):
Approved with minor comments/nits

enclave_encrypt/src/errors.rs
r-n-o: Yep! Good catch, the error message is off! But the comment is accurate. The sender context failing to be initialized is most likely because the receiver pubkey is invalid.

enclave_encrypt/src/errors.rs
r-n-o: ah yep! Again, good catch.

enclave_encrypt/src/errors.rs
r-n-o: 💯

omkarshanbhag (approved):
💯

Issue #12, opened 2025-05-02
Improve error handling for SDK functions


r-n-o:
When the API returns an error that happens to have a valid JSON shape (pretty common?) parsing this as an activity works. And it means the SDK returns a puzzling error:
```
Missing activity from response
```

Let's improve this and
1. Check for status (should be 200!)
2. If the activity is missing from the response, echo the underlying response text in the error surfaced to users so they can debug the problem themselves.

r-n-o:
Done with #13!

PR #13, opened 2025-05-06
Handle HTTP errors, and use Rust generics to DRY codegen'd client

r-n-o:
## Goals & motivation

The primary motivation for this PR is to address #12, where a 401 was surfaced as a `MissingActivity` error. While doing this I realized I'd have to add this new status-checking logic to 3 places, so I tried to DRY up the request logic in a single place: `process_request`, used by generated query client methods, and `process_activity`, which is used by generated activity client methods. Result? 300 lines of code gone 🎉 
image

Added a test to show the new status error check. It should now be much nicer for debugging: you get a proper `TurnkeyClientError::UnexpectedHttpStatus`, with the offending status code and associated HTTP response body!

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -2,7 +2,9 @@
 //! See <https://docs.turnkey.com>
 use serde::de::DeserializeOwned;
 use serde::Serialize;
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use reqwest::header::CONTENT_TYPE;
 use thiserror::Error;
 
 use generated::google::rpc::Status;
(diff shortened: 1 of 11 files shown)


client/src/lib.rs
```diff
@@ -153,31 +150,76 @@ impl TurnkeyClient {
...
+        let res = self
+            .http
+            .post(url)
+            .header("X-Stamp", stamp)
+            .body(post_body)
+            .send()
+            .await?;
+
+        let status = res.status();
+        let text = res.text().await?;
```
james-callahan, line 189: Is there a max size or timeout on the response size?

client/src/lib.rs
```diff
@@ -153,31 +150,76 @@ impl TurnkeyClient {
...
+
+        if !status.is_success() {
```
james-callahan, line 191: Shouldn't we look at the Content-Type to decide if it's json or not?
Just because something parses as valid json doesn't meant it's *meant* be json

client/src/lib.rs
r-n-o: Oh yeah, great point. I'm not specifying anything. So if there is a max size or timeout, it'll be the `reqwest` defaults.

client/src/lib.rs
james-callahan: > Our API should all be JSON. So if the content-type isn't JSON I can safely error out (with a proper error type).

It's also entirely possible that our reverse proxy (traefik) in certain situations responds with json document error messages of its own (e.g. if a service is internally unreachable)

r-n-o: Done, I think the handling is a lot more robust now! Also, added a test to make sure we fail with the right error.

client/src/lib.rs
r-n-o: This led me down a rabbit-hole: which settings should we support "first-party" and which settings should we not? In order to allow people to override arbitrary things in the reqwest client I've opted for a builder pattern which gives access to the reqwest:ClientBuilder object;

One thing that's really neat is that enabling debug logging for the unerlying reqwest client becomes possible out of the box! Dropped a note about this in the README (which I moved from the top-level crate)

client/src/lib.rs
```diff
@@ -153,50 +256,118 @@ impl TurnkeyClient {
...
+            return Err(TurnkeyClientError::UnexpectedHttpStatus(
+                status.as_u16(),
+                text,
+            ));
+        }
+
+        if content_type != mime::APPLICATION_JSON {
+            return Err(TurnkeyClientError::UnexpectedMimeType(
+                content_type.to_string(),
+            ));
```
james-callahan, line 315: You may want to log e.g. the first 200 bytes of the response as debug level logging?
It can be useful for weirder connectivity problems e.g. the user is behind a HTTP(S) proxy; or on captive portal wifi  that starts returning "you must be logged in" or something.

james-callahan (approved):

client/src/lib.rs
r-n-o: Good idea! I've logged this in https://github.com/tkhq/rust-sdk/issues/14 -- will spin up a separate branch/PR for it.

PR #18, opened 2025-05-09
Pre-publish cleanup

r-n-o:
This PR makes the crates that we want to publish ready to publish! Lots of little doc tweaks and little things here and there that was flagged by Rust's doctest and linter.


client/README.md
```diff
@@ -5,34 +5,43 @@ This crate contains an HTTP client to interact with the Turnkey API ([documentat
...
+// Create a new client:
+let client = turnkey_client::TurnkeyClient::builder().api_key(api_key).build().expect("client builder failed");
+
+// Make a request (for example, a signature request)
+let request = client.sign_raw_payload(
+    "your-turnkey-organization-id".to_string(),
     client.current_timestamp(),
     SignRawPayloadIntentV2 {
-        sign_with: address, // any Turnkey-generated address
+        sign_with: "0x123456".to_string(), // any Turnkey-generated address
```
andrewkmin, line 26: I believe these specific signing params are only valid for ecdsa signatures (eddsa requires `HASH_FUNCTION_NOT_APPLICABLE`)

proofs/Cargo.toml
```diff
@@ -1,27 +1,33 @@
 [package]
 name = "turnkey_proofs"
 version = "0.1.0"
-authors = ["tkhq"]
 edition = "2021"
-
+description = "Ttilities to parse and verify Turnkey secure enclave proofs"
```
andrewkmin, line 5: tiny nit: `Utilities`

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.81"
+channel = "1.85"
```
andrewkmin, line 2: curious about this — necessitated by the tooling? or just upgraded because newer == better? 😄

andrewkmin (approved):

rust-toolchain.toml
r-n-o: This was because of `cargo-release` actually. It required a newer Rust version. But yes, also: newer is better! :D

andrewkmin (approved):

PR #19, opened 2025-05-10
Update keywords to max 5 per crate

r-n-o:
```
error: failed to publish to registry at https://crates.io

Caused by:
  the remote server responded with an error (status 400 Bad Request): expected at most 5 keywords per crate
```

😭


andrewkmin (approved):
TIL max keywords was even a thing!

PR #30, opened 2025-07-18
Remove now-redundant lockfile for the proofs subcrate

cr-tk:
The commit caccc0612c7a0ca6527caafd38be086203af63d7 added the proofs crate to the workspace, moving lockfile ownership there instead.

Remove `proofs/Cargo.lock` to avoid confusion and dependabot false positives.


r-n-o (approved):
Thanks for catching this!

PR #32, opened 2025-07-21
Chore: update dependencies to already-reviewed versions

cr-tk:
Update some Rust dependencies to newer versions that we already trust in terms of security from reviews in https://github.com/tkhq/qos/ or other internal repositories.

The version bumps don't fix any known security issues.

I performed local builds and tests, but can't rule out that there are regressions that only show up with actual SDK usage. Some more testing could be useful before a future release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,7 +192,7 @@ dependencies = [
  "serde_bytes",
  "serde_cbor",
  "serde_repr",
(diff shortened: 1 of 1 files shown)


r-n-o (approved):
This looks good to me, thank you @cr-tk!

(and sorry for the delay, this got lost in my inbox)

PR #41, opened 2025-08-11
implement secp256k1 stamping

turnekybc:


client/src/lib.rs
```diff
@@ -175,18 +182,31 @@ impl Default for TurnkeyClientBuilder {
 }
 
 /// Base client. To create a new client, see `TurnkeyClient::builder`.
-#[derive(Debug)]
 pub struct TurnkeyClient {
     http: reqwest::Client,
     base_url: String,
-    api_key: TurnkeyP256ApiKey,
+    api_key: Box<dyn ApiKeyStamper>,
```
ruseinov, line 188: This could be e.g. a generic parameter instead of a box. or `impl ApiKeyStamper`

api_key_stamper/src/lib.rs
```diff
@@ -29,10 +38,22 @@ pub enum StamperError {
     HexDecode(String),
 }
 
+/// A stamped header to attach to an HTTP request.
```
r-n-o, line 41: "stamp header"

api_key_stamper/src/lib.rs
```diff
@@ -29,10 +38,22 @@ pub enum StamperError {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StampHeader {
+    pub name: &'static str,
```
r-n-o, line 44: why `&'static str` instead of `String`?

api_key_stamper/src/lib.rs
```diff
@@ -147,9 +168,116 @@ impl TurnkeyP256ApiKey {
...
+    pub fn stamp<S: AsRef<str>>(&self, request_body: S) -> Result<String, StamperError> {
+        let sig: K256Signature = self.signing_key.sign(request_body.as_ref().as_bytes());
+        let stamp = TurnkeyApiStamp {
+            public_key: hex::encode(self.compressed_public_key()),
+            signature: hex::encode(sig.to_der()),
+            scheme: SIGNATURE_SCHEME_SECP256K1.to_string(),
+        };
+        let json_stamp = serde_json::to_string(&stamp).unwrap();
+        Ok(BASE64_URL_SAFE_NO_PAD.encode(json_stamp.as_bytes()))
+    }
```
r-n-o, line 259: This should be removed now that you have the trait implementation

api_key_stamper/src/lib.rs
```diff
@@ -147,9 +168,116 @@ impl TurnkeyP256ApiKey {
...
+            name: API_KEY_STAMP_HEADER_NAME,
+            value: BASE64_URL_SAFE_NO_PAD.encode(json_stamp.as_bytes()),
+        })
+    }
+}
+
 #[cfg(test)]
 mod tests {
-    use p256::ecdsa::signature::Verifier;
+    use p256::ecdsa::signature::Verifier as _;
```
r-n-o, line 280: curious about why `as _` is needed here 🤔

r-n-o: Oh. Because it conflicts. Let's do `as P256Verifier`

api_key_stamper/src/lib.rs
```diff
@@ -196,10 +324,30 @@ mod tests {
...
             hex::decode(sig_hex_string).expect("signature should contain valid hex-encoded bytes");
-        let sig = Signature::from_der(&sig_bytes).expect("signature bytes should be valid DER");
+        let sig = P256Signature::from_der(&sig_bytes).expect("signature bytes should be valid DER");
         assert!(verifying_key.verify(b"Hi from TKHQ", &sig).is_ok());
     }
 
+    #[test]
+    fn test_secp256k1_generate_and_stamp() {
+        use k256::ecdsa::signature::Verifier as _;
+        use serde_json::Value;
```
r-n-o, line 334: hoist these with the other test imports

api_key_stamper/src/lib.rs
```diff
@@ -147,9 +168,116 @@ impl TurnkeyP256ApiKey {
...
+    }
+
+    /// Returns compressed SEC1 public key bytes
+    pub fn compressed_public_key(&self) -> Vec<u8> {
+        self.signing_key
+            .verifying_key()
+            .to_encoded_point(true)
+            .as_bytes()
+            .to_vec()
+    }
```
r-n-o, line 247: This is all copy/pasted from the P256 key implementation; I think it would be good to either consolidate with traits, or copy (or augment) tests so we have coverage for both p256 and k256. Your choice :D

api_key_stamper/src/lib.rs
```diff
@@ -29,10 +38,22 @@ pub enum StamperError {
...
 
+/// A stamped header to attach to an HTTP request.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StampHeader {
+    pub name: &'static str,
+    pub value: String,
+}
+
+/// Generic stamper that operates over raw bytes and returns an HTTP header.
+pub trait Stamp: Send + Sync {
```
r-n-o, line 49: Why `Send + Sync`? Feels a bit unnecessary (see these docs: https://doc.rust-lang.org/nomicon/send-and-sync.html)

client/src/lib.rs
```diff
@@ -83,17 +83,24 @@ pub enum TurnkeyClientError {
     StamperError(#[from] StamperError),
 }
 
-/// Builder for [`TurnkeyClient`].
-#[derive(Debug)]
-pub struct TurnkeyClientBuilder {
-    api_key: Option<TurnkeyP256ApiKey>,
+/// Builder for [`TurnkeyClient<S>`].
+pub struct TurnkeyClientBuilder<S: Stamp> {
+    api_key: Option<S>,
```
r-n-o, line 88: I don't really understand the need for this top-level generic. `api_key: Option<impl Stamp>`? Is the compiler giving you a hard time?

codegen/src/client.template.rs
```diff
@@ -9,6 +9,6 @@ use crate::generated::immutable::activity::v1 as immutable_activity;
 // This client is autogenerated. Please do not edit it manually
 // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-impl TurnkeyClient {
+impl<S: Stamp> TurnkeyClient<S> {
```
r-n-o, line 12: Not sure that's required; would love to find a way to keep to top-level client untouched; the only function which needs to know about this trait is the signing helper in lib.rs I believe. But maybe there's a reason why the compiler needs this?

client/src/lib.rs
turnekybc: Yes it's complaining:

```
`impl Trait` is not allowed in field types
`impl Trait` is only allowed in arguments and return types of functions and methods
```

ruseinov (approved):

codegen/src/client.template.rs
turnekybc: Ultimately this is required. I discussed with @r-n-o and we made the decision to keep this so that we can leverage generics.

api_key_stamper/src/lib.rs
turnekybc: Removed both stamp functions and am leveraging the trait.

api_key_stamper/src/lib.rs
turnekybc: Fixed the spelling here

api_key_stamper/README.md
```diff
@@ -44,10 +43,10 @@ let api_key = TurnkeyP256ApiKey::from_files(
 The API is straightforward, once you have a handle on an API key, call `stamp`:
 
 ```rust
-use turnkey_api_key_stamper::TurnkeyP256ApiKey;
+use turnkey_api_key_stamper::{TurnkeyP256ApiKey, Stamp, API_KEY_STAMP_HEADER_NAME};
```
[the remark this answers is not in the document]
turnekybc: I only needed to add `Stamp`. I removed the `API_KEY_STAMP_HEADER_NAME`.

api_key_stamper/src/lib.rs
```diff
@@ -196,8 +306,144 @@ mod tests {
...
+    #[test]
+    fn test_secp256k1_compressed_pub_len_is_33() {
+        let k = TurnkeySecp256k1ApiKey::generate();
+        assert_eq!(k.compressed_public_key().len(), 33);
+    }
+
+    #[test]
+    fn test_secp256k1_trait_stamp_header_round_trip_and_verify() {
+        use k256::ecdsa::signature::Verifier as _;
+        use k256::ecdsa::{Signature as K256Signature, VerifyingKey as K256VerifyingKey};
```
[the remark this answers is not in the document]
turnekybc: Hoisted!

client/src/lib.rs
```diff
@@ -671,6 +704,96 @@ mod test {
...
+            .await;
+
+        match result.unwrap_err() {
+            TurnkeyClientError::UnexpectedHttpStatus(status, body) => {
+                assert_eq!(status, 401);
+                assert_eq!(body, "{\"code\":2,\"details\":[],\"message\":\"some error\",\"turnkeyErrorCode\":\"SOME_ERROR_CODE\"}");
+            }
+            other => panic!("unexpected error: {:?}", other),
+        }
+    }
```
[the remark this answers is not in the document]
turnekybc: 😢 That's fair. I cleaned these up!

examples/src/bin/sub_organization_secp256k1.rs
```diff
@@ -0,0 +1,97 @@
...
+use turnkey_examples::load_api_key_from_env;
+
+use turnkey_client::TurnkeySecp256k1ApiKey;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let api_key = load_api_key_from_env()?;
+
+    // Sub-org root user gets a SECP256K1 API key:
+    let sub_organization_api_key = TurnkeySecp256k1ApiKey::generate();
```
r-n-o, line 20: Let's consolidate this in the `sub_organization` example, and create 2 API keys for the root user? One secp256k1, one p256?

As-is this example won't run on CI -- we can add this to this list of course:
https://github.com/tkhq/rust-sdk/blob/c08b31f3e23a06fab7a28b49a457d6b456565f95/Makefile#L40-L45

...but I'd rather consolidate and over time have a "maximal" sort of example demonstrating/testing the different features in one go.

r-n-o (approved):
👏

PR #42, opened 2025-08-12
Fix `u64` parsing in codegen with `serde_with::serde_as`

r-n-o:
This was a fun one to chase down: big ints which exceed the normal JSON integer range are serialized by GRPC gateway as strings. This makes sense looking back (https://github.com/grpc-ecosystem/grpc-gateway/issues/438#issuecomment-330739676), but it was very confusing at first when a user reported seeing:
```
Failed to decode response (invalid type: string "900", expected u64 at line 1 column 1750)
```

This fix is a bit more comprehensive than it needs to be but I think it's worth it: it already caught `session_expiry` which I wasn't aware of; that's fixed too! And going forward if we add more big ints to our protos in more places, codegen will automatically do the right thing.

The spirit of the fix is to let serde handle it; more specifically: [`serde_with`](https://docs.rs/serde_with/3.14.0/serde_with/index.html). Whenever the codegen spots a field that's a `big int` type or an `Option<big int>` type, we add the right type attribute, and we have to add an attribute on the struct itself to make this work. A bit painful but not too bad: the result can be seen with the minimal edits to the generated code.

### DIFF
--- a/codegen/src/transform.rs
+++ b/codegen/src/transform.rs
@@ -276,9 +276,9 @@ pub fn add_serde_as_for_large_int(field: &mut Field) -> bool {
                     }
                 }
             }
-            return false;
+            false
         }
-        _ => return false,
+        _ => false,
     }
 }
 


codegen/src/transform.rs
```diff
@@ -218,6 +235,53 @@ fn strip_prost_derive_from_attr(attr: &syn::Attribute) -> Option<proc_macro2::To
     })
 }
 
+// This modifies fields which are u64, i64, u128, i128 (or their option variants: Option<u64>, etc)
+// and uses `serde_with::serde_as` so these big ints can be parsed from JSON strings.
+// Returns a boolean to indicate whether the field was mutated or not.
+pub fn add_serde_as_for_large_int(field: &mut Field) -> bool {
```
andrewkmin, line 241: noice!

client/src/lib.rs
```diff
@@ -366,6 +367,34 @@ mod test {
         ));
     }
 
+    #[test]
+    fn api_key_serialization_and_parsing() {
```
andrewkmin, line 371: side note: this highlights some inconsistencies on our side: some activity intents expect expiration times as strings (e.g. `expiration_seconds`), whereas others expect uint64 (`session_expiry`) or even int32s (`otp_expiration_seconds, verification_token_expiration_seconds, session_expiration_seconds`). irrelevant to this change, just an observation as I was digging into some protos. 

and then on the flip side (Turnkey responses), org data stores fields like `expiration_seconds` as `uint64` for structs like API keys (which you're addressing here). 

just so I'm wrapping my head around this change as a whole correctly: we're modifying the codegen only for what parses Turnkey **responses**

andrewkmin (approved):
🏅

PR #64, opened 2025-11-13
Implement support for submitting presigned requests

luca992:
Adds `process_activity_with_stamp` & `process_request_with_stamp` apis to `TurnkeyClient` that submit requests using a provided stamp.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -235,10 +235,62 @@ impl<S: Stamp> TurnkeyClient<S> {
         request: Request,
         path: String,
     ) -> Result<Activity, TurnkeyClientError> {
+        let post_body = serde_json::to_string(&request)?;
+        self.process_activity_internal(post_body, path, None).await
+    }
+
+    /// POSTs an activity with a provided `stamp` and polls until the status is "COMPLETE"
+    ///
+    /// `process_activity` accepts a  arbitrary `Request` and `path` to POST to the Turnkey API.
+    /// It encapsulates the polling logic and is generally meant to be called by other
+    /// activity-specific client functions (e.g. `create_sub_organization`).
+    ///
+    /// Given the Turnkey API is backwards-compatible, this function can be used to submit old versions of activities.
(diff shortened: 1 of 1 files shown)


r-n-o (commented):
Thanks for the PR @luca992! Can you say more about why posting with a pre-stamped request is beneficial? The reason I'm asking is we might want to consider another interface potentially. Educated guesses:
* Right now the request signing is hardcoded to be done by the `TurnkeyClient`'s API key but maybe you envision signing being done by something else on the server? (different API key? Different curve?). If that's the case we already have the `Stamp` trait which you can implement. Would be good to know if it's something you tried to do or not (but I might be off because it might not be possible; all depends on your use case!)
* Maybe the issue is that you're not producing the requests or the stamp in the local process that posts them? (proxy?) If that's the case we might want to have a new struct (say `TurnkeyRequestProxy`) or a new set of methods with the "proxy" name somewhere to make it clear that it's the intended use-case. 

Another thought I'm having: we could have a `proxy_request` and `proxy_activity` method on the TurnkeyClient, and have these take a `Request` as input. The `post_body` parsing and validation can happen in a util function potentially? (advantage being: TurnkeyClient doesn't have to parse, interface is a bit more uniform)

r-n-o (commented):
Now that I understand the use-case a bit better I think it all makes sense! Thanks for taking the time to explain!

2 things to address:
* naming: `process_activity_with_stamp` => `proxy_activity` and `process_request_with_stamp` => `proxy_request` just to make it clear no signing is happening. I also like that these names contain "proxy" in them, to make the use-case clear. The "with_stamp" becomes implicit because it's required in the function params
* I'd like to see this used "for real" in our e2e tests. Mind writing a new example in our `examples` crate? I'm thinking a new `proxy.rs` example in https://github.com/tkhq/rust-sdk/tree/main/examples/src/bin

On that last point let me know if that's too much to take on: happy to work on that and add commits to your branch instead!

PR #117, opened 2026-04-07
add tvc image validation during deploys

annieke:
This PR adds
- a call to the validation endpoint when creating a new deploy
- creating a deploy with a hashlocked container URL, if the one provided wasn't hashlocked
- newly synced protos for this release
- some linting fix that wasn't already in


tvc/src/commands/deploy/create.rs
```diff
@@ -69,21 +110,32 @@ pub async fn run(args: Args) -> Result<()> {
...
+        .validate_tvc_image(validate_image_request)
+        .await
+        .context("failed to validate TVC image")?;
+
+    let pinned_image_url = pin_image_url(
+        &deploy_config.pivot_container_image_url,
+        &validate_image_response.resolved_image_digest,
+    );
+
+    if pinned_image_url != deploy_config.pivot_container_image_url {
```
emostov, line 130: what is the purpose of doing this extra work to ensure the image URL is hash pinned? Is it to help ensure the pivot hash resolves correctly?

emostov (approved):
🚢

tvc/src/commands/deploy/create.rs
annieke: TVC requires hashpinned image URLs (or deploys fail), so we're soft-enforcing it in the frontend and here

PR #118, opened 2026-04-08
feat(tvc): add pivot digest validation

annieke:
A TVC app will fail to deploy if the pivot digest does not match what's pulled from the container URL. We expect users to [computer their own digest via `docker create` and `sha256sum`](https://docs.turnkey.com/getting-started/verifiable-cloud-quickstart), and also expose it in the [helloworld template](https://github.com/tkhq/tvc-template/actions/runs/23811832929/attempts/1) - this PR adds the final piece to locally validate the pivot hash.

## Usage

Pivot hash validation can be individually invoked via `tvc deploy validate-pivot-digest`, but also through `--validate-pivot-digest` during `deploy create` and `deploy approve`. 

There is some overlap with https://github.com/tkhq/rust-sdk/pull/117. Given that image validation is lighter (just one API call) compared to this PR, my plan is to default to image validation during `deploy create`, and run this only when the flag is provided (but skip image validation since the functionality here is a superset of that one).

## Implementation

To avoid requiring docker, i added `oci-client` to pull the image, `flate2` and `tar` to read layers. Down to talk about reworking this change if these crates are too bloat-y.

### DIFF
--- a/tvc/src/commands/deploy/approve.rs
+++ b/tvc/src/commands/deploy/approve.rs
@@ -1,8 +1,10 @@
 //! Approve deploy command - cryptographically approve a QOS manifest.
 
+use super::validate_pivot_digest::print_result;
 use crate::config::app::KNOWN_SHARE_SET_KEYS;
 use crate::config::turnkey::{Config, StoredQosOperatorKey};
 use crate::pair::LocalPair;
+use crate::pivot_digest::{compute_pivot_digest, validate_expected_digest, PivotDigestSource};
 use crate::util::{read_file_to_string, write_file};
 use anyhow::{anyhow, bail, Context};
 use clap::{ArgGroup, Args as ClapArgs};
@@ -18,6 +20,13 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use turnkey_client::generated::{
     CreateTvcManifestApprovalsIntent, GetTvcDeploymentRequest, TvcManifestApproval,
 };
+
+struct DeployApprovalSource {
+    manifest: Manifest,
+    manifest_id: String,
+    pivot_image_url: Option<String>,
+    pivot_path: Option<String>,
+}
 
 /// Cryptographically approve a QOS manifest for a deployment with your operator's manifest set key.
 #[derive(Debug, ClapArgs)]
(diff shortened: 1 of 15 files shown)


tvc/src/commands/deploy/approve.rs
```diff
@@ -57,6 +66,14 @@ pub struct Args {
     #[arg(long, help_heading = "Operator signing key", value_name = "PATH")]
     pub operator_seed: Option<PathBuf>,
 
+    /// Locally validate the pivot digest before approval. Only supported with `--deploy-id`.
+    #[arg(long)]
+    pub validate_pivot_digest: bool,
+
+    /// Path to an unencrypted Docker-style pull secret JSON file.
+    #[arg(long, value_name = "PATH")]
+    pub pull_secret: Option<PathBuf>,
```
emostov, line 75: Do people normally store the pull secret as a file? either way i think they could also do --pull-secret <(echo "contents")

tvc/src/commands/deploy/approve.rs
```diff
@@ -76,15 +93,48 @@ pub struct Args {
...
+        let result = compute_pivot_digest(
+            &PivotDigestSource {
+                image_url: pivot_image_url,
+                pivot_path,
+            },
+            args.pull_secret.as_deref(),
+        )
+        .await?;
+        validate_expected_digest(&result.digest, &hex::encode(manifest.pivot.hash))?;
+        print_result(&result);
```
emostov, line 134: nit: i would prefer using path so it reads `validate_pivot_digest::print_result` bc otherwise print_result sounds vague

emostov:
>To avoid requiring docker, i added oci-client to pull the image, flate2 and tar to read layers. Down to talk about reworking this change if these crates are too bloat-y.

This could be a bit rough because we likely need to review all the source code for these. I can double check with our code review tooling since a small chance its already included in trusted repos

richardpringle:
> > To avoid requiring docker, i added oci-client to pull the image, flate2 and tar to read layers. Down to talk about reworking this change if these crates are too bloat-y.
> 
> This could be a bit rough because we likely need to review all the source code for these. I can double check with our code review tooling since a small chance its already included in trusted repos

Adding that if we add any type of oci support in qos, we *should* be able to revive this easily enough

PR #120, opened 2026-04-09
feat(enclave_encrypt): add persistent server key support

emostov:
Add support for server side decryption using a persistent secret.


enclave_encrypt/src/server.rs
```diff
@@ -206,3 +206,53 @@ impl EnclaveEncryptServerRecv {
         }
     }
 }
+
+/// A reusable receiving-side server abstraction backed by a stable target keypair.
+///
+/// Unlike [`EnclaveEncryptServerRecv`], this type does not consume its key material after a
+/// successful decrypt. It is intended for request transport decryption where the server key is
+/// derived from durable signer key material (e.g. the Encryption Quorum Key )
+pub struct BlobEnclaveEncryptServerRecv {
```
r-n-o: I don't think we should name this "blob". It's irrelevant, cryptographically (everything is a blob anyway!), and confusing for readers.

You already called out the meaningful difference between this new type and the `EnclaveEncryptServerRecv` type: it doesn't enforce one-shot encryption semantics.

How about `StableEnclaveEncryptServerRecv` or `DurableEnclaveEncryptServerRecv` or `ReusableEnclaveEncryptServerRecv`?

r-n-o: Another idea would be to combine types into one, and add some sort of optionality to `EnclaveEncryptServerRecv` (constructor params? Maybe builder pattern `::with_stable_receiver_key`?)

emostov: Renaming to `ReusableEnclaveEncryptServerRecv` 👍

enclave_encrypt/README.md
```diff
@@ -144,6 +144,8 @@ We achieve recipient authentication for both the server and client:
 
 The underlying HPKE spec does not provide forward secrecy on the recipient side since the target key can be long lived. To improve forward secrecy we specify that the target key should only be used once by the sender and receiver. We cannot enforce this strictly on the client-side because a client may choose to reuse their key. We could implement timestamp-based validation or rate limiting client-side but it wouldn't be a complete solution. For now we accept that a client can use an encryption bundle multiple times if it so desires. However we enforce one-time use of the key pair on the enclave side by deleting it once a successful decryption happens.
 
+For the encrypted blob API, we introduced a variant that leverages a long lived enclave encryption target. For the encrypted blob API threat model we do not consider Quorum Key reconstruction outside the enclave in scope.
```
r-n-o: I don't think this comment is relevant since this crate is standalone. The reader won't have any knowledge about "the encrypted blob API" or "Quorum Key reconstruction out side of the enclave". Let's talk about why one-shot semantics are okay to violate sometimes, maybe with some examples, but in general terms.

I think this crate has now deviated a bit? The top of this README says "encryption and decryption are "one-shot" using the HPKE standard (RFC 9180)"

We're still RFC compliant but we're implementing a different mode of HPKE. Instead of pre-keyed HPKE we're simply implementing "base mode" HPKE.

enclave_encrypt/src/client.rs
```diff
@@ -292,6 +292,26 @@ pub struct EnclaveEncryptClient {
 }
 
 impl EnclaveEncryptClient {
+    /// Encrypt a message directly to a fixed transport public key.
+    ///
+    /// This is intended for request ingress flows where the target public key is already known
+    /// and authenticated out of band.
+    pub fn encrypt_to_target_public_key(
```
r-n-o: I think this shouldn't be done here: this client is supposed to be used with one-shot semantics!

let's create `BaseModeHPKEClient` or `ReusableEnclaveEncryptClient`?

emostov: I can just make it a one off helper function since its stateless

enclave_encrypt/README.md
emostov: Check out my updated docs

enclave_encrypt/src/client.rs
```diff
@@ -500,7 +500,7 @@ impl EnclaveEncryptClient {
         }
     }
 
-    /// Decrypt a base64 serialized email recovery or auth payload.
+    /// Decrypt a base58 serialized email recovery or auth payload.
```
richardpringle, line 503: nice

enclave_encrypt/src/client.rs
```diff
@@ -544,6 +544,25 @@ impl EnclaveEncryptClient {
...
+) -> Result<Vec<u8>, EnclaveEncryptError> {
+    let receiver_public = <Kem as KemTrait>::PublicKey::from_bytes(target_public_key_bytes)
+        .map_err(EnclaveEncryptError::InvalidServerTarget)?;
+    let (ciphertext, encapped_public) = encrypt(&receiver_public, plaintext, TURNKEY_HPKE_INFO)?;
+    let msg = ClientSendMsg {
+        encapped_public: encapped_public.to_bytes().to_vec().try_into()?,
+        ciphertext,
+    };
+
+    serde_json::to_vec(&msg).map_err(|_| EnclaveEncryptError::FailedToSerializeData)
```
richardpringle: Haven't read through this codebase yet, but so all you're going to get from me is a bunch of pedantry and nits. My goal is to make you think "there's a better way," then resolve my comment and write the "better" (as long as you actually agree with me) next time. 

With that said, I like to keep data in the domain model as long as possible and only serialize at the edges. IMO, this function is doing too much and should only return a `ClientSendMsg` to be serialized later.

richardpringle: I also haven't seen how this is used in practice... so I could be way off base

enclave_encrypt/src/lib.rs
```diff
@@ -525,4 +530,72 @@ mod tests {
...
+
+        // Client encrypts to server's stable target public key
+        let client_ciphertext_bytes =
+            encrypt_to_server_target(&example_credential(), &target_public_key_bytes).unwrap();
+        let client_msg: ClientSendMsg = serde_json::from_slice(&client_ciphertext_bytes).unwrap();
+
+        // Persistent server can decrypt without wiping
+        assert_eq!(server.decrypt(&client_msg).unwrap(), example_credential());
+
+        // Server encrypts to client
```
richardpringle: I think this should be a separate test, it's testing a completely different code path

enclave_encrypt/src/lib.rs
```diff
@@ -525,4 +530,72 @@ mod tests {
...
+                .decrypt(&server_ciphertext_bytes, organization_id)
+                .unwrap(),
+            example_credential()
+        );
+        assert_eq!(
+            client.decrypt(&server_ciphertext_bytes, organization_id),
+            Err(EnclaveEncryptError::ClientAlreadyUsedToDecrypt)
+        );
+
+        // Fresh client: tampered server signature is rejected
```
richardpringle: I think this should also be another test

enclave_encrypt/src/server.rs
```diff
@@ -206,3 +207,53 @@ impl EnclaveEncryptServerRecv {
         }
     }
 }
+
+/// A reusable receiving-side server abstraction backed by a stable target keypair.
+///
+/// Unlike [`EnclaveEncryptServerRecv`], this type does not consume its key material after a
+/// successful decrypt. It is intended for request transport decryption where the server key is
+/// derived from durable signer key material (e.g. a encryption quorum key)
```
richardpringle, line 216: 🥇

enclave_encrypt/src/server.rs
```diff
@@ -206,3 +207,53 @@ impl EnclaveEncryptServerRecv {
...
+pub struct ReusableEnclaveEncryptServerRecv {
+    /// Server's encryption target secret.
+    target_private: <Kem as KemTrait>::PrivateKey,
+    /// Server's encryption target that the client encrypts to.
+    target_public: <Kem as KemTrait>::PublicKey,
+}
```
richardpringle, line 222:
```suggestion
pub struct ReusableEnclaveEncryptServerRecv<K: KemTrait> {
    /// Server's encryption target secret.
    target_private: K::PrivateKey,
    /// Server's encryption target that the client encrypts to.
    target_public: K::PublicKey,
}
```

enclave_encrypt/src/server.rs
```diff
@@ -206,3 +207,53 @@ impl EnclaveEncryptServerRecv {
...
+
+impl ReusableEnclaveEncryptServerRecv {
```
richardpringle, line 224: Goes with my comment above

```suggestion
impl<K: KemTrait> ReusableEnclaveEncryptServerRecv<K> {
```

enclave_encrypt/src/server.rs
```diff
@@ -206,3 +207,53 @@ impl EnclaveEncryptServerRecv {
...
+    ) -> Result<Self, EnclaveEncryptError> {
+        let target_private_bytes = encryption_secret_key.to_bytes();
+        let target_public_bytes = encryption_secret_key
+            .public_key()
+            .to_encoded_point(false)
+            .to_bytes();
+
+        let target_private = <Kem as KemTrait>::PrivateKey::from_bytes(&target_private_bytes)
+            .map_err(EnclaveEncryptError::InvalidTargetPrivateKey)?;
+        let target_public = <Kem as KemTrait>::PublicKey::from_bytes(&target_public_bytes)
```
richardpringle, line 255:
```suggestion
        let target_private = K::PrivateKey::from_bytes(&target_private_bytes)
            .map_err(EnclaveEncryptError::InvalidTargetPrivateKey)?;
        let target_public = K::PublicKey::from_bytes(&target_public_bytes)
```

enclave_encrypt/src/server.rs
```diff
@@ -206,3 +207,53 @@ impl EnclaveEncryptServerRecv {
...
+    /// Server's encryption target secret.
+    target_private: <Kem as KemTrait>::PrivateKey,
+    /// Server's encryption target that the client encrypts to.
+    target_public: <Kem as KemTrait>::PublicKey,
+}
+
+impl ReusableEnclaveEncryptServerRecv {
+    /// Create a server receiver from secret material. Meant to be easily interoperable
+    /// with `qos_p256::P256Pair::encryption_secret`.
+    pub fn from_encryption_key(
```
richardpringle: Wouldn't you rather implement `TryFrom` instead? I've always found that it's better to decouple type conversions from constructors. Logic would obviously be the same

enclave_encrypt/src/server.rs
```diff
@@ -206,3 +207,53 @@ impl EnclaveEncryptServerRecv {
...
+        Ok(Self {
+            target_private,
+            target_public,
+        })
+    }
+
+    /// Decrypt a message from a client.
+    pub fn decrypt(&self, msg: &ClientSendMsg) -> Result<Vec<u8>, EnclaveEncryptError> {
+        let encapped_public = <Kem as KemTrait>::EncappedKey::from_bytes(&*msg.encapped_public)
+            .map_err(EnclaveEncryptError::InvalidEncappedKey)?;
```
richardpringle: My comments all have a theme. I find it's better to decouple conversion from logic. It's flattens out logic and makes debugging easier when something does go wrong. It also make code a lot easier to change and evolve. 

This might not entirely make sense, but I would optimize for writing call-site code like this:

```rust
let encapped_msg = client_msg.try_into()?;
enclave_encrypt_server.decrypt(encapped_msg)?; 
```

It's definitely more verbose, but you make all your stack traces a little smaller. Also, the less you do in any one function, the more composable it is.

enclave_encrypt/README.md
```diff
@@ -144,6 +144,8 @@ We achieve recipient authentication for both the server and client:
...
 
+In addition, we also support long lived enclave side keys. This allows clients to reuse a stable enclave encryption target, which can be helpful to reduce communication overhead for scenarios that are latency or throughput sensitive. We still recommend that the client does not reuse their key. The library types that support this pattern are `server::ReusableEnclaveEncryptServerRecv` and `client::encrypt_to_server_target`. Clients are still recommended to use `EnclaveEncryptClient` to receive messages from the server since it ensures the client target key is deleted after use.
```
richardpringle: I feel like I'm reaching for things to say again here... But if "we recommend that the client does not reuse their key," should this be behind a feature flag that isn't a part of the default features?

richardpringle (commented):
Haha a whole lot of pedantry 😅 

Good for thoughts and discussion, but no necessary changes to the code.

I do however think you should break that test up into three tests though

enclave_encrypt/src/client.rs
emostov: I think the original justification was that the caller of this function can be unaware of the types going of the wire and we basically are saying that JSON bytes will go over the wire. This is consistent with the rest of the code. That being said. I agree we should move towards having serializers at the edge since it seems to be better practice for public apis

enclave_encrypt/src/server.rs
emostov: Kem is actually a type alias for the concrete type `type Kem = hpke::kem::DhP256HkdfSha256;`. I am casting as `KemTrait` because I want the associated type i.e https://docs.rs/hpke/0.13.0/hpke/trait.Kem.html#associatedtype.PublicKey

Lmk if you think there is a cleaner way to do this

enclave_encrypt/src/server.rs
emostov: https://github.com/tkhq/rust-sdk/pull/120/changes#r3083199308

enclave_encrypt/src/server.rs
emostov: cfd54ee2d7b070f1398187c1b66e345f6d68da21

enclave_encrypt/src/server.rs
emostov: https://github.com/tkhq/rust-sdk/pull/120/changes#r3083199308

enclave_encrypt/src/server.rs
emostov: . It feels a little confusing to read to have a try_into that takes a reference and then basically just takes one from the type to create a public key. I can be convinced, but for now this feels most transparent to the reader at this call site

enclave_encrypt/README.md
emostov: We don't have any client code paths that allow a client to re-use a key. So maybe this readme is mislead. Will change wording to _must_ or _can only_

enclave_encrypt/src/client.rs
```diff
@@ -544,6 +544,26 @@ impl EnclaveEncryptClient {
...
 }
 
+/// Encrypt a message directly to a fixed transport public key.
+///
+/// This is intended for request ingress flows where the target public key is already known
+/// and authenticated out of band.
+pub fn encrypt_to_server_target(
+    plaintext: &[u8],
+    server_target: &QuorumPublicKey,
+) -> Result<ClientSendMsg, EnclaveEncryptError> {
```
richardpringle: 👍

enclave_encrypt/src/quorum_public_key.rs
```diff
@@ -69,6 +69,20 @@ impl QuorumPublicKey {
         VerifyingKey::from_sec1_bytes(&self.bytes[65..])
             .map_err(|_| EnclaveEncryptError::InvalidVerifyingKeyBytes)
     }
+
+    /// Returns the encryption target `PublicKey` for this `QuorumPublicKey`
+    pub fn encrypt_public_key(&self) -> Result<p256::PublicKey, EnclaveEncryptError> {
+        p256::PublicKey::from_sec1_bytes(&self.bytes[..65])
```
richardpringle, line 75: Not for this PR, but there should be a constant somewhere that gets length 64.

enclave_encrypt/src/server.rs
richardpringle: I think you should just make the type generic. If it only works with one type that implements `KemTrait`, you can do `impl ReusableEnclaveEncryptServerRecv<Kem>` (no generic - `impl<K>`). 

Don't have to though

enclave_encrypt/src/server.rs
```diff
@@ -206,3 +207,55 @@ impl EnclaveEncryptServerRecv {
...
+            &encapped_public,
+            &self.target_private,
+            &self.target_public,
+            &msg.ciphertext,
+            TURNKEY_HPKE_INFO,
+        )
+    }
+}
+
+impl TryFrom<&p256::SecretKey> for ReusableEnclaveEncryptServerRecv {
```
richardpringle: :+1:

richardpringle (dismissed):

enclave_encrypt/src/client.rs
```diff
@@ -544,6 +544,26 @@ impl EnclaveEncryptClient {
     }
 }
 
+/// Encrypt a message directly to a fixed transport public key.
+///
+/// This is intended for request ingress flows where the target public key is already known
+/// and authenticated out of band.
+pub fn encrypt_to_server_target(
```
r-n-o: Let's move this to `quorum_public_key.rs` given this has nothing to do with the non-reusable client defined in this file (but it seems it's tied to the `QuorumPublicKey` type!)

enclave_encrypt/src/lib.rs
```diff
@@ -525,4 +528,23 @@ mod tests {
...
+        // Create a persistent server receiver from the quorum encryption secret
+        let server: ReusableEnclaveEncryptServerRecv =
+            quorum_key_pair.encryption_key().try_into().unwrap();
+
+        // Client encrypts to server's stable target public key
+        let client_msg = encrypt_to_server_target(
+            &example_credential(),
+            &quorum_key_pair.public_key().try_into().unwrap(),
+        )
+        .unwrap();
```
r-n-o: Idea (not sure if you've already debated this): encrypt can be a method on the QuorumPublicKey? quorum_key_pair.public_key().encrypt?

Actually: is `quorum_key_pair.encryption_key().encrypt(...)` not a thing?

r-n-o: I don't know I'm just getting confused here maybe but this still doesn't feel like the correct API.

We're _either_ encrypting to a key and decrypting with that key without extra logic: in this case I'd recommend simply using the QOS types or augmenting them.

But I don't think that's what's happening because we're not using `quorum_key_pair.public_key().encrypt(...)` directly.

Instead we're casting the P256PublicKey (qos type) into this crate's "QuorumPublicKey" type (which is just an array of bytes). Then we're taking the first half of these bytes (the encryption public key), and finally we're using HPKE encrypt on this public key.

I'd recommend making a single interface for encryption clients: ReusableEnclaveEncryptClient::from_quorum_public_key.

Then you can have your server: `ReusableEnclaveEncryptServer::from_quorum_key_pair` or `ReusableEnclaveEncryptServer::from_quorum_seed`

On top of these, you need a single method: client has "encrypt", server has "decrypt". Nothing else.

With that the test (client code in general, really!) should be a lot easier to digest:

```rust
let test_msg = "...".as_bytes();
let server_key_pair = P256Pair::generate().unwrap();
let server_public = &server_key_pair.public_key();

let client = ReusableEnclaveEncryptClient::from_quorum_public_key(&server_public);

let server = ReusableEnclaveEncryptServer::from_quorum_key_pair(&server_key_pair);

let encrypted = client.encrypt(test_msg);
let decrypted = server.decrypt(&encrypted);

assert_eq!(decrypted, test_msg);
```

r-n-o (commented):
Sorry I keep coming up with ideas; every iteration of this PR genuinely helps clarify though! The clear thing to iterate on is client code and what we're asking people to do. I don't know if we're going for quorum-key-specific vs quorum-key agnostic; my suggestion here is to go _more_ quorum-key-specific since we're sort of tied to QOS already. But there's another set of APIs we could have to focus enclave encrypt on a straight encrypt to pubkey without baking in QOS assumptions.

enclave_encrypt/src/lib.rs
emostov: Yes, that makes sense. I can implement `From<QuorumPublicKey> for ReusableEnclaveEncryptClient` and then initializing it will be `let client: ReusableEnclaveEncryptClient = server_public.into()`

enclave_encrypt/src/lib.rs
```diff
@@ -525,4 +528,25 @@ mod tests {
...
+    #[test]
+    fn client_to_reusable_server() {
+        let plaintext = b"thirst mutilator";
+        let quorum_key_pair = P256Pair::generate().unwrap();
+        let quorum_public_key: QuorumPublicKey = quorum_key_pair.public_key().try_into().unwrap();
+        let quorum_public_key_clone = quorum_public_key.clone();
+
+        // Create a persistent server receiver from the quorum encryption secret
+        let server: ReusableEnclaveEncryptServerRecv =
+            quorum_key_pair.encryption_key().try_into().unwrap();
```
r-n-o: Let's implement `TryFrom` for P256Pair? I'd reduce the knowledge clients have to have about encryption vs signing keys. You just create a `ReusableEnclaveEncryptServerRecv` from a P256Pair.

enclave_encrypt/src/lib.rs
```diff
@@ -525,4 +528,25 @@ mod tests {
...
+
+        // Client encrypts to server's stable target public key
+        let client: ReusableEnclaveEncryptClientSend = quorum_public_key.into();
```
r-n-o, line 543: And for the client I think we should stick with the qos types, for consistency (given we leverage qos types for the server side anyway). No need to go through the intermediate `QuorumPublicKey` type we have in this crate?

`ReusableEnclaveEncryptClientSend::from(P256Pair::generate().unwrap().public_key())` should just work

r-n-o (commented):
This is looking great! Couple of suggestions to make the API more consistent but small potatoes.

enclave_encrypt/src/server.rs
```diff
@@ -206,3 +207,64 @@ impl EnclaveEncryptServerRecv {
...
+            .map_err(EnclaveEncryptError::InvalidTargetPrivateKey)?;
+        let target_public = <Kem as KemTrait>::PublicKey::from_bytes(&target_public_bytes)
+            .map_err(EnclaveEncryptError::InvalidServerTarget)?;
+
+        Ok(Self {
+            target_private,
+            target_public,
+        })
+    }
+}
```
r-n-o: I'm not sure what the best practices are in Rust but I'd like for this to be a private interface so people aren't wondering "should I get a secret key or a p256 key pair", to create my server?

We want the interface to be painfully obvious: provide a `qos_p256::P256Pair`, get a `ReusableEnclaveEncryptServerRecv`. That's the only way.

(I personally find an explicit `::new` returning a Result clearer than a TryFrom impl but that might be taste and convention; I like methods because they autocomplete better; do you find it as easy to discover TryFrom implementations compared to methods on a struct?)

enclave_encrypt/src/quorum_public_key.rs
```diff
@@ -69,6 +69,20 @@ impl QuorumPublicKey {
...
+    }
+}
+
+impl TryFrom<qos_p256::P256Public> for QuorumPublicKey {
+    type Error = EnclaveEncryptError;
+
+    fn try_from(value: qos_p256::P256Public) -> Result<Self, Self::Error> {
+        QuorumPublicKey::from_bytes(value.to_bytes())
+    }
 }
```
r-n-o, line 86: I wouldn't add this here. Instead, let's add `TryFrom<qos_p256::P256Public> for ReusableEnclaveEncryptClientSend` directly?

emostov: I lean towards that client side apis shoulnd only be using public types from this crate to keep things simple and consistent. That means the client side code should use `QuorumPublicKey`. The reason being is that the client side code never uses `qos_p256` and `QuorumPublicKey` is meant to have helper functions that describe our production public keys - which is exactly how we want clients to use this. I think in practice this TryFrom helper will only be useful in tests because we test both the server and client sides together

Example of helper function on `QuorumPublicKey`

https://github.com/tkhq/rust-sdk/blob/3804262b3de23589eab65945c49809b00c4ecee2/enclave_encrypt/src/quorum_public_key.rs#L48-L51

emostov: This is working backwards from the principles that:

- server side code is being used in QOS enclaves, which will already be using qos_p256 and thus know that library surface area
- client side potentially may only be using the client to just encrypt and there should be a happy path with turnkey's signer

r-n-o: Ah yes, good point re: this being used in different contexts 👍 

I'm onboard!

enclave_encrypt/src/server.rs
emostov: I removed this in 8bd000a04a0ef10f67434f77a21456332d4026ca

For the constructor for `TryInto` - I talked to @richardpringle and we agreed that 
1. It's not idomatic to have a fallible constructor
2. Try to do type conversions whenever possible
3. We assume when crafting crate APIs that readers are looking at the rust docs. When looking at the rustdocs, if there is no constructor and private fields the next most obvious thing is looking for a `From`/`TryFrom` imp

enclave_encrypt/src/lib.rs
emostov: resolved here https://github.com/tkhq/rust-sdk/pull/120#discussion_r3102805964

enclave_encrypt/src/lib.rs
emostov: done in [8bd000a](https://github.com/tkhq/rust-sdk/pull/120/commits/8bd000a04a0ef10f67434f77a21456332d4026ca)

enclave_encrypt/src/server.rs
richardpringle: @r-n-o your reputation for great PR comments precedes you! I had to think about this one...

It was me that convinced @emostov to implement the `TryFrom` trait. I would say Rust best practice is to try and do type conversion where possible so that you can think about your code paths like state-machines. It limits the API surface for making mistakes. It's also best practice _not_ to have fallible constructors where possible (which _usually_ means that you need to do whatever type manipulation outside of the constructor). As for optimizing for autocomplete... that's just not something that you do when following best practices or you would never use traits. I highly highly recommend having cargo docs built locally and often referring to them. They are currently the best way that I know of to see trait implementations and implementors. If I see a type with private fields and no constructors, I know that there has to be a type conversion and I immediately go to look for `From` or `TryFrom` implementations. 

With that said...

@emostov, to @r-n-o's point, I think we can just have a single `TryFrom<P256Pair>` if we don't want a public impl that converts from a secret-key. 

Also, do we want to do an owned type conversion instead of borrowing so that the original allocation can be re-used? We also probably want to zeroize this thing on `Drop`, don't we?

richardpringle: Ah nvm, you already did it

emostov: For zeroize on drop - good call out: from the top of my head the `Kem` type should zeroize and `P256Pair` should zerioze. The reason I took a ref to `P256Pair` is that I'm pretty sure it doesn't implement clone and we almost always pass it by reference in apps. That leaves `target_private_bytes` .. and that appears to not implement zeroize so might need to wrap it

emostov: Pushed a commit to zeroize secret bytes https://github.com/tkhq/rust-sdk/pull/120/changes/c16833aa6bd1e31486dbe6d36d187dd10556baff

r-n-o: Down to keep `TryFrom` instead of implementing a fallible constructor; let's go with lang convention 👍

Cargo.toml
```diff
@@ -18,8 +18,8 @@ resolver = "2"
 
 [workspace.dependencies]
 # QuorumOS
-qos_core = { version = "0.5.0", default-features = false }
-qos_p256 = { version = "0.5.0", default-features = false }
+qos_core = { version = "0.6.1", default-features = false }
+qos_p256 = { version = "0.6.1", default-features = false }
```
r-n-o, line 22: Want to do this in a separate PR for clarity? Or do you need this in this PR? (if you do, let's keep it here; if you don't let's make a separate PR I'd say, changelog will be clearer)

r-n-o (approved):
🎖️

Cargo.toml
emostov: Sorry missed this one - I need it to get access to the `P256Pair` encryption secret

PR #123, opened 2026-04-14
fix: relax exact tokio version pin to semver-compatible range

### LINKED ISSUE #122: Library crate should not pin exact tokio version (=1.47.1)
## Problem

The workspace `Cargo.toml` pins tokio with an exact version constraint:

```toml
tokio = { version = "=1.47.1", default-features = false }
```

This propagates to the published `turnkey_client` crate on crates.io, which means **any application that depends on `turnkey_client` must use tokio 1.47.1 exactly**. If the application (or any of its other dependencies) requires a different tokio 1.x version, Cargo cannot resolve the dependency tree.

## Impact

For example, an application using `tokio = "1.51.1"` alongside `turnkey_client = "0.6.2"` gets:

```
error: failed to select a version for `tokio`.
    ... required by package `turnkey_client v0.6.2`
    versions that meet the requirements `=1.47.1` are: 1.47.1
    all possible versions conflict with previously selected packages.
    previously selected package `tokio v1.51.1`
```

This is particularly problematic when other ecosystem crates (e.g., Solana SDK, alloy) also depend on tokio and may require newer patch versions.

I've opened a PR with this one-line fix: (linked below).

SonOfMosiah:
## Summary

- Change `tokio = "=1.47.1"` to `tokio = "1.47"` in workspace `Cargo.toml`
- Fixes #122 — the exact pin prevents downstream crates from resolving tokio when any other dependency needs a different 1.x patch version

## Context

Library crates published to crates.io should use semver-compatible ranges (`"1.47"` = `^1.47` = `>=1.47.0, <2.0.0`) rather than exact pins (`"=1.47.1"`). The exact pin makes the crate unusable alongside any dependency that requires a different tokio 1.x version — for example, applications using Solana SDK or alloy which pull in newer tokio patches.

Reference: https://users.rust-lang.org/t/impact-of-pinning-dependencies/17634

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ mime = { version = "0.3.17", default-features = false }
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls", "http2"] }
 
 # Async runtime
-tokio = { version = "=1.47.1", default-features = false }
+tokio = { version = "1.47", default-features = false }
 
 # Protocol buffers
 prost = { version = "0.12", default-features = false, features = ["std"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ mime = { version = "0.3.17", default-features = false }
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls", "http2"] }
 
 # Async runtime
-tokio = { version = "1.47", default-features = false }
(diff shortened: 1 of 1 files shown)


emostov:
Thank you for the contribution! We ended up merging effectively the same code here https://github.com/tkhq/rust-sdk/pull/127

PR #126, opened 2026-04-21
feat(tvc): add provisioning-details command

annieke:
This PR adds the functionality for the tvc cli to get the provisioning details of a deployment, and decode it into a human readable summary. The summary provides the deployment's provisioning ephemeral key, which a user can use to re-encrypt their quorum key shares.

The functionality is close to `qos_client get-ephemeral-key-hex`, with output that's compatible with `qos_client`, but wraps it to be for tvc instead of a local machine.


tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,311 @@
...
+/// Get provisioning details for a deployment.
+#[derive(Debug, ClapArgs)]
+#[command(about, long_about = None)]
+pub struct Args {
+    /// ID of the deployment.
+    #[arg(short = 'd', long, env = "TVC_DEPLOY_ID")]
+    pub deploy_id: String,
+
+    /// Never use for sensitive applications! Skip attestation document verification.
+    #[arg(long, alias = "unsafe-skip-attestation")]
```
emostov: what is the motivation for this specific alias?

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,311 @@
...
+        .get_tvc_deployment_provisioning_details(request)
+        .await
+        .context("failed to fetch deployment provisioning details")?;
+
+    if response.attestation_document.is_empty() {
+        bail!("attestation document missing in provisioning details response");
+    }
+    if response.manifest_envelope.is_empty() {
+        bail!("manifest envelope missing in provisioning details response");
+    }
```
emostov, line 98: more idmiomatic way to do this is probs like

```
let (Some(a), Some(m)) = (args.attestation_doc.as_ref(), args.manifest_envelope_out.as_ref()) else {
  bail!(..error essage)
}
```

Since presumably if you want to ensure they are Some here then downstream logic can be simplified to not handle the option

and if you wanted more granular message you could do a match

```
let (a, m) = match (args.attestation_doc_out.as_ref(), args.manifest_envelope_out.as_ref()) {
 (Some(a), Some(m)) => (a, m),
  None, Some(_) => bail!(..),
  Some(_), None => bail!(..),
}
```

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,311 @@
...
+    println!("{label} written to: {}", path.display());
+    Ok(())
+}
+
+fn parse_manifest_envelope(bytes: &[u8]) -> anyhow::Result<ManifestEnvelope> {
+    if let Ok(manifest_envelope) = serde_json::from_slice::<ManifestEnvelope>(bytes) {
+        return Ok(manifest_envelope);
+    }
+
+    ManifestEnvelope::try_from_slice_compat(bytes)
```
emostov: I think we can get rid of this since we are no longer supporting json in the future

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,311 @@
...
+            cose_sign1_der,
+            &qos_nsm::nitro::cert_from_pem(qos_nsm::nitro::AWS_ROOT_CERT_PEM)
+                .context("failed to parse AWS Nitro root certificate")?,
+            validation_time_secs(validation_time_override)?,
+        )
+        .context("failed to parse and verify attestation document")?
+    };
+
+    if !dangerous_skip_attestation {
+        qos_nsm::nitro::verify_attestation_doc_against_user_input(
```
emostov: smol nit: might as well move this to else statement above so its clear in one block what we do when we don't want to skip

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,311 @@
...
+        )
+        .context("attestation document did not match manifest expectations")?;
+    }
+
+    Ok(AttestationSummary {
+        ephemeral_key: extract_ephemeral_key(
+            attestation_doc
+                .public_key
+                .as_ref()
+                .map(|public_key| public_key.as_ref()),
```
emostov, line 204: I think if we had `extract_ephemeral_key(Option<&Vec<u8>>)` you would need this map after `as_ref()`. Not that its necessarily better just took me a sec to figure out the types here

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,311 @@
...
+        )?,
+        module_id: attestation_doc.module_id,
+        digest: attestation_doc.digest.into(),
+        timestamp_ms: attestation_doc.timestamp,
+        user_data: attestation_doc
+            .user_data
+            .as_ref()
+            .map(|user_data| user_data.as_ref().to_vec()),
```
emostov: could you just do `user_data.clone()`?

tvc/src/commands/deploy/provisioning_details.rs
emostov: Either way its fine just thinking out loud

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,311 @@
...
+        nonce: attestation_doc
+            .nonce
+            .as_ref()
+            .map(|nonce| nonce.as_ref().to_vec()),
+        pcrs: attestation_doc
+            .pcrs
+            .iter()
+            .filter(|(index, _)| **index <= SUMMARY_PCR_MAX_INDEX)
```
emostov: Why filter out higher PCR values? Just because we don't check them?

tvc/src/commands/deploy/provisioning_details.rs
emostov: same for attestation_doc below

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,311 @@
...
+            .map(|(index, pcr)| (*index, pcr.as_ref().to_vec()))
+            .collect(),
+        certificate_len: attestation_doc.certificate.len(),
+        ca_bundle_cert_count: attestation_doc.cabundle.len(),
+    })
+}
+
+#[cfg(test)]
```
emostov: Not idiomatic to have a test only function outside of a test modules

tvc/README.md
```diff
@@ -46,4 +46,23 @@ tvc deploy approve \
...
+# enclave ephemeral key plus parseable attestation fields.
+tvc deploy provisioning-details \
+  --deploy-id <DEPLOYMENT_UUID>
+
+# Save the raw attestation doc and manifest envelope locally for later
+# re-encryption / submission flows.
+tvc deploy provisioning-details \
+  --deploy-id <DEPLOYMENT_UUID> \
+  --attestation-doc-out provisioning.attestation.cose \
+  --manifest-envelope-out provisioning.manifest-envelope.json
```
emostov: Just an idea but what if we had it output a single json like:

```json
{
    attestation_doc_cose_sign1_bytes: base64,
    manifest_envelope: {
         manifest: { .. },
         approvals: [..],
   }
   fetch_time: <unix timestamp of when this was fetched>,
   deploy_id: <id of deploy>,
   app_id: <id of app>,
   ephemeral_public_key: 0xThe-ephemeral-public-key-to-help-id-the-enclave-for-debugging
}
```

We could call it like "boot attestation" or "boot bundle" or "provision-bundle.json"?

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,311 @@
...
+    if !dangerous_skip_attestation {
+        qos_nsm::nitro::verify_attestation_doc_against_user_input(
+            &attestation_doc,
+            &manifest_envelope.manifest.qos_hash(),
+            &manifest_envelope.manifest.enclave.pcr0,
+            &manifest_envelope.manifest.enclave.pcr1,
+            &manifest_envelope.manifest.enclave.pcr2,
+            &manifest_envelope.manifest.enclave.pcr3,
+        )
+        .context("attestation document did not match manifest expectations")?;
```
emostov: are we checking approvals anywhere?

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,311 @@
...
+            .as_ref()
+            .map(hex::encode)
+            .unwrap_or_else(|| "(none)".to_string())
+    );
+    println!("PCRs:");
+    for (index, pcr) in &summary.pcrs {
+        println!("  PCR{index}: {}", hex::encode(pcr));
+    }
+    println!("Certificate Length: {} bytes", summary.certificate_len);
+    println!("CA Bundle Certificates: {}", summary.ca_bundle_cert_count);
```
emostov, line 284: we should include approvals in summary

tvc/src/commands/deploy/provisioning_details.rs
annieke: oops didn't catch this during a rename; will remove

tvc/src/commands/deploy/provisioning_details.rs
emostov: should be a `check_approvals` method

annieke: not right now, my thought was tvc users will get to this step after submitting approvals but before provisioning, so there will be some checks in the tvc backend already. will add though since it doesn't hurt

tvc/src/commands/deploy/provisioning_details.rs
annieke: yeah after pcr4 all the values were long strings of 0s so it felt noisy. any worth keeping them in?

tvc/src/commands/deploy/provisioning_details.rs
annieke: i think your comment was for the lines below this instead of this line (since the responses are plain `Vec`s instead of `Option`s), so will change that instead

tvc/src/commands/deploy/provisioning_details.rs
emostov: I think my justification is that it would be nice to find a reason to exit early for "invalid" data - in this case we could say missing approvals would make the envelope invalid. We will want to check approvals again before re-encrypting shares

emostov: https://github.com/tkhq/qos/blob/d69bcd348073ed7322c0d0423a35b7ab56831de8/src/qos_core/src/protocol/services/boot.rs#L568

tvc/README.md
annieke: i'm not sure if you had a specific idea around 0x-prefixing the ephemeral pubkey, but addressing this now and just naming the field `ephemeral_public_key_hex` instead of the 0x prefix. let me know if i'm missing something though

emostov: thats good! just added that to indicate its hex but didn't actually means we needed the 0x

tvc/src/commands/deploy/provisioning_details.rs
annieke: the attestation doc returns these fields as ByteBuf, but I just consumed them instead of borrowing and copying to make it a bit less verbose

tvc/src/commands/deploy/provisioning_details.rs
annieke: you mean borsh?

tvc/src/commands/deploy/provisioning_details.rs
annieke: 🫥

tvc/src/commands/deploy/provisioning_details.rs
annieke: added manifest approval threshold checking but just ignoring share set for now since this is supposed to be pre-share-posting

tvc/src/commands/deploy/provisioning_details.rs
emostov: I think thats fine. No need that I know of - was just curious

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,480 @@
...
+    digest: NsmDigest,
+    timestamp_ms: u64,
+    user_data: Option<Vec<u8>>,
+    nonce: Option<Vec<u8>>,
+    pcrs: Vec<(usize, Vec<u8>)>,
+    certificate_len: usize,
+    ca_bundle_cert_count: usize,
+    manifest_set_threshold: u32,
+    manifest_set_approvals: Vec<ApprovalSummary>,
+    share_set_approvals: Vec<ApprovalSummary>,
```
emostov, line 63: I actually don't know if we ever will store the manifest approvals here .. I need to check QOS code. Regardless though you will never have to worry about checking these - they are just for audit trail. So in that regard they are harmless to keep around, just need to remember so people don't get concerned in the future if its empty or something

annieke: when i tried on an existing deploy this was a share of the boot strap quorum key, so i just skipped doing anything with them but printed them out. would be empty before provisioning though

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,480 @@
...
+            manifest_envelope.clone(),
+            fetched_at_unix_ms,
+            &summary.ephemeral_key,
+        );
+        write_provision_bundle(path, &bundle).await?;
+        println!();
+    }
+
+    if args.dangerous_skip_verification {
+        println!("Skipping attestation and approval verification (--dangerous-skip-verification)");
```
emostov: smol nit: should we just put this println right below 124?

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,476 @@
...
+        qos_nsm::nitro::verify_attestation_doc_against_user_input(
+            &attestation_doc,
+            &manifest_envelope.manifest.qos_hash(),
+            &manifest_envelope.manifest.enclave.pcr0,
+            &manifest_envelope.manifest.enclave.pcr1,
+            &manifest_envelope.manifest.enclave.pcr2,
+            &manifest_envelope.manifest.enclave.pcr3,
+        )
+        .context("attestation document did not match manifest expectations")?;
+
```
emostov, line 195: sorry micro nit: it would be nice to have all verification in one place to make it less prone to becoming spaghetti. I think in this block we can have 

```
manifest_envelope
        .check_approvals()
        .context("failed to verify manifest approvals")?;
```

And then we can remove verify_manifest_approvals and its call sites

tvc/README.md
```diff
@@ -46,4 +46,23 @@ tvc deploy approve \
...
+# default, and print the enclave ephemeral key plus parseable attestation
+# fields and approval summaries.
+tvc deploy provisioning-details \
+  --deploy-id <DEPLOYMENT_UUID>
+
+# Save a single provisioning bundle JSON for later re-encryption / submission
+# flows.
+tvc deploy provisioning-details \
+  --deploy-id <DEPLOYMENT_UUID> \
+  --provision-bundle-out provisioning-bundle.json
```
emostov, line 54: I would only add this example to the README since its the one I would expect people to use in the normal deploy path

emostov (commented):
super close

tvc/src/commands/deploy/provisioning_details.rs
emostov: I think it might make sense to have a helper function that does this whole code block:

attestation_doc_from_der
verify_attestation_doc_against_user_input
manifest_envelope.check_approvals()

The reason I am being pedantic is that I think making it easy to read these checks helps audit-ability and I think we will re-use this exact logic in share re-encryption command. I also think you could use the new function for tests and get rid of that test only helper function

annieke: good call! also did a readability pass after this

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,484 @@
...
+            &hex::decode(MOCK_PCR1).unwrap(),
+            &hex::decode(MOCK_PCR2).unwrap(),
+            &hex::decode(MOCK_PCR3).unwrap(),
+        )
+        .context("attestation document did not match manifest expectations")
+        .unwrap_err();
+
+        assert!(err
+            .to_string()
+            .contains("attestation document did not match manifest expectations"));
```
emostov: non blocking but I'm realizing this test is just showing that its an error but in a kind of funny way since we are just wrapping the error in `"attestation document did not match manifest expectations"` and then just checking that exact string exists. Which is essentially just saying "this is an error" but not checking contents of error. A shorter way to do this would just be to call `is_err()` and assert true or `unwrap_err()` and just see that it doesn't panic, on the result of verify_attestation_doc_against_user_input

tvc/src/commands/deploy/provisioning_details.rs
```diff
@@ -0,0 +1,484 @@
...
+
+        qos_nsm::nitro::verify_attestation_doc_against_user_input(
+            &attestation_doc,
+            &hex::decode(MOCK_USER_DATA_NSM_ATTESTATION_DOCUMENT).unwrap(),
+            &hex::decode(MOCK_PCR0).unwrap(),
+            &hex::decode(MOCK_PCR1).unwrap(),
+            &hex::decode(MOCK_PCR2).unwrap(),
+            &hex::decode(MOCK_PCR3).unwrap(),
+        )
+        .context("attestation document did not match manifest expectations")
```
emostov: these context are unnescary I beleive - should just be able to unwrap to show now error

But also - these tests are just testing library functions. Can we test our wrapper around them: `verify_provisioning_details`

tvc/src/commands/deploy/provisioning_details.rs
emostov: these tests are just testing qos_nsm library functions. Can we test our wrapper around them: verify_provisioning_details?

tvc/src/commands/deploy/provisioning_details.rs
annieke: `verify_provisioning_details` requires an attestation doc whose user_data matches the qos_hash and also a manifest approval with a signature that verifies over it, and none of the mocks currently fulfill this requirement. a passing testing required a real document, so only have the failure case for now.

will remove extra context! worth the real document test to you?

annieke: added real bundle as a fixture!

emostov (approved):
🚢

PR #128, opened 2026-04-24
feat(tvc): make tvc cli interactive

daniilrrr:
wip


daniilrrr:
close in favor of more recent #139

PR #129, opened 2026-04-29
feat(tvc): add re-encrypt-share command

annieke:
This PR adds `tvc re-encrypt-share`, which takes a share encrypted to the operator public key and re-encrypts it to a recently fetched attestation doc's ephemeral key. It is designed to work with the output of `tvc deploy provisioning-details`, and can be used on an offline and airgapped machine.

### DIFF
--- a/tvc/src/cli.rs
+++ b/tvc/src/cli.rs
@@ -33,6 +33,7 @@ impl Cli {
                 AppCommands::Create(args) => commands::app::create::run(args).await,
                 AppCommands::Init(args) => commands::app::init::run(args).await,
             },
+            Commands::ReEncryptShare(args) => commands::re_encrypt_share::run(args).await,
             Commands::Login(args) => commands::login::run(args).await,
         }
     }
@@ -52,6 +53,8 @@ enum Commands {
         #[command(subcommand)]
         command: AppCommands,
     },
+    /// Re-encrypt a share for enclave provisioning.
+    ReEncryptShare(commands::re_encrypt_share::Args),
 }
 
 #[derive(Debug, Subcommand)]
--- a/tvc/src/commands/re_encrypt_share.rs
+++ b/tvc/src/commands/re_encrypt_share.rs
@@ -1,0 +1,329 @@
+//! Re-encrypt share command.
+
+use crate::operator_key::load_operator_pair;
+use crate::pair::Pair;
+use crate::provisioning::ProvisionBundle;
+use crate::util::write_file;
+use anyhow::{anyhow, Context};
+use clap::Args as ClapArgs;
+use qos_core::protocol::services::boot::{Approval, ManifestEnvelope, QuorumMember};
+use qos_core::protocol::QosHash;
+use serde::Serialize;
+use std::path::{Path, PathBuf};
+use zeroize::Zeroizing;
+
+/// Re-encrypt a share with an ephemeral key.
+#[derive(Debug, ClapArgs)]
+#[command(about, long_about = None)]
+pub struct Args {
+    /// Path to the share file to re-encrypt.
+    /// Assumed to be encrypted by the operator key.
+    #[arg(long, env = "SHARE_PATH")]
+    pub share_path: PathBuf,
+
+    /// Provision bundle from a deployment.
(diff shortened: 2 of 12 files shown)


tvc/src/commands/re_encrypt_share.rs
```diff
@@ -0,0 +1,329 @@
...
+        help_heading = "Operator encryption key",
+        value_name = "OPERATOR_PATH"
+    )]
+    pub operator_seed: Option<PathBuf>,
+
+    /// Never use for sensitive applications! Skip attestation and manifest approval verification.
+    #[arg(long)]
+    pub dangerous_skip_verification: bool,
+
+    /// Output path for the re-encrypted share.
```
emostov, line 42:
```suggestion
    /// Output path for the re-encrypted share output (share ciphertext + manifest approval).
```

tvc/src/commands/re_encrypt_share.rs
```diff
@@ -0,0 +1,329 @@
...
+/// Run the re-encrypt-share command.
+pub async fn run(args: Args) -> anyhow::Result<()> {
+    if args.dangerous_skip_verification {
+        eprintln!(
+            "WARNING: Skipping verification! This is dangerous and should not be used for sensitive applications."
+        );
+    }
+
+    let provision_bundle = read_provision_bundle(&args.provision_bundle).await?;
+    let encrypted_share = tokio::fs::read(&args.share_path)
```
emostov, line 62: if this is raw share we should at least trim the string to remove new line chars / white space. Maybe also do some basic validation (should it be valid hex? etc)

emostov: One other thing that we can modify in follow up PRs: I think the shares should actually be stored in one json file along with the quorum pub key and threshold - basically what I was proposing as the output of the QK genesis ceremony

tvc/src/commands/re_encrypt_share.rs
```diff
@@ -0,0 +1,329 @@
...
+            pub_key: other_pair.public_key().to_bytes(),
+        }]);
+
+        assert!(
+            find_share_set_member(&manifest_envelope, &operator_pair.public_key().to_bytes())
+                .is_err()
+        );
+    }
+
+    #[tokio::test]
```
emostov, line 296: I don't think this needs to be a tokio test since it doesn't look like any async code

tvc/src/cli.rs
```diff
@@ -52,6 +53,8 @@ enum Commands {
         #[command(subcommand)]
         command: AppCommands,
     },
+    /// Re-encrypt a share for enclave provisioning.
+    ReEncryptShare(commands::re_encrypt_share::Args),
```
emostov, line 57: I don't feel strongly but wonder if this command should be under deploy? Not sure since it doesn't actually touch our api directly so could be deceptive to put it under deploy. On the other hand it just doesn't feel neat here

tvc/src/provisioning.rs
```diff
@@ -0,0 +1,343 @@
...
+        dangerous_skip_verification: bool,
+    ) -> anyhow::Result<P256Public> {
+        self.ephemeral_public_key_with_validation_time(dangerous_skip_verification, None)
+    }
+
+    fn ephemeral_public_key_with_validation_time(
+        &self,
+        dangerous_skip_verification: bool,
+        validation_time_override: Option<u64>,
+    ) -> anyhow::Result<P256Public> {
```
emostov, line 52: nice i like this

emostov (commented):
Took a first pass but need to come back to review tests

tvc/src/cli.rs
annieke: I was a bit torn up about this too! Decided to separate it since the functionality isn't strictly tied to a deploy. 

What do you think about putting it under `tvc offline` or `tvc keys`?

tvc/src/commands/re_encrypt_share.rs
annieke: this test includes filereading which is async 👀

annieke:
closing this after combining into https://github.com/tkhq/rust-sdk/pull/131

PR #131, opened 2026-04-30
feat(tvc): quorum keygen and re-encrypt

annieke:
This PR adds commands for quorum keygen and re-encrypting a share from an ephemeral key. Re-encryption is designed to work with the output of `tvc deploy provisioning-details`.

Both commands are best run on airgapped machines, and do not require an internet connection.


tvc/src/commands/keys/generate_quorum_key.rs
```diff
@@ -0,0 +1,266 @@
...
+#[derive(Debug, ClapArgs)]
+#[command(about, long_about = None)]
+pub struct Args {
+    /// Path to the quorum key config file.
+    pub config_file: PathBuf,
+
+    /// Output file path for the generated quorum key metadata.
+    #[arg(
+        long,
+        default_value = "quorum_key_metadata.json",
```
emostov, line 24: ❤️

tvc/src/commands/keys/generate_quorum_key.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    Ok(parsed)
+}
+
+fn parse_operator_public_key(operator_public_key: &str) -> Result<(String, P256Public)> {
+    let bytes = hex::decode(operator_public_key).context("public key must be bare hex encoded")?;
+    let public = P256Public::from_bytes(&bytes)
+        .map_err(|e| anyhow::anyhow!("invalid QOS P-256 key: {e:?}"))?;
+
+    Ok((hex::encode(public.to_bytes()), public))
```
emostov:
```suggestion
    Ok((public.to_hex_bytes().to_string(), public))
```

tvc/src/commands/keys/generate_quorum_key.rs
```diff
@@ -0,0 +1,266 @@
...
+fn parse_operator_public_keys(operator_public_keys: &[String]) -> Result<Vec<OperatorPublicKey>> {
+    let mut parsed = Vec::with_capacity(operator_public_keys.len());
+    let mut seen = HashSet::new();
+
+    for (index, key) in operator_public_keys.iter().enumerate() {
+        let (normalized, public) = parse_operator_public_key(key)
+            .with_context(|| format!("invalid operator public key at index {index}"))?;
+
+        if !seen.insert(normalized.clone()) {
+            anyhow::bail!("duplicate operator public key at index {index}");
```
emostov: we should interpolate the address as well

tvc/src/commands/keys/generate_quorum_key.rs
emostov: actually this is slightly longer so nvm

tvc/src/commands/keys/generate_quorum_key.rs
```diff
@@ -0,0 +1,266 @@
...
+    let plaintext_shares = PlaintextShares(plaintext_shares);
+
+    let encrypted_shares = operator_publics
+        .iter()
+        .zip(plaintext_shares.0.iter())
+        .map(|(operator_public, share)| {
+            operator_public
+                .public
+                .encrypt(share)
+                .map_err(|e| anyhow::anyhow!("failed to encrypt quorum key share: {e:?}"))
```
emostov: should we just create `EncryptedShareMetadata` here? Don't feel strongly but seems slightly easier to read then re-zipping over a similar pair of sets below when we have all the info to create `EncryptedShareMetadata` here

tvc/src/commands/keys/init_quorum_key.rs
```diff
@@ -0,0 +1,54 @@
...
+    println!("  threshold : >= 2 and <= shares");
+    println!();
+    println!("Edit the file to fill in your values, then run:");
+    println!("  tvc keys generate-quorum-key {}", args.output.display());
+
+    Ok(())
+}
+
+/// Load the operator public key from the active org's config.
+async fn load_operator_public_key() -> Option<String> {
```
emostov, line 49: One thing I am wondering - how will this work in an airgap env? Don't need to address now but might want it for a follow up

emostov: Actually, realizing this can be run online so nvm

tvc/src/config/quorum_key.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// True if any field still contains a placeholder.
+    pub fn has_placeholders(&self) -> bool {
+        self.operator_public_keys
+            .iter()
+            .any(|k| k.starts_with("<FILL_IN"))
+    }
+
+    /// Validate numeric constraints. Called by `generate_quorum_key`
+    /// before consuming the config.
+    pub fn validate(&self) -> anyhow::Result<()> {
```
emostov, line 49: i know we check for dupe pub keys somewhere else but should we be doing it here?

tvc/src/cli.rs
```diff
@@ -81,3 +94,11 @@ enum AppCommands {
     /// Generate a template app configuration file.
     Init(commands::app::init::Args),
 }
+
+#[derive(Debug, Subcommand)]
+enum KeysCommands {
+    /// Generate and encrypt a quorum key from a config file.
```
emostov: `Generate and shamir split a quorum key, encrypting each shamir share to an operator key."

tvc/src/commands/keys/generate_quorum_key.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// Output file path for the generated quorum key metadata.
+    #[arg(
+        long,
+        default_value = "quorum_key_metadata.json",
+        env = "QUORUM_KEY_OUT"
+    )]
+    pub quorum_key_out: PathBuf,
+}
+
+#[derive(Debug, Serialize)]
```
emostov: I think we can also add deserialize derives and then drop the dupe types in the tests. But maybe I will see that when I read the next commit

tvc/tests/keys_generate_quorum_key.rs
```diff
@@ -0,0 +1,138 @@
...
+        .map(|pair| hex::encode(pair.public_key().to_bytes()))
+        .collect::<Vec<_>>();
+
+    fs::write(
+        &config_path,
+        serde_json::to_vec_pretty(&json!({
+            "shares": 3,
+            "threshold": 2,
+            "operatorPublicKeys": [
+                operator_public_keys[0].to_uppercase(),
```
emostov, line 44: maybe dq but why is this one `.to_uppercase()`? is that supposed to test something specific

tvc/src/provisioning.rs
```diff
@@ -0,0 +1,344 @@
...
+
+        Ok(bundled_public_key)
+    }
+}
+
+pub(crate) fn verify_provisioning_details(
+    cose_sign1_der: &[u8],
+    manifest_envelope: &ManifestEnvelope,
+    validation_time_override: Option<u64>,
+) -> anyhow::Result<()> {
```
emostov: should we return the attestation_doc from here so the caller can opt into using it

emostov: that way you don't need to unsafe_attestation_doc_from_der just after calling this

tvc/src/provisioning.rs
```diff
@@ -0,0 +1,344 @@
...
+            manifest_envelope: sample_manifest_envelope(),
+            fetched_at_unix_ms: 1_712_345_678_901,
+            deployment_id: "deploy-123".to_string(),
+            ephemeral_public_key_hex: hex::encode(expected_public_key.to_bytes()),
+        };
+
+        let public_key = bundle.ephemeral_public_key(true).unwrap();
+
+        assert_eq!(public_key.to_bytes(), expected_public_key.to_bytes());
+    }
```
emostov, line 341: do we have test coverage for pcrs matching between attestation doc and manifest? also manifest hash being right manifest in attestation doc?

tvc/src/quorum_key_metadata.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+pub(crate) fn decode_p256_public_key_hex(public_key_hex: &str) -> anyhow::Result<P256Public> {
+    let bytes =
+        hex::decode(public_key_hex.trim()).context("public key must be bare hex encoded")?;
+
+    P256Public::from_bytes(&bytes).map_err(|e| anyhow!("invalid QOS P-256 key: {e:?}"))
+}
+
+pub(crate) fn normalize_p256_public_key_hex(public_key_hex: &str) -> anyhow::Result<String> {
```
emostov, line 59: just an observation but wondering if we should always pass keys around in bytes and just convert them to hex at the edge so we don't need to worry about casing mismatch in hex

i think if we had a type for wrapping vecs and arrays that ensured they always debug/display/serialize to hex that would make this pattern much easier

emostov: (not something that we should address in this pr)

tvc/src/commands/keys/re_encrypt_share.rs
```diff
@@ -0,0 +1,491 @@
...
+use serde::Serialize;
+use std::path::{Path, PathBuf};
+use zeroize::Zeroizing;
+
+/// Re-encrypt a share with an ephemeral key.
+#[derive(Debug, ClapArgs)]
+#[command(about, long_about = None)]
+pub struct Args {
+    /// Path to the quorum key metadata file, which includes the share to re-encrypt.
+    /// Assumed to be encrypted by the operator key.
```
emostov: i think this document needs to be updated to say something along the lines that this is the same shape as the output of quorum key generate

tvc/src/commands/keys/re_encrypt_share.rs
```diff
@@ -0,0 +1,491 @@
...
+    pub provision_bundle: PathBuf,
+
+    /// Path to the file containing the master seed for the operator key.
+    /// If not provided, uses the operator key from the logged-in org config.
+    #[arg(
+        long,
+        help_heading = "Operator encryption key",
+        value_name = "OPERATOR_PATH"
+    )]
+    pub operator_seed: Option<PathBuf>,
```
emostov, line 38: should we make this a global arg for the cli? Not sure, but feels like something if we have it here we would want to allow overrides all the time

tvc/src/commands/keys/re_encrypt_share.rs
```diff
@@ -0,0 +1,491 @@
...
+    })?;
+    serde_json::from_slice(&contents).with_context(|| {
+        format!(
+            "failed to parse quorum key metadata file: {}",
+            path.display()
+        )
+    })
+}
+
+async fn read_provision_bundle(path: &Path) -> anyhow::Result<ProvisionBundle> {
```
emostov: since we have AI to easily generate code writing generic code doesn't matter as much. But just pointing out you could have a generic helper that would do this where `T: serde::Deserialize`

emostov: but i guess in this case we would lose the more accurate error messages

tvc/src/commands/keys/re_encrypt_share.rs
```diff
@@ -0,0 +1,491 @@
...
+}
+
+async fn build_re_encrypted_share_output(
+    quorum_key_metadata: &QuorumKeyMetadata,
+    provision_bundle: &ProvisionBundle,
+    operator_pair: &dyn Pair,
+    dangerous_skip_verification: bool,
+) -> anyhow::Result<ReEncryptedShareOutput> {
+    if !dangerous_skip_verification {
+        ensure_quorum_key_matches_manifest(quorum_key_metadata, provision_bundle)?;
```
emostov: i think we shouldn't skip this. Verification skipping is mostly for debug mode when the PCRs are zero'ed out, but i think you will always still want the correct quorum key

tvc/src/commands/keys/re_encrypt_share.rs
```diff
@@ -0,0 +1,491 @@
...
+    }
+
+    let ephemeral_public_key =
+        provision_bundle.ephemeral_public_key(dangerous_skip_verification)?;
```
emostov, line 90: one thing is we may want a way to not skip verification but pass in an override time for attestation docs that are more then 3 hours old. Doesn't need to be here but I think we should track in a ticket if not here

emostov, line 90: I'm wondering if we want to do manifest verification similar to approve manifest? Could also be a follow up, but at the very least I think you would want to make sure it has the manifest set you would expect since you are trusting those people to review the manifest

emostov:
Left some questions but overall really like it!

tvc/src/commands/keys/generate_quorum_key.rs
annieke: as in derive and store the address of each key? i'm assuming we will need to split into the encryption key address and signing key address too. would we use the addresses in tvc?

tvc/tests/keys_generate_quorum_key.rs
annieke: i think the model had a lot of paranoia about mixed-case keys and 0x prefixes, so this is a vestige of that

tvc/src/quorum_key_metadata.rs
annieke: yeah there is a lot of converting them back and forth, down to followup with this suggestion!

tvc/src/commands/keys/re_encrypt_share.rs
annieke: i'm down to make this a global arg, or remove this one in favor of say yubikey support after some feedback! this was pretty convenient for testing, and potentially useful if sharing an airgapped machine 👀

emostov: I think we can just leave as is for now if useful for testing. Also don't want to blow up scope here

tvc/src/commands/keys/generate_quorum_key.rs
annieke: i THINK this is included in the second commit but let me know if i'm using the dupe types somehow still

tvc/src/quorum_key_metadata.rs
annieke: TVC-19 created

tvc/src/commands/keys/re_encrypt_share.rs
annieke: created TVC-20 for this since i'd like to download more mental context on time override use cases before adding

tvc/src/commands/keys/re_encrypt_share.rs
annieke: i like this change! cleaner and we can always do with that for reading

tvc/src/commands/keys/re_encrypt_share.rs
annieke: do you mean the interactive review step from approve manifest? we do still compare the manifest envelope and the attestation doc, but i think you saw that and it's less explicit

if interactive i will create a followup issue!

tvc/src/provisioning.rs
annieke: i remember almost doing this in the last PR, but it required `aws_nitro_enclaves_nsm_api::api::AttestationDoc` as a type and we had to bump rust-sdk's `aws-nitro-enclaves-nsm-api` to 0.4 to use it. i'll bump it now if there's not a lot of churn, and ticket it if there is

tvc/src/provisioning.rs
annieke: hmm not here, but the comparison here basically just wraps qos_nsm's verify_attestation_doc_against_user_input which tests [PCR mismatches](https://github.com/tkhq/qos/blob/f897ce869d7fe69f048ec5c56242073ee8c62562/src/qos_nsm/src/nitro/mod.rs#L725) and [user_data mismatch ](https://github.com/tkhq/qos/blob/f897ce869d7fe69f048ec5c56242073ee8c62562/src/qos_nsm/src/nitro/mod.rs#L684).

tvc/src/commands/keys/re_encrypt_share.rs
emostov: Yes lets put a follow up issue in linear but make it low prio / good first issue type thing

tvc/Cargo.toml
```diff
@@ -21,6 +21,7 @@ turnkey_client = { workspace = true }
 turnkey_enclave_encrypt = { workspace = true }
 
 anyhow = { workspace = true }
+aws-nitro-enclaves-nsm-api = { version = "0.4", default-features = false }
```
emostov: Do we explicitly use this somewhere now?

annieke: Yup for the attestation doc type!

emostov: Its not significant since we already have the dep, but I just don't see it used - also like to specify dep version in workspace root and then inherit from the workspace in the crate

emostov (commented):
Just one last thing https://github.com/tkhq/rust-sdk/pull/131/changes/9e8fd48c73fd4de7aa84577578fc4394c054ec86#r3175385072

tvc/Cargo.toml
emostov: Oh i see your comment

emostov: ill approve

emostov (dismissed):

tvc/Cargo.toml
emostov: i have a pref for defining crate version in workspace but not worth blocking this PR since its otherwise all good so approved

annieke: Oh nah this is my b for missing, will fix if you can reapprove in 15

emostov (approved):

PR #132, opened 2026-05-05
feat(tvc): debug mode in app and deployment intents

daniilrrr:
## Summary

Debug-mode deployments expose enclave logs and zero out attestation PCRs, permanently marking an app's quorum key insecure. To make that risk explicit, debug-mode deployment requires a **two-step opt-in**: an app-level flag set at creation, plus a per-deployment flag that the server only accepts for capable apps. Attempting a debug-mode deployment on an app that isn't opted in will error from `mono`.

This is implemented in [mono PR 6314](https://github.com/tkhq/mono/pull/6314/).

This PR adds equivalent functionality to the Rust SDK with the updated protos/intents. The SDK config fields, env vars, and CLI args all use a unified `dangerous*` term to make the risk explicit; the proto intent fields (`enable_debug_mode_deployments`, `debug_mode`) keep their server-defined names.

### `tvc app create`

New flag:

| Flag | Env var | Default |
| --- | --- | --- |
| `--dangerous-enable-debug-mode-deployments` | `TVC_DANGEROUS_ENABLE_DEBUG_MODE_DEPLOYMENTS` | `false` |

New JSON config field:

```json
{
  "dangerousEnableDebugModeDeployments": false
}
```

> [!WARNING]
> Cannot be changed after the app is created. To move between debug-capable and non-debug, create a new app with a fresh quorum key.

---

### `tvc deploy create`

| Old | New |
| --- | --- |
| `--debug-mode` | `--dangerous-deploy-debug-mode` |
| `TVC_DEBUG_MODE` | `TVC_DANGEROUS_DEPLOY_DEBUG_MODE` |
| `debugMode` (config) | `dangerousDeployDebugMode` (config) |

`--debug-mode` was previously a no-op, so this breaking change is fine.

The server rejects debug-mode deployments for apps that weren't created with `--dangerous-enable-debug-mode-deployments`.

The JSON deploy config's `dangerousDeployDebugMode` field is a plain `bool` (defaults to `false` when omitted) rather than `bool | null`. Existing configs with `true`, `false`, or the field omitted keep working unchanged.

---

### Override order

Standard value flags follow the existing `clap` order:

```
CLI flag > env var > config file
```

**Exception — the debug flags are opt-in only.** As bare booleans, clap can't tell "flag absent" from "explicitly false", so they don't follow strict override order: passing the flag (or env var) turns debug mode **on**, but omitting it does **not** turn off debug mode enabled in the config. Debug mode is on if *either* source enables it; the flag can never disable it.

---

### Proto

Synced `proto/external/data/v1/tvc.proto` and `proto/immutable/activity/v1/activity.proto` from [mono PR #6314](https://github.com/tkhq/mono/pull/6314) to pick up the new `enable_debug_mode_deployments` field on `CreateTvcAppIntent` and `TvcApp`.

### DIFF
--- a/tvc/src/commands/app/create.rs
+++ b/tvc/src/commands/app/create.rs
@@ -147,13 +147,13 @@ fn build_create_tvc_app_intent(app_config: &AppConfig) -> CreateTvcAppIntent {
         share_set_id: app_config.share_set_id.clone(),
         share_set_params: share_set_params.as_ref().map(to_tvc_operator_set_params),
         enable_egress: app_config.external_connectivity,
-        enable_debug_mode_deployments: app_config.enable_debug_mode_deployments.into(),
+        enable_debug_mode_deployments: app_config.dangerous_enable_debug_mode_deployments.into(),
     }
 }
 
 fn apply_overrides(mut config: AppConfig, args: &Args) -> AppConfig {
     if args.dangerous_enable_debug_mode_deployments {
-        config.enable_debug_mode_deployments = args.dangerous_enable_debug_mode_deployments;
+        config.dangerous_enable_debug_mode_deployments = args.dangerous_enable_debug_mode_deployments;
     }
     config
 }
@@ -196,7 +196,7 @@ mod tests {
             }),
             share_set_id: None,
             share_set_params: None,
-            enable_debug_mode_deployments: false,
+            dangerous_enable_debug_mode_deployments: false,
         }
     }
 
@@ -244,12 +244,12 @@ mod tests {
         assert_eq!(intent.enable_debug_mode_deployments, Some(false));
     }
 
-    /// An explicit `enableDebugModeDeployments: true` in the config flows into
+    /// An explicit `dangerousEnableDebugModeDeployments: true` in the config flows into
     /// the intent so the server records the app's debug-mode capability.
     #[test]
     fn build_intent_forwards_debug_mode_from_config() {
         let mut config = valid_config();
-        config.enable_debug_mode_deployments = true;
+        config.dangerous_enable_debug_mode_deployments = true;
 
         let intent = build_create_tvc_app_intent(&config);
         assert_eq!(intent.enable_debug_mode_deployments, Some(true));
@@ -266,7 +266,7 @@ mod tests {
         };
 
         let config = apply_overrides(config, &args);
-        assert!(config.enable_debug_mode_deployments);
+        assert!(config.dangerous_enable_debug_mode_deployments);
     }
 
     /// Omitting the CLI flag must NOT override a config that enables debug-mode
@@ -275,14 +275,14 @@ mod tests {
     #[test]
     fn absent_dangerous_flag_preserves_config_debug_mode() {
         let mut config = valid_config();
-        config.enable_debug_mode_deployments = true;
+        config.dangerous_enable_debug_mode_deployments = true;
         let args = Args {
             config_file: PathBuf::new(),
             dangerous_enable_debug_mode_deployments: false,
         };
 
         let config = apply_overrides(config, &args);
-        assert!(config.enable_debug_mode_deployments);
+        assert!(config.dangerous_enable_debug_mode_deployments);
     }
 
     #[test]
--- a/tvc/src/commands/app/create.rs
+++ b/tvc/src/commands/app/create.rs
@@ -148,7 +148,7 @@ fn build_create_tvc_app_intent(app_config: &AppConfig) -> CreateTvcAppIntent {
         share_set_id: app_config.share_set_id.clone(),
         share_set_params: share_set_params.as_ref().map(to_tvc_operator_set_params),
         enable_egress: app_config.external_connectivity,
-        enable_debug_mode_deployments: Some(app_config.enable_debug_mode_deployments),
+        enable_debug_mode_deployments: app_config.enable_debug_mode_deployments.into(),
     }
 }
 
--- a/tvc/src/commands/app/create.rs
+++ b/tvc/src/commands/app/create.rs
@@ -22,6 +22,10 @@ pub struct Args {
     /// secure-enclave logs and emit zero'd attestation PCRs, so remote
     /// attestation cannot succeed. Cannot be changed after app creation; setting
     /// this true means the app's quorum key is considered permanently insecure.
+    ///
+    /// Opt-in only: passing this flag (or setting the env var) permits debug-mode
+    /// deployments even if the config file leaves it off, but omitting it does NOT
+    /// override a config file that enables it.
     #[arg(long, env = "TVC_DANGEROUS_ENABLE_DEBUG_MODE_DEPLOYMENTS")]
     pub dangerous_enable_debug_mode_deployments: bool,
 }
@@ -152,7 +156,9 @@ fn build_create_tvc_app_intent(app_config: &AppConfig) -> CreateTvcAppIntent {
 }
 
 fn apply_overrides(mut config: AppConfig, args: &Args) -> AppConfig {
-    config.enable_debug_mode_deployments = args.dangerous_enable_debug_mode_deployments;
+    if args.dangerous_enable_debug_mode_deployments {
+        config.enable_debug_mode_deployments = args.dangerous_enable_debug_mode_deployments;
+    }
     config
 }
 
@@ -261,6 +267,22 @@ mod tests {
         let args = Args {
             config_file: PathBuf::new(),
             dangerous_enable_debug_mode_deployments: true,
+        };
+
+        let config = apply_overrides(config, &args);
+        assert!(config.enable_debug_mode_deployments);
+    }
+
+    /// Omitting the CLI flag must NOT override a config that enables debug-mode
+    /// deployments: the flag is opt-in only and can never turn it off, so a
+    /// `true` config survives an absent flag.
+    #[test]
+    fn absent_dangerous_flag_preserves_config_debug_mode() {
+        let mut config = valid_config();
+        config.enable_debug_mode_deployments = true;
+        let args = Args {
+            config_file: PathBuf::new(),
+            dangerous_enable_debug_mode_deployments: false,
         };
 
         let config = apply_overrides(config, &args);
--- a/tvc/src/commands/app/create.rs
+++ b/tvc/src/commands/app/create.rs
@@ -22,10 +22,6 @@ pub struct Args {
     /// secure-enclave logs and emit zero'd attestation PCRs, so remote
     /// attestation cannot succeed. Cannot be changed after app creation; setting
     /// this true means the app's quorum key is considered permanently insecure.
-    ///
-    /// Opt-in only: passing this flag (or setting the env var) permits debug-mode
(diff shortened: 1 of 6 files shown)


daniilrrr:
reminder to self to regenerate protos from `mono/main` once PR #6314 merges, to make sure no other changes are missed

tvc/src/commands/app/create.rs
```diff
@@ -17,11 +17,19 @@ pub struct Args {
...
+    /// this true means the app's quorum key is considered permanently insecure.
+    #[arg(long, env = "TVC_DANGEROUS_ENABLE_DEBUG_MODE_DEPLOYMENTS")]
+    pub dangerous_enable_debug_mode_deployments: bool,
 }
 
 /// Run the app create command.
 pub async fn run(args: Args) -> Result<()> {
-    let app_config = load_or_fill_app_config(&args.config_file).await?;
+    let mut app_config = load_or_fill_app_config(&args.config_file).await?;
+    apply_overrides(&mut app_config, &args);
```
richardpringle: bit of a nit

I'm not sure you need this extra function. I understand why you would want to have it if you had more than one override, but I think we should put that in there if and only if we have more overrides. 

In any case, I think it's better to keep the newly constructed `app_config` immutable and apply any overrides _inside_ the `apply_overrides` function.

tvc/src/commands/app/create.rs
```diff
@@ -140,9 +148,14 @@ fn build_create_tvc_app_intent(app_config: &AppConfig) -> CreateTvcAppIntent {
         share_set_id: app_config.share_set_id.clone(),
         share_set_params: share_set_params.as_ref().map(to_tvc_operator_set_params),
         enable_egress: app_config.external_connectivity,
+        enable_debug_mode_deployments: Some(app_config.enable_debug_mode_deployments),
```
richardpringle: This seems like a nit, but it's save me from so many extra edits when refactoring:

```suggestion
        enable_debug_mode_deployments: app_config.enable_debug_mode_deployments.into(),
```

Generally, I prefer being explicit over implicit, but `Into::into` is just so much better for code maintenance. If we decided to go from `Option(AppConfig)` to `AppConfig` (make it mandatory or whatever) you wouldn't even need to change the code.

tvc/src/commands/app/create.rs
```diff
@@ -220,6 +234,39 @@ mod tests {
...
+    /// An explicit `enableDebugModeDeployments: true` in the config flows into
+    /// the intent so the server records the app's debug-mode capability.
+    #[test]
+    fn build_intent_forwards_debug_mode_from_config() {
+        let mut config = valid_config();
+        config.enable_debug_mode_deployments = true;
+
+        let intent = build_create_tvc_app_intent(&config);
+        assert_eq!(intent.enable_debug_mode_deployments, Some(true));
+    }
```
richardpringle, line 256: I love simple tests

tvc/src/commands/deploy/create.rs
```diff
@@ -108,6 +106,19 @@ pub struct Args {
     /// override `pivotContainerEncryptedPullSecret` from the config file.
     #[arg(long, value_name = "PATH", env = "TVC_PIVOT_PULL_SECRET")]
     pub pivot_pull_secret: Option<PathBuf>,
+
+    /// Deploy in debug mode, which forwards secure enclave logs to the host and
```
richardpringle, line 112: I would add a `# WARNING` in this doc comment to draw attention to the fact that something "special" is going on here. 

Can put it somewhere that makes more sense, but I think it catches the eye:

Image

tvc/src/commands/deploy/create.rs
```diff
@@ -135,7 +146,7 @@ fn build_create_intent(
         pivot_args: deploy_config.pivot_args.clone(),
         expected_pivot_digest: deploy_config.expected_pivot_digest.clone(),
         pivot_container_encrypted_pull_secret,
-        debug_mode: deploy_config.debug_mode,
+        debug_mode: Some(deploy_config.debug_mode),
```
richardpringle:
```suggestion
        debug_mode: deploy_config.debug_mode.into(),
```

tvc/src/commands/deploy/create.rs
```diff
@@ -543,6 +551,30 @@ mod tests {
...
+    fn dangerous_debug_mode_flag_enables_debug_mode() {
+        let file = write_config(&file_config()); // file has debug_mode = false
+        let args = Args {
+            config_file: Some(file.path().to_path_buf()),
+            dangerous_deploy_debug_mode: true,
+            ..empty_args()
+        };
+        let resolved = run_resolve(&args).unwrap();
+        assert!(resolved.debug_mode);
+    }
```
richardpringle, line 587: Don't need to change this, but this is why I don't like to hide logic inside functions. `resolve_deploy_config` Should just take `Option<DeployConfig>` as an argument which is the result of `read_config_file`. When we flatten out our logic, we don't have to do things like write an actual file and read it in just to see if the file that's read in is properly overwritten.

Cargo.lock
```diff
@@ -3728,6 +3728,15 @@ dependencies = [
...
 ]
 
+[[package]]
+name = "thread_local"
+version = "1.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f60246a4944f24f6e018aa17cdeffb7818b76356965d03b07d6a9886e8962185"
+dependencies = [
+ "cfg-if",
+]
```
richardpringle, line 3738: Why did this change? I don't think there should be any `Cargo.lock` changes without `Cargo.toml` changes, right? At least they probably shouldn't be coupled with logic that doesn't use the changes.

richardpringle (dismissed):
Just want you to change the doc warning and the `Cargo.lock` change (unless it's needed for some reason)

Everything else is optional

Cargo.lock
emostov: +1 for reverting if possible

daniilrrr: not sure why it changed, will attempt to revert

daniilrrr: Richard fixed it in https://github.com/tkhq/rust-sdk/pull/147, so I will rebase and this should go away

tvc/src/commands/app/create.rs
daniilrrr: agree, done

tvc/src/commands/deploy/create.rs
daniilrrr: done

tvc/src/commands/app/create.rs
daniilrrr: I like the clarity and readability of `apply_overrides()` even though it is overkill for one update. i've changed it to be a one-liner with the resulting `config` immutable, which I think is what you meant

tvc/src/commands/deploy/create.rs
daniilrrr: agree that this should be cleaned up and flattened into a pipeline of more pure functions. Won't do in this PR

tvc/src/commands/deploy/create.rs
```diff
@@ -108,6 +106,21 @@ pub struct Args {
...
+    ///
+    /// # WARNING
+    ///
+    /// Only valid for apps created with `--dangerous-enable-debug-mode-deployments`.
+    /// Debug-mode deployments permanently mark the app's quorum key as insecure;
+    /// to return to a secure posture, create a new app with a fresh quorum key.
+    ///
+    /// Overrides `debugMode` in the config file when set.
+    #[arg(long, env = "TVC_DANGEROUS_DEPLOY_DEBUG_MODE")]
+    pub dangerous_deploy_debug_mode: bool,
```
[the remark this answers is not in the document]
daniilrrr: updated

tvc/src/commands/app/create.rs
```diff
@@ -17,11 +17,18 @@ pub struct Args {
     /// Path to the app configuration file (JSON).
     #[arg(short = 'c', long, value_name = "PATH", env = "TVC_APP_CONFIG")]
     pub config_file: PathBuf,
+
+    /// Permit debug-mode deployments for this app. Debug-mode deployments expose
+    /// secure-enclave logs and emit zero'd attestation PCRs, so remote
+    /// attestation cannot succeed. Cannot be changed after app creation; setting
+    /// this true means the app's quorum key is considered permanently insecure.
+    #[arg(long, env = "TVC_DANGEROUS_ENABLE_DEBUG_MODE_DEPLOYMENTS")]
+    pub dangerous_enable_debug_mode_deployments: bool,
```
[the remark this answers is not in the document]
daniilrrr: updated

daniilrrr:
sync'd latest protos from `main` including those in [mono#6314 ](https://github.com/tkhq/mono/pull/6314/)

tvc/src/config/app.rs
```diff
@@ -73,6 +78,7 @@ impl AppConfig {
             }),
             share_set_id: None,
             share_set_params: None,
+            enable_debug_mode_deployments: false,
```
emostov: nit: consider giving this a name that matches the CLI arg such that all configs, env vars, and cli args use the same term

emostov (dismissed):

emostov:
I reviewed everything. Left one nit, but feel free to merge as is 👍

tvc/src/commands/app/create.rs
```diff
@@ -17,11 +17,18 @@ pub struct Args {
...
 }
 
 /// Run the app create command.
 pub async fn run(args: Args) -> Result<()> {
-    let app_config = load_or_fill_app_config(&args.config_file).await?;
+    let app_config = apply_overrides(load_or_fill_app_config(&args.config_file).await?, &args);
```
richardpringle, line 31: 👍

tvc/src/commands/deploy/create.rs
```diff
@@ -108,6 +108,19 @@ pub struct Args {
     /// override `pivotContainerEncryptedPullSecret` from the config file.
     #[arg(long, value_name = "PATH", env = "TVC_PIVOT_PULL_SECRET")]
     pub pivot_pull_secret: Option<PathBuf>,
+
+    /// Deploy in debug mode, which forwards secure enclave logs to the host and
+    /// zeroes attestation PCRs. This defeats the purpose of a secure enclave,
+    /// so it should only be used to debug non-prod applications and view application
+    /// logs.
+    ///
+    /// # WARNING
```
richardpringle, line 117: 👍

richardpringle (dismissed):

tvc/src/config/app.rs
daniilrrr: we use the `DANGEROUS_` prefix in a few other places in the CLI, which is appropriate here

So do you think in this `AppConfig` it should also be called `dangerous_enable_debug_mode_deployments` ?

emostov: Yes - I think it makes sense to add the dangerous prefix here
My thinking is that we want to make it clear to agents that all these config options are semantically the same - which is probably most obvious by giving them the exact same name

emostov (approved):

PR #134, opened 2026-05-06
feat(tvc): programmatic `tvc deploy create`

daniilrrr:
Adding ENV+flag overrides for `tvc deploy create` command and ENV var overrides for authenticating. 

Allows you to run `tvc deploy create` from Github CI. You first need to run `tvc login` and save the values into Github settings as secrets or variables. Sensitive values like the private key should be Github Secrets. More [info](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow)

Sample command
```bash
TVC_ORG_ID=...                      \
TVC_API_KEY_PUBLIC=...              \
TVC_API_KEY_PRIVATE=...             \
TVC_APP_ID=...                      \
TVC_QOS_VERSION=...                 \
TVC_PIVOT_PATH=...                  \
TVC_PIVOT_IMAGE_URL=...             \
TVC_EXPECTED_PIVOT_DIGEST=...       \
  tvc deploy create
```

### DIFF
--- a/tvc/README.md
+++ b/tvc/README.md
@@ -23,6 +23,38 @@ Special rules:
 
 `tvc deploy create` accepts `--config-file` *or* the equivalent flags (`--app-id`, `--qos-version`, `--pivot-image-url`, `--pivot-path`, `--expected-pivot-digest`, plus optional fields). `tvc app create` and `tvc keys generate-quorum-key` require `--config-file` because their configs include nested arrays.
 
+## Authentication
+
+For **local use**, run `tvc login` once and the CLI will read `~/.config/turnkey/` thereafter.
+
+For **programmatic use** (GitHub Actions, etc.), set these four env vars to authenticate directly without touching disk:
+
+| Env | Source |
+|---|---|
+| `TVC_ORG_ID` | your Turnkey organization UUID |
+| `TVC_API_BASE_URL` | e.g. `https://api.turnkey.com` |
+| `TVC_API_KEY_PUBLIC` | hex-encoded compressed P256 public key |
+| `TVC_API_KEY_PRIVATE` | hex-encoded P256 private key |
+
+When all four are present, every command authenticates directly from env. Setting some but not all is rejected.
+
+The typical flow: run `tvc login` once locally to generate an API key, register the public key in the Turnkey dashboard, then store the values in your CI's secret store (e.g. `TVC_API_KEY_PRIVATE` as a GitHub Secret, the rest as GitHub Variables).
+
+### Example: CI deploy with no config file
+
+```bash
+TVC_ORG_ID=...                      \
+TVC_API_BASE_URL=...                \
+TVC_API_KEY_PUBLIC=...              \
+TVC_API_KEY_PRIVATE=...             \
+TVC_APP_ID=...                      \
+TVC_QOS_VERSION=...                 \
+TVC_PIVOT_PATH=...                  \
+TVC_PIVOT_IMAGE_URL=...             \
+TVC_EXPECTED_PIVOT_DIGEST=...       \
+  tvc deploy create
+```
+
 ## Usage
 
 ### Create an App
--- a/tvc/README.md
+++ b/tvc/README.md
@@ -27,16 +27,17 @@ Special rules:
 
 For **local use**, run `tvc login` once and the CLI will read `~/.config/turnkey/` thereafter.
 
-For **programmatic use** (GitHub Actions, etc.), set these four env vars to authenticate directly without touching disk:
+For **programmatic use** (GitHub Actions, etc.), set these three env vars to authenticate directly without touching disk:
 
 | Env | Source |
 |---|---|
 | `TVC_ORG_ID` | your Turnkey organization UUID |
-| `TVC_API_BASE_URL` | e.g. `https://api.turnkey.com` |
 | `TVC_API_KEY_PUBLIC` | hex-encoded compressed P256 public key |
 | `TVC_API_KEY_PRIVATE` | hex-encoded P256 private key |
 
-When all four are present, every command authenticates directly from env. Setting some but not all is rejected.
+`TVC_API_BASE_URL` is optional and defaults to `https://api.turnkey.com`.
+
+When all three required vars are present, every command authenticates directly from env. Setting some but not all is rejected.
 
 The typical flow: run `tvc login` once locally to generate an API key, register the public key in the Turnkey dashboard, then store the values in your CI's secret store (e.g. `TVC_API_KEY_PRIVATE` as a GitHub Secret, the rest as GitHub Variables).
 
@@ -44,7 +45,6 @@ The typical flow: run `tvc login` once locally to generate an API key, register 
 
 ```bash
 TVC_ORG_ID=...                      \
-TVC_API_BASE_URL=...                \
 TVC_API_KEY_PUBLIC=...              \
 TVC_API_KEY_PRIVATE=...             \
 TVC_APP_ID=...                      \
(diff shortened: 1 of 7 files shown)


tvc/src/cli.rs
```diff
@@ -3,9 +3,27 @@
...
+  4. Built-in default
+
+Special rules:
+  --pivot-args replaces the config file's list entirely (does not append)
+
+Authentication:
+  Local: run `tvc login` once; commands then read ~/.config/turnkey/.
+  CI:    set TVC_ORG_ID, TVC_API_BASE_URL, TVC_API_KEY_PUBLIC, and
+         TVC_API_KEY_PRIVATE env vars to authenticate without files.
+         Setting some but not all four will fail to avoid unexpected operations";
```
emostov: The wording here is a little hard for me to grok
```suggestion
         Setting some but not all four will have undefined behavour.";
```
Or alternatively can just say it won't work

emostov: I think base url should not be required. 99.99% of people will only ever use the public api url

tvc/src/commands/deploy/create.rs
```diff
@@ -66,28 +103,76 @@ fn pin_image_url(image_url: &str, resolved_digest: &str) -> String {
...
+            let content = std::fs::read_to_string(path)
+                .with_context(|| format!("failed to read config file: {}", path.display()))?;
+            serde_json::from_str(&content)
+                .with_context(|| format!("failed to parse config file: {}", path.display()))?
+        }
+        None => {
+            let mut t = DeployConfig::template(None);
+            // Strip the template's "<REMOVE_ME...>" hint so flag-only mode
+            // doesn't ship it to the API for public images.
+            t.pivot_container_encrypted_pull_secret = None;
```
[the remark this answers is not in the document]
daniilrrr: this is fine, will be addressed in subsequent PRs

tvc/src/commands/deploy/create.rs
```diff
@@ -66,28 +103,76 @@ fn pin_image_url(image_url: &str, resolved_digest: &str) -> String {
...
+    if let Some(v) = args.public_ingress_port {
+        config.public_ingress_port = v;
+    }
+}
+
+fn resolve_deploy_config(args: &Args) -> Result<DeployConfig> {
+    let mut config = match &args.config_file {
+        Some(path) => {
+            let content = std::fs::read_to_string(path)
+                .with_context(|| format!("failed to read config file: {}", path.display()))?;
```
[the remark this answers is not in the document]
daniilrrr: added some simple precedence unit tests

tvc/src/cli.rs
daniilrrr: that makes sense. I'll have it default to the public API URL but still be overridable

tvc/README.md
```diff
@@ -8,6 +8,53 @@ CLI for [Turnkey Verifiable Cloud](https://turnkey.com) - see [this guide](https
...
+2. Environment variable (e.g. `TVC_APP_ID`)
+3. Config file value (`--config-file`)
+4. Built-in default
+
+Special rules:
+
+- `--pivot-args` replaces the config file's list entirely (does not append).
+
+`tvc deploy create` accepts `--config-file` *or* the equivalent flags (`--app-id`, `--qos-version`, `--pivot-image-url`, `--pivot-path`, `--expected-pivot-digest`, plus optional fields). `tvc app create` and `tvc keys generate-quorum-key` require `--config-file` because their configs include nested arrays.
+
```
emostov: we should remove this section and instead just say to check the help command to see if supports env vars/command line flags

tvc/README.md
```diff
@@ -8,6 +8,53 @@ CLI for [Turnkey Verifiable Cloud](https://turnkey.com) - see [this guide](https
...
+TVC_ORG_ID=...                      \
+TVC_API_KEY_PUBLIC=...              \
+TVC_API_KEY_PRIVATE=...             \
+TVC_APP_ID=...                      \
+TVC_QOS_VERSION=...                 \
+TVC_PIVOT_PATH=...                  \
+TVC_PIVOT_IMAGE_URL=...             \
+TVC_EXPECTED_PIVOT_DIGEST=...       \
+  tvc deploy create
+```
```
emostov: remove this section

tvc/src/commands/deploy/create.rs
```diff
@@ -1,21 +1,58 @@
...
+use anyhow::{bail, Context, Result};
 use clap::Args as ClapArgs;
 use std::path::PathBuf;
 use std::time::{SystemTime, UNIX_EPOCH};
 use turnkey_client::generated::{CreateTvcDeploymentIntent, ValidateTvcImageRequest};
 
-/// Create a new TVC deployment from a config file.
+/// Create a new TVC deployment from a config file or CLI flags.
 #[derive(Debug, ClapArgs)]
 #[command(about, long_about = None)]
```
emostov, line 49: TODO: update the long about to explain precedence and how to use this with/without env vars, cli arg overrides, and config file

emostov: This should probably include example usage with env vars as well. Instead of putting that info in a readme it should go here so its easy for agents to discover how to use this command

we should also update deploy init about to explain that you can skip init if you just use args/env vars

tvc/src/commands/deploy/create.rs
```diff
@@ -66,28 +103,76 @@ fn pin_image_url(image_url: &str, resolved_digest: &str) -> String {
...
-
-    // Validate config
-    if deploy_config.has_placeholders() {
-        anyhow::bail!(
-            "Config file contains placeholder values (<FILL_IN_...>). \
-             Please edit {} and fill in all required values.",
-            args.config_file.display()
+fn apply_overrides(config: &mut DeployConfig, args: &Args) {
+    if let Some(v) = &args.app_id {
+        config.app_id = v.clone();
```
emostov, line 143: It would be nice to have a verbose mode that printed the overrides and how precedents are overriding each other

tvc/README.md
emostov: something like this should only be in a help command text

tvc/README.md
```diff
@@ -8,6 +8,53 @@ CLI for [Turnkey Verifiable Cloud](https://turnkey.com) - see [this guide](https
...
+
+| Env | Source |
+|---|---|
+| `TVC_ORG_ID` | your Turnkey organization UUID |
+| `TVC_API_KEY_PUBLIC` | hex-encoded compressed P256 public key |
+| `TVC_API_KEY_PRIVATE` | hex-encoded P256 private key |
+
+`TVC_API_BASE_URL` is optional and defaults to `https://api.turnkey.com`.
+
+When all three required vars are present, every command authenticates directly from env. Setting some but not all is rejected.
```
emostov: remove this line

tvc/README.md
emostov: we might want to remove config precedence section all together and just keep in about for commands

tvc/src/commands/deploy/create.rs
```diff
@@ -66,28 +103,76 @@ fn pin_image_url(image_url: &str, resolved_digest: &str) -> String {
...
-        )
-    })?;
-
-    // Validate config
-    if deploy_config.has_placeholders() {
-        anyhow::bail!(
-            "Config file contains placeholder values (<FILL_IN_...>). \
-             Please edit {} and fill in all required values.",
-            args.config_file.display()
+fn apply_overrides(config: &mut DeployConfig, args: &Args) {
```
emostov, line 141: rename to apply_command_line_arg_overrides

annieke: ^did you decide against this?

annieke (approved):
lgtm and well-documented!

PR #135, opened 2026-05-07
chore(tvc): allow non-default app shareSets

annieke:
This change enables users to specify a new shareSet when creating an app (instead of always defaulting to the known shares). It also allows custom specified share-sets to be approved (instead of gating on only known shares). When a share set isn't specified, the cli still defaults to known shares. 

The second commit includes the straggler files from running `make fmt`.

### DIFF
--- a/tvc/src/commands/deploy/approve.rs
+++ b/tvc/src/commands/deploy/approve.rs
@@ -1,6 +1,5 @@
 //! Approve deploy command - cryptographically approve a QOS manifest.
 
-use crate::config::app::KNOWN_SHARE_SET_KEYS;
 use crate::config::turnkey::Config;
 use crate::operator_key::load_operator_pair;
 use crate::util::{read_file_to_string, write_file};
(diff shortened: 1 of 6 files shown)


tvc/src/commands/deploy/approve.rs
```diff
@@ -307,40 +325,154 @@ fn review_manifest_set(set: &ManifestSet) -> anyhow::Result<()> {
...
+        let share_set = deployment_share_set
+            .ok_or_else(|| anyhow!("share set not found in deployment response"))?;
+        return expected_share_set_from_api_operator_set(share_set);
+    }
+
+    if let Some(path) = &args.app_config {
+        return expected_share_set_from_app_config(path).await;
+    }
+
+    expected_share_set_from_operator_set_params(&AppConfig::share_set_params())
```
emostov: Leaving this note for myself here to reason through this edge case - `AppConfig::share_set_params()` returns the dev share set params. I think this just maintains expected behavour but want to double check nothing weird

tvc/src/commands/deploy/approve.rs
```diff
@@ -380,5 +516,5 @@ async fn fetch_manifest_from_deploy(deploy_id: &str) -> anyhow::Result<(Manifest
 
     println!("✓ Manifest loaded (manifest_id: {})", tvc_manifest.id);
 
-    Ok((manifest, tvc_manifest.id))
+    Ok((manifest, tvc_manifest.id, share_set))
```
emostov: maybe a dq but why add the share set if the manifest already has it?

tvc/src/config/app.rs
```diff
@@ -91,5 +99,125 @@ impl AppConfig {
...
+
+        if self.manifest_set_id.is_some() && self.manifest_set_params.is_some() {
+            anyhow::bail!("Cannot specify both manifestSetId and manifestSetParams");
+        }
+        if self.manifest_set_id.is_none() && self.manifest_set_params.is_none() {
+            anyhow::bail!("Must specify either manifestSetId or manifestSetParams");
+        }
+        if self.share_set_id.is_some() && self.share_set_params.is_some() {
+            anyhow::bail!("Cannot specify both shareSetId and shareSetParams");
+        }
```
emostov, line 123:
```suggestion
        }
        // It is fine if both share set id and params are none since we support a default dev share set
```

tvc/src/commands/deploy/approve.rs
annieke: this is my b; i thought the logic was comparing this deployment share set to the the manifest share set for a sanity check that both are the same, but ~i think it ends up just comparing the deployment share set to itself.~ ok no again it does compare the fetched share set to manifest share set

i'm actually inclined to remove this expected_share_set comparison and model this more like the manifest set approval - does that feel right to you? i think the old check comparing to the known set made ai want to force a expected vs actual check, but we really just want eyes on the manifest

tvc/src/commands/deploy/approve.rs
annieke: down to rename this too; the defaulting logic is a bit more convoluted than i'd like looking at this after testing

tvc/src/config/app.rs
```diff
@@ -91,5 +99,125 @@ impl AppConfig {
...
+
+        if let Some(params) = &self.share_set_params {
+            if params.threshold < MIN_SHARE_SET_THRESHOLD {
+                anyhow::bail!(
+                    "shareSetParams.threshold must be >= {MIN_SHARE_SET_THRESHOLD}, got {}",
+                    params.threshold
+                );
+            }
```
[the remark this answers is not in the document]
annieke: to me this validation might make more sense in the server, lmk opinions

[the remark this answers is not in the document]
emostov: ya i think thats fine to not have here

emostov (approved):

PR #139, opened 2026-05-18
feat(tvc): interactive CLI mode for `tvc app ...` and `tvc deploy...`

daniilrrr:
Adds interactive mode for some TVC commands, and supports filling placeholders step by step. 

`delete` commands were untouched

AI generated content below: 
___
## Summary

Make `tvc` interactive in a TTY without breaking CI. Users in a terminal get walked through missing values; CI runners get a clear "flag X is required in non-interactive mode" error instead of a hang.

## What's new

**Prompts module (`src/prompts.rs`)**
- `inquire`-backed `text` / `required_text` / `confirm` / `select` / `password`
- `is_interactive()` — true only when stdin is a TTY and `TVC_NON_INTERACTIVE` is unset
- `bail_if_non_interactive(flag_hint)` — fails fast with a message naming the flag to set instead of stalling
- `confirm_or_bail(message, operation)` — wraps `confirm` and bails with `"operation cancelled by user: {operation}"` on No; used by all 5 `deploy approve` section confirmations

**Interactive walking in commands**
- `tvc login` — `Select` widget for org choice; `Text` for new org ID/alias; `Select` for API environment (prod/preprod/dev/local)
- `tvc deploy approve` — refuses to run without `--dangerous-skip-interactive` in non-interactive mode; uses `Select` when multiple saved operator IDs are available (was: silently picked the first)
- `tvc {app,deploy} init` — new `--interactive` flag walks every placeholder and writes a filled config instead of a template
- `tvc {app,deploy} create` — if the config file is missing or has `<FILL_IN_...>` placeholders, walks the gaps in a TTY and offers to save the filled config back. Non-interactive behavior is unchanged for required fields: bails with the list of missing flags

**Non-interactive guard**
- `TVC_NON_INTERACTIVE=1` forces non-interactive everywhere
- Commands that would prompt instead bail with `--<flag> is required in non-interactive mode (set --<flag> or run in a TTY without TVC_NON_INTERACTIVE)`
- `--interactive` on init commands conflicts with `TVC_NON_INTERACTIVE` and errors loudly
- The `<REMOVE_ME_IF_PIVOT_CONTAINER_URL_IS_PUBLIC>` sentinel in `deploy create` configs now bails non-interactively with a message that names the JSON field and the resolution flag, instead of silently shipping the placeholder to the API

## Dependencies

- `inquire = "0.9"` (runtime)
- `rexpect = "0.5"` (dev-only, for PTY tests)

Both added to workspace `Cargo.toml` and `tvc/Cargo.toml`.

## Testing
- build `cargo build -p tvc` locally and alias your tvc command via `alias tvc='./target/debug/tvc'`
- Run `tvc login`, `tvc deploy`, and `tvc app` commands

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,8 +92,7 @@ walkdir = { version = "2.5", default-features = false }
 # Environment and configuration
 dotenvy = { version = "0.15.0", default-features = false }
 clap = { version = "4.5", features = ["std", "derive", "help", "usage", "error-context"], default-features = false }
-inquire = { version = "0.9" }
-
+inquire = { version = "0.9", default-features = false, features = ["crossterm"] }
 # Development dependencies
 assert_cmd = { version = "2", default-features = false }
 predicates = { version = "3", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,10 +92,12 @@ walkdir = { version = "2.5", default-features = false }
 # Environment and configuration
 dotenvy = { version = "0.15.0", default-features = false }
 clap = { version = "4.5", features = ["std", "derive", "help", "usage", "error-context"], default-features = false }
+inquire = { version = "0.9" }
 
 # Development dependencies
 assert_cmd = { version = "2", default-features = false }
 predicates = { version = "3", default-features = false }
+rexpect = { version = "0.5", default-features = false }
 tempfile = { version = "3.19.1", default-features = false }
 signature = "2"
 wiremock = { version = "0.6", default-features = false }
--- a/tvc/src/commands/deploy/create.rs
+++ b/tvc/src/commands/deploy/create.rs
@@ -2,10 +2,12 @@
 
 use crate::client::build_client;
 use crate::config::deploy::DeployConfig;
+use crate::config::turnkey::Config;
+use crate::prompts;
 use crate::pull_secret::encrypt_pivot_pull_secret;
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use clap::Args as ClapArgs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::time::{SystemTime, UNIX_EPOCH};
 use turnkey_client::generated::{CreateTvcDeploymentIntent, ValidateTvcImageRequest};
 
@@ -174,36 +176,76 @@ fn apply_overrides(config: &mut DeployConfig, args: &Args) {
     }
 }
 
-fn resolve_deploy_config(args: &Args) -> Result<DeployConfig> {
-    let mut config = match &args.config_file {
-        Some(path) => {
-            let content = std::fs::read_to_string(path)
-                .with_context(|| format!("failed to read config file: {}", path.display()))?;
-            serde_json::from_str(&content)
-                .with_context(|| format!("failed to parse config file: {}", path.display()))?
+/// Read a config from the given path, returning the parsed DeployConfig and
+/// a flag indicating whether the file existed (interactive flow may treat
+/// missing files as "start from template" when --config-file was provided).
+fn read_config_file(path: &Path) -> Result<Option<DeployConfig>> {
+    match std::fs::read_to_string(path) {
+        Ok(content) => {
+            let config: DeployConfig = serde_json::from_str(&content)
+                .with_context(|| format!("failed to parse config file: {}", path.display()))?;
+            Ok(Some(config))
         }
+        Err(_) if prompts::is_interactive() => Ok(None),
+        Err(e) => Err(anyhow!(e))
+            .with_context(|| format!("failed to read config file: {}", path.display())),
+    }
+}
+
+async fn resolve_deploy_config(args: &Args) -> Result<DeployConfig> {
+    let (mut config, file_loaded) = match &args.config_file {
+        Some(path) => match read_config_file(path)? {
+            Some(c) => (c, true),
+            None => (DeployConfig::template(None), false),
+        },
         None => {
             let mut t = DeployConfig::template(None);
             // Strip the template's "<REMOVE_ME...>" hint so flag-only mode
             // doesn't ship it to the API for public images.
             t.pivot_container_encrypted_pull_secret = None;
-            t
+            (t, false)
         }
     };
 
     apply_overrides(&mut config, args);
 
-    let missing = config.missing_required_fields();
-    if !missing.is_empty() {
-        let suggestion = if args.config_file.is_some() {
-            "Edit the config file or override via flag/TVC_* env."
-        } else {
-            "Provide via flag, TVC_* env, or --config-file."
-        };
-        bail!(
-            "missing required values: {}. {suggestion}",
-            missing.join(", ")
-        );
+    // After flag/env overrides, anything still placeholder-shaped needs to
+    // come from somewhere. Non-interactive: bail with the flag-list. Interactive:
+    // walk prompts for the remaining holes.
+    if !config.missing_required_fields().is_empty() {
+        if !prompts::is_interactive() {
+            let missing = config.missing_required_fields();
+            let suggestion = if args.config_file.is_some() {
+                "Edit the config file or override via flag/TVC_* env."
+            } else {
+                "Provide via flag, TVC_* env, or --config-file."
+            };
+            bail!(
+                "missing required values: {}. {suggestion}",
+                missing.join(", ")
+            );
+        }
+
+        let saved_app_id = Config::load().await.ok().and_then(|c| c.get_last_app_id());
+        config = config.fill_interactively(saved_app_id.as_deref())?;
+
+        // If the user passed a --config-file and we walked any prompts (either
+        // because the file was missing or had placeholders), offer to write the
+        // filled config back so the next run is hands-free.
+        if let Some(path) = &args.config_file {
+            let prompt = if file_loaded {
+                format!("Save filled config to {}?", path.display())
+            } else {
+                format!("Write a new config file at {}?", path.display())
+            };
+            if prompts::confirm(&prompt, true)? {
+                let json =
+                    serde_json::to_string_pretty(&config).context("failed to serialize config")?;
+                std::fs::write(path, json)
+                    .with_context(|| format!("failed to write config file: {}", path.display()))?;
+                println!("Wrote {}", path.display());
+            }
+        }
     }
 
     Ok(config)
@@ -211,7 +253,7 @@ fn resolve_deploy_config(args: &Args) -> Result<DeployConfig> {
 
 /// Run the deploy create command.
 pub async fn run(args: Args) -> Result<()> {
-    let deploy_config = resolve_deploy_config(&args)?;
+    let deploy_config = resolve_deploy_config(&args).await?;
 
     println!("Creating deployment for app '{}'...", deploy_config.app_id);
 
@@ -365,6 +407,17 @@ mod tests {
         f
     }
 
+    // Resolution tests run inside the tokio runtime since resolve_deploy_config
+    // now reaches into Config::load(). They're wired to call .block_on() via
+    // a small helper.
+    fn run_resolve(args: &Args) -> Result<DeployConfig> {
+        tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .unwrap()
+            .block_on(resolve_deploy_config(args))
+    }
+
     #[test]
     fn flag_overrides_file_value() {
         let file = write_config(&file_config());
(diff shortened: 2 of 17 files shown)


daniilrrr:
## Example usage

### First-time setup: `tvc login`

```
$ tvc login
No organization configured.
You can find your Organization ID at: https://app.turnkey.com/dashboard/welcome

? Organization ID: a1b2c3d4-1234-5678-90ab-cdef01234567
? Organization alias: default
? Select Turnkey API URL:
> prod     (https://api.turnkey.com)
  preprod  (https://api.preprod.turnkey.com)
  dev      (https://api.dev.turnkey.com)
  local    (http://localhost:8080)

Selected org: default (a1b2c3d4-...)

Generating API key...

API Key Generated!
Public Key: 02d3ad9c...
Add this API key to your Turnkey dashboard:
  1. Go to https://app.turnkey.com/dashboard/users
  ...
Press Enter when done...

Verifying credentials...
Successfully logged in!
```

### Manual config walking: `tvc app init --interactive`

```
$ tvc app init --interactive
? App name: my-payments-app
? Manifest set name: payments-prod
? Operator 'operator-1' public key: (04abc...) ↵     # default offered from logged-in op key

Created app config: app.json

Run: tvc app create --config-file app.json
```

### Auto-recovery when `deploy create` finds placeholders

You ran `tvc deploy init` earlier but left some `<FILL_IN_...>` values in the file. Instead of bailing, `deploy create` walks the gaps and offers to save the filled config back:

```
$ tvc deploy create --config-file deploy.json
? App ID: app_xyz789
? QOS version: 0.6.1
? Pivot container image URL: ghcr.io/team/payments:v1.4.2
? Pivot path (inside container): /usr/local/bin/payments
? Expected pivot digest (sha256:...): sha256:abc123...
? Is the container image in a public registry? Yes
? Save filled config to deploy.json? Yes
Wrote deploy.json

Creating deployment for app 'app_xyz789'...
Deployment created successfully!

Deployment ID: tvc_dep_abc

Next steps:
  - Run `tvc deploy approve --deploy-id tvc_dep_abc` to approve the manifest
```

### Walking a manifest approval: `tvc deploy approve`

```
$ tvc deploy approve --deploy-id tvc_dep_abc

========================================
         MANIFEST APPROVAL
========================================

NAMESPACE
─────────────────────────────────────
  Name:       turnkey-prod
  Nonce:      42
  Quorum Key: 04451028...

? Approve namespace? Yes

ENCLAVE (AWS Nitro)
─────────────────────────────────────
  PCR0 (image):     a1b2c3...
  PCR1 (kernel):    d4e5f6...
  ...

? Approve enclave configuration? Yes
? Approve pivot binary? Yes
? Approve manifest set? Yes
? Approve share set? Yes

========================================
    ALL SECTIONS APPROVED
========================================

Approval posted successfully!
```

### Multiple saved operators: pick interactively

```
$ tvc deploy approve --manifest manifest.json --manifest-id mft_xyz
? Select approving operator:
> op_a1b2c3  (alice)
  op_d4e5f6  (bob)
  op_g7h8i9  (charlie)
```

Previously, this silently picked the first one. Pass `--operator-id` to skip the prompt.

### Non-interactive mode: fail fast, never hang

```
$ TVC_NON_INTERACTIVE=1 tvc login
Error: --org is required in non-interactive mode (set --org or run in a TTY without TVC_NON_INTERACTIVE)

$ TVC_NON_INTERACTIVE=1 tvc deploy approve --manifest manifest.json
Error: --dangerous-skip-interactive is required in non-interactive mode (set --dangerous-skip-interactive or run in a TTY without TVC_NON_INTERACTIVE)

$ TVC_NON_INTERACTIVE=1 tvc app init --interactive
Error: --interactive conflicts with TVC_NON_INTERACTIVE=1

# Pull-secret sentinel left in the config from `tvc deploy init`:
$ TVC_NON_INTERACTIVE=1 tvc deploy create -c deploy.json
Error: pivotContainerEncryptedPullSecret is placeholder. Set the field to null in the config file (public image), or pass --pivot-pull-secret <PATH> (private image).

# Rejecting a manifest section during approval:
$ tvc deploy approve --deploy-id tvc_dep_abc
... (walks NAMESPACE, ENCLAVE, PIVOT, then user says No)
Error: operation cancelled by user: approval
```

daniilrrr:
Separated the `ReplayHint` nice-to-have out from this PR. Working on it on this [branch](https://github.com/tkhq/rust-sdk/tree/daniil/interactive-cli-n03-replayhint)

daniilrrr:
if we can't live with adding the `inquire` dependency, maybe we could gate its usage and "interactive mode" generally behind an `interactive` Rust feature flag that we flag as "less-secure" than the flag-based/ ENV var based approach. 

tbh I kinda like this for code organization purposes as well, rather than the current layout with various `if is_interactive()` checks littered in the functions. Downside is that customers would need to build a different feature when building the sdk, but that might be ok since these would be non-prod and educational initial usages of the SDK! Prod and non-trivial usage will be non-interactive anyway 

it would be additional maintainability burden to maintain 2 code paths, interactive and non-interactive, for our commands tho

emostov:
https://docs.google.com/spreadsheets/d/1V_PogeHg-qN7g4pGZLZFbPJlAUPQTJ3jcGpMIFFSkno/edit?usp=sharing

tvc/src/commands/deploy/create.rs
```diff
@@ -174,44 +177,101 @@ fn apply_overrides(config: &mut DeployConfig, args: &Args) {
...
-        let suggestion = if args.config_file.is_some() {
-            "Edit the config file or override via flag/TVC_* env."
-        } else {
-            "Provide via flag, TVC_* env, or --config-file."
-        };
-        bail!(
-            "missing required values: {}. {suggestion}",
-            missing.join(", ")
-        );
+    let config_updated = resolve_placeholders(&mut config, args).await?;
```
emostov, line 218: Thinking through the case of someone deploying with just env vars. Assuming env vars are full, resolve_placeholders will not prompt the user since apply_overrides above will fill out the config and thus leave no placeholders (I think). So it looks like in that case we wouldn't need an explicit opt out of the interactive logic

However, if they do provide a config that maybe has some placeholders but then override with env vars/ args, I think offer_to_save_config will stall on a prompt - so their needs to be an obvious way to opt out of interactive

IMO we should be consistent and either require a flag to opt out of interactive or vice versa

tvc/src/config/deploy.rs
```diff
@@ -35,15 +42,59 @@ impl DeployConfig {
...
+    /// Walk the user through any placeholder fields and fill them in.
+    /// Non-placeholder fields are preserved unchanged so partial edits work.
+    ///
+    /// `saved_app_id` is offered as the default for the App ID prompt when set.
+    pub fn fill_interactively(&mut self, saved_app_id: Option<&str>) -> Result<()> {
+        if self.app_id.starts_with("<FILL_IN") {
+            self.app_id = prompts::required_text("App ID", saved_app_id)?;
+        }
+        if self.qos_version.starts_with("<FILL_IN") {
+            self.qos_version = prompts::required_text("QOS version", None)?;
```
emostov, line 61: out of scope: would be nice if we had an endpoint with supported QOS version and we could query it here. Or maybe we can just keep a hardcoded list in the CLI and have it default to most recent

tvc/src/commands/login.rs
```diff
@@ -4,13 +4,45 @@ use crate::config::turnkey::{
...
+            ApiEnv::Prod => API_BASE_URL_PROD,
+            ApiEnv::Preprod => API_BASE_URL_PREPROD,
+            ApiEnv::Dev => API_BASE_URL_DEV,
+            ApiEnv::Local => API_BASE_URL_LOCAL,
+        }
+    }
+}
+
+impl std::fmt::Display for ApiEnv {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
emostov, line 37: imo this shouldn't be an interactive option - its only really here for internal testing. Customers should always use prod

emostov:
Took a first pass at logic. There is a bit here so might take a sec. Getting started on dep review

tvc/src/config/deploy.rs
emostov: looks like this is pretty easy with the inquire selectable macro or select prompt. we should maintain enums for stuff like qos version. potentially we can define in proto so we are confident it will get updated here anytime we update protos from internal repos

Cargo.toml
```diff
@@ -90,10 +90,12 @@ walkdir = { version = "2.5", default-features = false }
 # Environment and configuration
 dotenvy = { version = "0.15.0", default-features = false }
 clap = { version = "4.5", features = ["std", "derive", "help", "usage", "error-context"], default-features = false }
+inquire = { version = "0.9" }
```
emostov: can we turn default features off to see if we can get rid of some transitive deps

tvc/tests/pty.rs
```diff
@@ -0,0 +1,122 @@
...
+    session.exp_string("ALL SECTIONS APPROVED").unwrap();
+    session.exp_string("\"signature\"").unwrap();
+    session.exp_eof().unwrap();
+}
+
+/// Rejecting at the third section (pivot) bails immediately with the exact
+/// "operation cancelled by user: approval" string and never reaches the
+/// manifest-set section.
+///
+/// Replaces the deleted `tests/deploy_approve.rs::approve_interactive_reject`.
```
emostov: we should remove stuff like this that references old code in the new code

emostov:
dependency audit complete

tvc/tests/pty.rs
daniilrrr: yeah my b this is AI slop I let slip in

tvc/src/config/deploy.rs
daniilrrr: made a ticket, feels DevX to me cc @richardpringle 
https://linear.app/turnkey/issue/TVC-64/devx-queryable-endpoint-for-supported-qos-version

Cargo.toml
daniilrrr: tested, we need the backend of `crossterm` but we do shed 2 unused crates: `fuzzy-matcher`, `thread_local`. `crossterm` still pulls in several transitive deps but at least it's something

emostov: yes - something is better then nothing. fwiw i also like to turn off default feature and then selectively re-enable

tvc/src/commands/login.rs
daniilrrr: we prompt in `main` [currently](https://github.com/tkhq/rust-sdk/blob/main/tvc/src/commands/login.rs#L165) and I think it's useful for internal dev. Agree that for GA this should be defaulted to Prod URL. 

IMO a solution is to create a follow up ticket for pre-GA to remove this selector while still allowing you to supply your own URL via flag or Env Var if necessary. Does that work for you?

tvc/src/commands/deploy/create.rs
daniilrrr: yep we've got the `TVC_NON_INTERACTIVE` flag which, if set, will error descriptively in these interactive prompts. every `is_interactive()` check looks at this flag. 

I've highlighted this in the `LONG_ABOUT`

tvc/src/commands/login.rs
emostov: ok yup done https://linear.app/turnkey/issue/SYS-96/default-tvc-cli-url-selection-to-prod

tvc/src/commands/deploy/create.rs
emostov: wdyt about making it an arg as well for consistency with our other env vars? adding it to the long about is great, but if it was also an arg it would be discoverable from the auto generated clap --help

non-blocking, but just something to think about

tvc/src/config/deploy.rs
```diff
@@ -35,15 +42,59 @@ impl DeployConfig {
...
             public_ingress_port: 3000,
         }
     }
 
+    /// Walk the user through any placeholder fields and fill them in.
+    /// Non-placeholder fields are preserved unchanged so partial edits work.
+    ///
+    /// `saved_app_id` is offered as the default for the App ID prompt when set.
+    pub fn fill_interactively(&mut self, saved_app_id: Option<&str>) -> Result<()> {
+        if self.app_id.starts_with("<FILL_IN") {
```
emostov, line 57: out of scope: would be nice if we could fetch app IDs so they could select

tvc/src/commands/deploy/create.rs
emostov: https://linear.app/turnkey/issue/TVC-66/make-non-interactive-mode-explicit-and-discoverable

emostov (approved):
approving with follow up in mind https://linear.app/turnkey/issue/TVC-66/make-non-interactive-mode-explicit-and-discoverable

PR #140, opened 2026-05-18
rust codegen to follow new activites.json format

hadrelandon:
resolve activity type from proto annotation instead of `activities.json`. previously rust looked up activities by route url and read activity type string from `activities.json`, now we read the `enum: ["ACTIVITY_TYPE_*"]` annotation directly from each request message and use it as the key into `activities.json` to find intent and result type

### DIFF
--- a/client/src/generated/immutable.activity.v1.rs
+++ b/client/src/generated/immutable.activity.v1.rs
@@ -3747,7 +3747,7 @@ pub struct SparkClaimLeaf {
 }
 #[derive(Debug)]
 /// Build a transfer: for each leaf, derive the new leaf private key via HD
-/// (SIGNING_HD on new_leaf_derivation), compute the claim tweak scalar
+/// (SigningLeaf on new_leaf_derivation), compute the claim tweak scalar
 /// = old_priv - new_priv (mod n), Feldman-split the tweak across operators,
 /// and ECIES-encrypt the new leaf private key to receiver_public_key as
 /// the per-leaf secret_cipher carried inside each per-operator package.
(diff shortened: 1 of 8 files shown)


client/src/generated/immutable.activity.v1.rs
```diff
@@ -3839,9 +3839,6 @@ pub struct SparkSignatureRequest {
     /// @inject_tag: validate:"required,dive"
     #[serde(default)]
     pub operator_commitments: ::prost::alloc::vec::Vec<SparkFrostCommitment>,
-    /// @inject_tag: validate:"omitempty"
-    #[serde(default)]
-    pub adaptor_public_key: ::core::option::Option<::prost::alloc::string::String>,
```
moeodeh3, line 3844: lets rebase and regenerate, I don't think this is supposed to be removed based on [this](https://github.com/tkhq/mono/blob/05a8c624df28b87b929dd8657c63613463b39ec4/proto/immutable/activity/v1/activity.proto#L6307)

it looks like it was added back as optional!

proto/external/activity/v1/activity.proto
```diff
@@ -1869,7 +1869,7 @@ message CreateTvcManifestApprovalsRequest {
   string type = 1 [
     (google.api.field_behavior) = REQUIRED,
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
-      enum: ["ACTIVITY_TYPE_APPROVE_TVC_DEPLOYMENT"]
+      enum: ["ACTIVITY_TYPE_CREATE_TVC_MANIFEST_APPROVALS"]
```
moeodeh3, line 1872: 🔥

moeodeh3 (commented):
LGTM, just rebase and regenerate and I can approve!

moeodeh3 (approved):
LGTM, lets wait on this weeks prod release to go live before merging this in!

emostov:
> LGTM, lets wait on this weeks prod release to go live before merging this in!

Should be fine to merge newer types. Generally just need to wait for types to go live before publishing release

PR #141, opened 2026-05-19
Add client-side debug logging to TVC CLI

richardpringle:
Adds client-side debug logging to the TVC CLI, focused on making auth/config issues easier to diagnose without changing normal CLI output.

Main changes:
* initialize `tracing-subscriber` with `env-filter`, disabled by default unless `RUST_LOG` is set
* add debug events around command dispatch, config load/save, auth source selection, key path checks, and Turnkey client construction
* keep secret material out of logs; debug output records presence/path/source metadata only

Validated manually with:
```bash
RUST_LOG=debug cargo run -p tvc -- login
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,8 @@ zeroize = { version = "1.8.1", default-features = false }
 # Error handling
 thiserror = { version = "2.0.12", default-features = false }
 anyhow = { version = "1.0", default-features = false }
+tracing = { version = "0.1.41", default-features = false, features = ["std"] }
+tracing-subscriber = { version = "0.3.23", default-features = false, features = ["ansi", "env-filter", "fmt"] }
 
 # Web/HTTP
 mime = { version = "0.3.17", default-features = false }
--- a/tvc/src/commands/login.rs
+++ b/tvc/src/commands/login.rs
@@ -23,6 +23,11 @@ pub struct Args {
 
 /// Run the login command.
 pub async fn run(args: Args) -> anyhow::Result<()> {
+    tracing::debug!(
+        org_arg_present = args.org.is_some(),
+        "running login command"
+    );
+
     // Load existing config
     let mut config = Config::load().await?;
 
@@ -75,11 +80,20 @@ async fn select_or_create_org(
     config: &mut Config,
     org_arg: Option<&str>,
 ) -> Result<(String, OrgConfig)> {
+    tracing::debug!(
+        org_arg_present = org_arg.is_some(),
+        configured_org_count = config.orgs.len(),
+        active_org = ?config.active_org,
+        "selecting organization"
+    );
+
     // If --org provided, try to find it by alias or ID
     if let Some(org) = org_arg {
         if let Some((alias, org_config)) = find_org(config, org) {
+            tracing::debug!(org_alias = %alias, "selected existing organization from argument");
             return Ok((alias.clone(), org_config.clone()));
         }
+        tracing::debug!("organization argument did not match configured organizations");
         bail!("Organization '{org}' not found. Run `tvc login` without --org to set up a new organization.");
     }
 
@@ -88,6 +102,7 @@ async fn select_or_create_org(
 
     if org_count == 0 {
         // No orgs configured - prompt for new org
+        tracing::debug!("no organizations configured; prompting for new organization");
         println!("No organization configured.");
         return prompt_for_new_org(config).await;
     }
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -45,6 +45,8 @@ zeroize = { version = "1.8.1", default-features = false }
 # Error handling
 thiserror = { version = "2.0.12", default-features = false }
 anyhow = { version = "1.0", default-features = false }
+tracing = { version = "0.1.41", default-features = false, features = ["std"] }
+tracing-subscriber = { version = "0.3.23", default-features = false, features = ["ansi", "env-filter", "fmt"] }
```
emostov, line 49: do debug logs get compiled in my default?

tvc/src/commands/login.rs
```diff
@@ -75,11 +80,20 @@ async fn select_or_create_org(
...
+        "selecting organization"
+    );
+
     // If --org provided, try to find it by alias or ID
     if let Some(org) = org_arg {
         if let Some((alias, org_config)) = find_org(config, org) {
+            tracing::debug!(org_alias = %alias, "selected existing organization from argument");
             return Ok((alias.clone(), org_config.clone()));
         }
+        tracing::debug!("organization argument did not match configured organizations");
```
daniilrrr, line 96: nit - this log and perhaps others will be unnecessary once #139 lands

tvc/src/commands/login.rs
```diff
@@ -75,11 +80,20 @@ async fn select_or_create_org(
     config: &mut Config,
     org_arg: Option<&str>,
 ) -> Result<(String, OrgConfig)> {
+    tracing::debug!(
```
daniilrrr, line 83: nit - not sure if we have a style preference to prefer qualified import paths or not, but i'd prefer `debug!()` here and throughout to reduce visual clutter

daniilrrr (approved):
👍

tvc/src/commands/login.rs
```diff
@@ -133,6 +153,7 @@ async fn prompt_for_new_org(config: &mut Config) -> Result<(String, OrgConfig)>
 
     // Prompt for API base URL
     let api_base_url = prompt_for_api_url()?;
+    tracing::debug!(org_alias = %alias, api_base_url = %api_base_url, "adding prompted organization");
```
emostov, line 156:
```suggestion
    tracing::debug!(org_alias = %alias, %api_base_url, "adding prompted organization");
```
i think these are equivalent but my suggestion is the shorthand

tvc/src/commands/login.rs
```diff
@@ -311,6 +342,8 @@ async fn verify_credentials(
     org_id: &str,
     api_base_url: &str,
 ) -> Result<WhoamiResult> {
+    tracing::debug!(api_base_url = %api_base_url, "verifying credentials with whoami");
```
emostov, line 345:
```suggestion
    tracing::debug!(%api_base_url, "verifying credentials with whoami");
```

tvc/src/commands/login.rs
emostov: I don't feel strongly though. I think you current approach is less brittle in the case of a variable rename

tvc/src/logging.rs
```diff
@@ -0,0 +1,19 @@
...
+/// `RUST_LOG`, for example `RUST_LOG=tvc=debug`, to enable structured debug
+/// diagnostics on stderr.
+pub fn init() {
+    let filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("off"));
+
+    tracing_subscriber::fmt()
+        .compact()
+        .with_env_filter(filter)
+        .with_target(true)
+        .with_writer(std::io::stderr)
```
emostov, line 17: is this just to make sure the output doesn't get squashed?

emostov (approved):
Nice! Had some questions, but overall looks good and we can always make follow ups

tvc/src/logging.rs
richardpringle: Usually std for debug logs to come out of stderr and normal cli output to come out of stdout

Cargo.toml
richardpringle: The code gets compiled but the actual event isn't constructed [see here](https://docs.rs/tracing/latest/tracing/index.html#subscribers).

I don't know exactly how this works under the hood, but there are no allocations if the even isn't created.

tvc/src/commands/login.rs
richardpringle: Leaving it here for you to delete in #139. If it's here redundantly, 🤷, we can delete after

tvc/src/commands/login.rs
richardpringle: I merged the code, will do this in a follow up

tvc/src/commands/login.rs
richardpringle: I prefer your suggestion. Will do so in a follow up

PR #143, opened 2026-05-20
Follow up on #141

richardpringle:
Simple nit fixes

- **chore(tvc): Replace `tracing` calls with qualified path**
- **chore(tvc): Use tracing short hand where possible**


daniilrrr (approved):
thx

emostov (approved):

PR #146, opened 2026-05-27
Default login api-url to prod

richardpringle:
Linear: SYS-96

Summary:
- Default new `tvc login` orgs to prod without prompting for API URL.
- Add `--api-base-url` / `TVC_API_BASE_URL` as explicit override paths.
- Preserve existing org URLs unless an override is provided.

Tests:
- `cargo test -p tvc login`

### DIFF
--- a/tvc/src/commands/login.rs
+++ b/tvc/src/commands/login.rs
@@ -1,8 +1,7 @@
 //! Login command for authenticating with Turnkey.
 
 use crate::config::turnkey::{
-    Config, KeyCurve, OrgConfig, StoredApiKey, StoredQosOperatorKey, API_BASE_URL_DEV,
-    API_BASE_URL_LOCAL, API_BASE_URL_PREPROD, API_BASE_URL_PROD,
+    Config, KeyCurve, OrgConfig, StoredApiKey, StoredQosOperatorKey, API_BASE_URL_PROD,
 };
 use crate::prompts;
 use anyhow::{anyhow, bail, Context, Result};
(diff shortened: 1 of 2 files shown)


emostov (approved):

tvc/src/commands/login.rs
```diff
@@ -52,20 +20,29 @@ pub struct Args {
     /// If not provided, will prompt interactively.
     #[arg(long, env = "TVC_ORG")]
     pub org: Option<String>,
+    /// Turnkey API base URL. Defaults to production for newly configured orgs.
+    #[arg(long, env = "TVC_API_BASE_URL", value_name = "URL")]
+    pub api_base_url: Option<String>,
```
daniilrrr, line 25: nit: "BASE_URL" seems redundant to customers? 


```suggestion
    #[arg(long, env = "TVC_API_URL", value_name = "URL")]
    pub api_url: Option<String>,
```

tvc/src/commands/login.rs
```diff
@@ -52,20 +20,29 @@ pub struct Args {
...
     let mut config = Config::load().await?;
 
     // Select or create org
-    let (alias, org_config) = select_or_create_org(&mut config, args.org.as_deref()).await?;
+    let (alias, org_config) = select_or_create_org(
+        &mut config,
+        args.org.as_deref(),
+        args.api_base_url.as_deref(),
+    )
+    .await?;
```
daniilrrr, line 45: i think code and logic would be much clearer if we resolve the base_url here before going into the `select_or_create_org()` and nested logic. 

Some ideas:
- we could add a field/enum to `tvc::config::turnkey::Config` that defaults to `API_BASE_URL_PROD`, or the actual URL is fine too. Then, pass the `config` into `select_or_create_org()`. That seems clean
- if you dont want to do that, call `update_api_base_url_from_override()` here instead of within the func, and then you don't have to do it later

either approach reads cleaner to me than the nested `if let Some()` in current logic

daniilrrr (approved):
suggestion to do the Config resolution / override up front for clarity

in general, since we now do Clap flag > env > file overrides throughout the SDK, I think we should follow the pattern of 
1. load config
2. resolve config overrides + flags
3. pass around now-immutable config to functions 

I will also check my #132 to make sure I'm following this pattern

tvc/src/commands/login.rs
richardpringle: URL on its own can include paths, query-params, etc. BASE_URL implies that this is a full scheme + host.

tvc/src/commands/login.rs
richardpringle: 100% agree. I poked around and tried to make a small change, but a larger refactor in order:

https://linear.app/turnkey/issue/SYS-100/flatten-tvc-login-org-selection-and-api-base-url-override-flow

PR #147, opened 2026-05-29
Fix cargo lock

richardpringle:
Not sure why this got out of sync

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3729,6 +3729,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "thread_local"
+version = "1.1.9"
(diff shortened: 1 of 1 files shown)


richardpringle:
Got out of sync here: 8c8ed63

daniilrrr (approved):
not sure how this go off track either. Maybe I compiled locally with different options (dev? other?) and not `--release` ? Maybe we should add a `pre-commit` or `make lint` for this repo to mandate compiling locally identically to how we release ?

also, for this Rust-only repo, should we strictly require that all PRs be up-to-date with latest `main` before merging? There's not much PR submission throughput so I don't think it's much additional friction

r-n-o (approved):

PR #149, opened 2026-06-01
TVC-66: Make non-interactive mode explicit in tvc

richardpringle:
Linear: TVC-66

Summary:
- Add a global `--non-interactive` flag backed by `TVC_NON_INTERACTIVE`, using
  clap boolean parsing so falsey values do not force non-interactive mode.
- Thread a single `is_non_interactive: bool` into each prompt-capable command's
  `run(args, is_non_interactive)` entry point, which dispatches internally to
  mode-specific input builders feeding a shared mode-agnostic executor. Prompt
  helpers remain TTY UI primitives.
- Make non-interactive config/login paths fail fast without prompting or
  offering to save filled config, and document the behavior in help output.
- Replace `Vec<ValidationError>` + `is_empty()` checks with typed
  `DeployConfigValidationErrors` / `AppConfigValidationErrors` wrappers that
  expose `has_non_placeholder_error()` for the interactive prompt loop.
- Split `Args` overrides into `Overrides` substructs for `deploy create` and
  `app create` (clap `#[command(flatten)]`), so override fields are scoped
  separately from path/secret inputs.

Tests:
- `cargo fmt -p tvc -- --check`
- `cargo test -p tvc` (full suite; key files: `non_interactive.rs`,
  `config_validation.rs`, `deploy_create_help.rs`)

@tkhq/dependency-reviewers, I added `thiserror` as a dependency to `tvc`. That's
the only change, it was already in the dependency tree.

### DIFF
--- a/tvc/Cargo.toml
+++ b/tvc/Cargo.toml
@@ -30,6 +30,7 @@ inquire = { workspace = true }
 p256 = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+thiserror = { workspace = true }
 toml = { workspace = true }
 tokio = { workspace = true, features = ["fs"] }
 zeroize = { workspace = true }
--- a/tvc/src/cli.rs
+++ b/tvc/src/cli.rs
@@ -1,7 +1,7 @@
 //! CLI parsing and dispatch.
 
 use crate::commands;
-use clap::{Parser, Subcommand};
+use clap::{ArgAction, Parser, Subcommand, builder::BoolishValueParser};
 use tracing::debug;
 
 const LONG_ABOUT: &str = "\
@@ -26,12 +26,26 @@ Authentication:
   Local: run `tvc login` once; commands then read ~/.config/turnkey/.
   CI:    set TVC_ORG_ID, TVC_API_KEY_PUBLIC, and TVC_API_KEY_PRIVATE
          to authenticate without files. Env vars take precedence over local
-         config files. Setting some but not all three required vars will error.";
+         config files. Setting some but not all three required vars will error.
+
+Interactive behavior:
+  By default, commands may prompt when stdin is a TTY. Use --non-interactive
+  or set TVC_NON_INTERACTIVE=true to disable prompts and fail fast instead.";
 
 /// CLI command parsing and dispatch.
 #[derive(Debug, Parser)]
 #[command(about = "CLI for building with Turnkey Verifiable Cloud", long_about = LONG_ABOUT)]
 pub struct Cli {
+    /// Disable interactive prompts and fail fast when required values are missing.
+    #[arg(
+        long,
+        global = true,
+        env = "TVC_NON_INTERACTIVE",
+        action = ArgAction::SetTrue,
+        value_parser = BoolishValueParser::new()
+    )]
+    non_interactive: bool,
+
     #[command(subcommand)]
     command: Commands,
 }
@@ -40,27 +54,41 @@ impl Cli {
     /// Run the CLI.
     pub async fn run() -> anyhow::Result<()> {
         let args = Cli::parse();
-        debug!(command = args.command.name(), "dispatching");
+        debug!(
+            command = args.command.name(),
+            non_interactive = args.non_interactive,
+            "dispatching"
+        );
+
+        let non_interactive = args.non_interactive;
 
         match args.command {
             Commands::Deploy { command } => match command {
-                DeployCommands::Approve(args) => commands::deploy::approve::run(args).await,
+                DeployCommands::Approve(args) => {
+                    commands::deploy::approve::run(args, non_interactive).await
+                }
                 DeployCommands::GetStatus(args) => commands::deploy::get_status::run(args).await,
                 DeployCommands::ProvisioningDetails(args) => {
                     commands::deploy::provisioning_details::run(args).await
                 }
                 DeployCommands::PostShare(args) => commands::deploy::post_share::run(args).await,
                 DeployCommands::Status(args) => commands::deploy::status::run(args).await,
-                DeployCommands::Create(args) => commands::deploy::create::run(args).await,
-                DeployCommands::Init(args) => commands::deploy::init::run(args).await,
+                DeployCommands::Create(args) => {
+                    commands::deploy::create::run(args, non_interactive).await
+                }
+                DeployCommands::Init(args) => {
+                    commands::deploy::init::run(args, non_interactive).await
+                }
                 DeployCommands::Delete(args) => commands::deploy::delete::run(args).await,
                 DeployCommands::Restore(args) => commands::deploy::restore::run(args).await,
             },
             Commands::App { command } => match command {
                 AppCommands::Status(args) => commands::app::status::run(args).await,
                 AppCommands::List(args) => commands::app::list::run(args).await,
-                AppCommands::Create(args) => commands::app::create::run(args).await,
-                AppCommands::Init(args) => commands::app::init::run(args).await,
+                AppCommands::Create(args) => {
+                    commands::app::create::run(args, non_interactive).await
+                }
+                AppCommands::Init(args) => commands::app::init::run(args, non_interactive).await,
                 AppCommands::SetLiveDeploy(args) => commands::app::set_live_deploy::run(args).await,
                 AppCommands::Delete(args) => commands::app::delete::run(args).await,
             },
@@ -75,7 +103,7 @@ impl Cli {
                     commands::keys::re_encrypt_share::run(args).await
                 }
             },
-            Commands::Login(args) => commands::login::run(args).await,
+            Commands::Login(args) => commands::login::run(args, non_interactive).await,
         }
     }
 }
--- a/tvc/src/commands/deploy/approve.rs
+++ b/tvc/src/commands/deploy/approve.rs
@@ -3,6 +3,7 @@
 use crate::config::turnkey::Config;
 use crate::operator_key::load_operator_pair;
 use crate::prompts;
+use crate::prompts::{bail_required_in_non_interactive, stdin_can_prompt};
 use crate::util::{read_file_to_string, write_file};
 use anyhow::{Context, anyhow, bail};
 use clap::{ArgGroup, Args as ClapArgs};
@@ -75,63 +76,190 @@ pub struct Args {
     pub skip_post: bool,
 }
 
-/// Run the approve deploy command.
-pub async fn run(args: Args) -> anyhow::Result<()> {
+struct PostApprovalPlan<'a, 'b> {
+    manifest_id: &'a str,
+    operator_id: &'b str,
+}
+
+struct PostTarget {
+    manifest_id: String,
+    operator_id: String,
+}
+
+struct ResolvedApproveInputs {
+    manifest: Manifest,
+    operator_seed: Option<PathBuf>,
+    approval_out: Option<PathBuf>,
+    dry_run: bool,
+    post_target: Option<PostTarget>,
+}
+
+pub async fn run(args: Args, is_non_interactive: bool) -> anyhow::Result<()> {
+    let inputs = if is_non_interactive {
+        build_inputs_non_interactive(args).await?
+    } else {
+        build_inputs_interactive(args).await?
+    };
+
+    run_with_resolved_inputs(inputs).await
+}
+
+async fn build_inputs_interactive(args: Args) -> anyhow::Result<ResolvedApproveInputs> {
     let do_prompt_user = !args.dangerous_skip_interactive;
 
-    // Guard: Bail fast before fetching the manifest if we cannot prompt the user
-    if do_prompt_user {
-        prompts::bail_if_non_interactive("--dangerous-skip-interactive")?;
+    // Guard: bail fast before fetching the manifest if review prompts are
+    // required but the caller has no TTY to answer them.
+    if do_prompt_user && !stdin_can_prompt() {
+        bail_required_in_non_interactive("--dangerous-skip-interactive")?;
     }
 
-    // Fetch manifest - track manifest_id if fetched from API
-    let (manifest, fetched_manifest_id) = match (&args.manifest, &args.deploy_id) {
-        (Some(path), _) => (read_manifest_from_path(path).await?, None),
-        (_, Some(deploy_id)) => {
-            let (manifest, manifest_id) = fetch_manifest_from_deploy(deploy_id).await?;
-            (manifest, Some(manifest_id))
-        }
-        (None, None) => bail!("a manifest source is required"),
-    };
+    let (manifest, fetched_manifest_id) = load_manifest(&args).await?;
 
     if do_prompt_user {
         interactive_approve(&manifest)?;
     }
 
-    if args.dry_run {
+    let post_target = if !args.dry_run && !args.skip_post {
+        let manifest_id = resolve_manifest_id(&args, fetched_manifest_id.as_deref())?;
+        let operator_id = match &args.operator_id {
+            Some(id) => id.clone(),
+            None => {
+                let saved_ids = load_saved_operator_ids().await?;
+                match &*saved_ids {
+                    [] => bail!(
+                        "--operator-id is required to post approval to API. \
+                         No saved operator IDs found. \
+                         Use --skip-post to only generate the approval locally."
+                    ),
+                    [id] => id.clone(),
+                    _ if stdin_can_prompt() => prompts::select(
+                        "Select approving operator",
+                        saved_ids.iter().map(String::as_str).collect::<Vec<_>>(),
+                    )?
+                    .to_string(),
+                    _ => bail!(
+                        "--operator-id is required to post approval to API when multiple saved operator IDs are available"
+                    ),
+                }
+            }
+        };
+        Some(PostTarget {
+            manifest_id,
+            operator_id,
+        })
+    } else {
+        None
+    };
+
+    Ok(ResolvedApproveInputs {
+        manifest,
+        operator_seed: args.operator_seed,
+        approval_out: args.approval_out,
+        dry_run: args.dry_run,
+        post_target,
+    })
+}
+
+async fn build_inputs_non_interactive(args: Args) -> anyhow::Result<ResolvedApproveInputs> {
+    if !args.dangerous_skip_interactive {
+        bail_required_in_non_interactive("--dangerous-skip-interactive")?;
+    }
+
+    let (manifest, fetched_manifest_id) = load_manifest(&args).await?;
+
+    let post_target = if !args.dry_run && !args.skip_post {
+        let manifest_id = resolve_manifest_id(&args, fetched_manifest_id.as_deref())?;
+        let operator_id = match &args.operator_id {
+            Some(id) => id.clone(),
+            None => {
+                let saved_ids = load_saved_operator_ids().await?;
+                match &*saved_ids {
+                    [] => bail!(
+                        "--operator-id is required to post approval to API. \
+                         No saved operator IDs found. \
+                         Use --skip-post to only generate the approval locally."
+                    ),
+                    [id] => id.clone(),
+                    _ => bail!(
+                        "--operator-id is required to post approval to API when multiple saved operator IDs are available"
+                    ),
+                }
+            }
+        };
+        Some(PostTarget {
+            manifest_id,
+            operator_id,
+        })
+    } else {
+        None
+    };
+
+    Ok(ResolvedApproveInputs {
+        manifest,
+        operator_seed: args.operator_seed,
+        approval_out: args.approval_out,
+        dry_run: args.dry_run,
+        post_target,
+    })
+}
+
+async fn run_with_resolved_inputs(inputs: ResolvedApproveInputs) -> anyhow::Result<()> {
+    if inputs.dry_run {
         println!("Dry run complete. No approval generated.");
         return Ok(());
     }
 
-    let pair: Box<dyn crate::pair::Pair> =
-        Box::new(load_operator_pair(args.operator_seed.as_deref()).await?);
+    let approval = sign_and_output(
+        inputs.operator_seed.as_deref(),
+        inputs.approval_out.as_deref(),
+        &inputs.manifest,
+    )
+    .await?;
 
-    let approval = generate_approval(pair, &manifest).await?;
+    if let Some(target) = inputs.post_target {
+        let plan = PostApprovalPlan {
+            manifest_id: target.manifest_id.as_str(),
+            operator_id: target.operator_id.as_str(),
+        };
+        post_approval_to_api(plan, &approval).await?;
+    }
+
+    Ok(())
+}
+
+async fn load_manifest(args: &Args) -> anyhow::Result<(Manifest, Option<String>)> {
+    match (&args.manifest, &args.deploy_id) {
+        (Some(path), _) => Ok((read_manifest_from_path(path).await?, None)),
+        (_, Some(deploy_id)) => {
+            let (manifest, manifest_id) = fetch_manifest_from_deploy(deploy_id).await?;
+            Ok((manifest, Some(manifest_id)))
+        }
+        (None, None) => bail!("a manifest source is required"),
+    }
+}
+
+async fn sign_and_output(
+    operator_seed: Option<&Path>,
+    approval_out: Option<&Path>,
+    manifest: &Manifest,
+) -> anyhow::Result<Approval> {
+    let pair: Box<dyn crate::pair::Pair> = Box::new(load_operator_pair(operator_seed).await?);
+
+    let approval = generate_approval(pair, manifest).await?;
     let json = serde_json::to_string_pretty(&approval)?;
 
-    // Write to file or stdout
-    if let Some(ref path) = args.approval_out {
+    if let Some(path) = approval_out {
         write_file(path, &json).await?;
         println!("Approval written to: {}", path.display());
     } else {
         println!("{json}");
     }
 
-    // Post to API if not skipped
-    if !args.skip_post {
-        post_approval_to_api(&args, &approval, fetched_manifest_id.as_deref()).await?;
-    }
-
-    Ok(())
+    Ok(approval)
 }
 
-async fn post_approval_to_api(
-    args: &Args,
-    approval: &Approval,
-    fetched_manifest_id: Option<&str>,
-) -> anyhow::Result<()> {
-    // Use fetched manifest_id (from --deploy-id) or fall back to --manifest-id arg
-    let manifest_id = fetched_manifest_id
+fn resolve_manifest_id(args: &Args, fetched_manifest_id: Option<&str>) -> anyhow::Result<String> {
+    fetched_manifest_id
         .map(|s| s.to_string())
         .or_else(|| args.manifest_id.clone())
         .ok_or_else(|| {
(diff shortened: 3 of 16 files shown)


docs/tvc-cli-prompt-flow.md
```diff
@@ -0,0 +1,608 @@
+# TVC CLI Prompt Flow
```
daniilrrr: remember to delete this file

tvc/Cargo.toml
```diff
@@ -30,6 +30,7 @@ inquire = { workspace = true }
 p256 = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+thiserror = { workspace = true }
```
daniilrrr, line 33: requires dep reviewer but `thiserror` is v commonly used, so I don't expect any issue

tvc/src/cli.rs
```diff
@@ -40,27 +54,53 @@ impl Cli {
...
+            non_interactive = args.non_interactive,
+            "dispatching"
+        );
+
+        let non_interactive = args.non_interactive;
 
         match args.command {
             Commands::Deploy { command } => match command {
-                DeployCommands::Approve(args) => commands::deploy::approve::run(args).await,
+                DeployCommands::Approve(args) => {
```
daniilrrr, line 67: for every command here that has an interactive and non-interactive version, let's do an explicit `match` 

```
DeployCommands::Approve(args) => {
                    match non_interactive {
                        true => {commands::deploy::approve::run_non_interactive(args).await}
                        false => {commands::deploy::approve::run_interactive(args).await}
                    }
                }
```

I think the current version is a little over-LLM'd and relies on the ordering of the commands to be correct at the expense of clarity.
___
a second point is that, at a glance, in order to make maintainability easier I believe all of these funcs with 2 versions should instead be 

```run_interactive(args) -> run(args: Args, is_non_interactive: bool)```

to not have to maintain 2 code paths for each interactive command.

I plan to argue that on a separate comment as I read the rest of the PR. We will see if I retain this view

File: tvc/src/config/app.rs [hunk 4]
```diff
@@ -196,6 +224,104 @@ impl AppConfig {
     }
 }
 
+fn collect_operator_set_placeholder_errors(
+    prefix: &'static str,
+    params: &OperatorSetParams,
+    errors: &mut Vec<AppConfigValidationError>,
+) {
+    if params.name.starts_with("<FILL_IN") {
```
daniilrrr, line 232: extract `"<FILL_IN"` as a constant to use everywhere and prevent fat finger bugs later

tvc/src/config/app.rs
```diff
@@ -155,32 +158,57 @@ impl AppConfig {
...
+                errors.push(AppConfigValidationError::ThresholdTooLow {
+                    field: "shareSetParams.threshold",
+                    minimum: MIN_SHARE_SET_THRESHOLD,
+                    actual: params.threshold,
+                });
             }
         }
 
-        Ok(())
+        AppConfigValidationErrors::from_vec(errors)
```
daniilrrr: this is clever but a little indirect. I understand why you did it. 

personal opinion but I don't like how it implies that errors will always be produced as opposed to ending with `Ok(())` in the happy path. This is a nit tho

i think for a little extra clarity you could rename here and in `deploy.rs`

```AppConfigValidationErrors::ok_or_errors(errors)```

File: tvc/src/config/app.rs [hunk 6]
```diff
@@ -155,32 +158,57 @@ impl AppConfig {
             })
     }
 
-    pub fn validate(&self) -> anyhow::Result<()> {
-        if self.has_placeholders() {
-            anyhow::bail!("config contains placeholder values (<FILL_IN_...)");
+    pub fn validate(&self) -> Result<(), AppConfigValidationErrors> {
```
daniilrrr, line 161: this is nice and unit testable with the `AppConfigValidationError` error types. really appreciate this

tvc/src/config/deploy.rs
```diff
@@ -123,6 +126,101 @@ impl DeployConfig {
     pub fn pull_secret_is_placeholder(&self) -> bool {
         self.pivot_container_encrypted_pull_secret.as_deref() == Some(PULL_SECRET_PLACEHOLDER)
     }
+
+    pub fn validate(&self) -> Result<(), DeployConfigValidationErrors> {
+        let mut errors = Vec::new();
+        if self.app_id.starts_with("<FILL_IN") {
+            errors.push(DeployConfigValidationError::Placeholder {
+                field: "--app-id",
```
daniilrrr: do the underscore version here (`app_id`) to be consistent with `app.rs/validate()` func, not the flag version

tvc/src/commands/login.rs
```diff
@@ -25,109 +26,137 @@ pub struct Args {
...
+        org_arg_present = args.org.is_some(),
+        api_base_url_override_present = args.api_base_url.is_some(),
+        "running non-interactive login command"
+    );
+
+    let org_query = match args.org {
+        Some(query) => query,
+        None => {
+            bail_required_in_non_interactive("--org")?;
+            unreachable!("bail_required_in_non_interactive always returns an error");
```
daniilrrr: icky unreachable

i have a way to kill this `unreachable!` that you may like

refactor the `bail_` to use a public inner helper 

```rust
pub fn bail_required_in_non_interactive(flag_hint: &str) -> Result<()> {
    Err(error_required_in_non_interactive(flag_hint))
}

pub fn error_required_in_non_interactive(flag_hint: &str) -> Error {
    anyhow!(
        "{flag_hint} is required in non-interactive mode \
         (set {flag_hint} or run in a TTY without --non-interactive \
         / {NON_INTERACTIVE_ENV}=true)"
    )
}

```

then call the inner helper here in a one-liner

```rust
    let org_query = args.org.ok_or_else(|| error_required_in_non_interactive("--org"))?;
```

tvc/src/commands/login.rs
```diff
@@ -25,109 +26,137 @@ pub struct Args {
...
+        }
+    };
+
+    let config = Config::load().await?;
+    let plan = LoginPlan {
+        org: OrgPlan::Existing(org_query),
+        api_base_url_override: args.api_base_url,
+        api_key_policy: ApiKeyPolicy::RequireExisting,
+    };
+    execute_login(config, plan).await
```
daniilrrr: fabulous refactoring job, really nice to centralize everything into shared code

tvc/src/commands/deploy/init.rs
```diff
@@ -23,14 +23,21 @@ pub struct Args {
...
+    }
+    run(args).await
+}
+
+pub async fn run_non_interactive(args: Args) -> Result<()> {
+    if args.interactive {
+        bail_interactive_conflicts_with_non_interactive()?;
     }
+    run(args).await
+}
```
daniilrrr: at the risk of overengineering, should we refactor these checks into the `prompts` package and do them at the beginning of every `run_interactive()` and `run_non_interactive()` version of EVERY command for safety? They're already duplicated between `deploy/init.rs` here and `app/init.rs`. Something like 

```rust
/// In interactive mode, the `--interactive` fill flag only requires a real TTY.
  pub fn require_tty_for_interactive(interactive: bool) -> Result<()> {
      if interactive {
          ensure_stdin_is_tty()?;
      }
      Ok(())
  }

  /// In non-interactive mode, the `--interactive` fill flag is contradictory.
  pub fn reject_interactive_flag(interactive: bool) -> Result<()> {
      if interactive {
          bail_interactive_conflicts_with_non_interactive()?;
      }
      Ok(())
  }

  Then both app/init.rs and deploy/init.rs collapse to:

  pub async fn run_interactive(args: Args) -> Result<()> {
      prompts::require_tty_for_interactive(args.interactive)?;
      run(args).await
  }

  pub async fn run_non_interactive(args: Args) -> Result<()> {
      prompts::reject_interactive_flag(args.interactive)?;
      run(args).await
  }
```

tvc/src/commands/deploy/approve.rs
```diff
@@ -103,93 +101,168 @@ pub async fn run(args: Args) -> anyhow::Result<()> {
         return Ok(());
     }
 
+    let approval = sign_and_output(&args, &manifest).await?;
+
+    if !args.skip_post {
```
daniilrrr: nit - reduce this long conditional and just return early instead. 

`    if args.skip_post { return Ok(()); } `

makes it easier for me to read and IMO, usually, any set of nested braces you can kill is a win

daniilrrr: once you return early, the shape is clearer to me and can be refactored to a flatter procedure without conditionals
1. try to resolve the `operator_id`, if not configured 
2. construct `PostApprovalPlan`
3. `post_approval_to_api()`

which reads nicely

daniilrrr: can also collapse operator resolution logic between here and the `run_non_interactive()` part

daniilrrr: LLM code below. because `resolve_operator_id()` is used by both paths we need to pass it a bool indicating the origin of the call, which is against the overall spirit of your PR, but it's preeeetty clean 

```rust
pub async fn run_interactive(args: Args) -> anyhow::Result<()> {
      let do_prompt_user = !args.dangerous_skip_interactive;

      // Guard: bail fast before fetching the manifest if review prompts are
      // required but the caller has no TTY to answer them.
      if do_prompt_user && !stdin_can_prompt() {
          bail_required_in_non_interactive("--dangerous-skip-interactive")?;
      }

      let (manifest, fetched_manifest_id) = load_manifest(&args).await?;

      if do_prompt_user {
          interactive_approve(&manifest)?;
      }

      if args.dry_run {
          println!("Dry run complete. No approval generated.");
          return Ok(());
      }

      let approval = sign_and_output(&args, &manifest).await?;

      if args.skip_post {
          return Ok(());
      }

      post_approval(
          &args,
          &approval,
          fetched_manifest_id.as_deref(),
          OperatorPrompt::Allow,
      )
      .await
  }

  pub async fn run_non_interactive(args: Args) -> anyhow::Result<()> {
      if !args.dangerous_skip_interactive {
          bail_required_in_non_interactive("--dangerous-skip-interactive")?;
      }

      let (manifest, fetched_manifest_id) = load_manifest(&args).await?;

      if args.dry_run {
          println!("Dry run complete. No approval generated.");
          return Ok(());
      }

      let approval = sign_and_output(&args, &manifest).await?;

      if args.skip_post {
          return Ok(());
      }

      post_approval(
          &args,
          &approval,
          fetched_manifest_id.as_deref(),
          OperatorPrompt::Disallow,
      )
      .await
  }

  And the helpers they rely on (place near PostApprovalPlan):

  /// Whether operator-id disambiguation may fall back to an interactive prompt.
  enum OperatorPrompt {
      Allow,
      Disallow,
  }

  async fn post_approval(
      args: &Args,
      approval: &Approval,
      fetched_manifest_id: Option<&str>,
      prompt: OperatorPrompt,
  ) -> anyhow::Result<()> {
      let manifest_id = resolve_manifest_id(args, fetched_manifest_id)?;
      let operator_id = resolve_operator_id(args, prompt).await?;

      let plan = PostApprovalPlan {
          manifest_id: &manifest_id,
          operator_id: &operator_id,
      };
      post_approval_to_api(plan, approval).await
  }

  async fn resolve_operator_id(args: &Args, prompt: OperatorPrompt) -> anyhow::Result<String> {
      if let Some(id) = &args.operator_id {
          return Ok(id.clone());
      }

      let saved_ids = load_saved_operator_ids().await?;
      match saved_ids.as_slice() {
          [] => bail!(
              "--operator-id is required to post approval to API. \
               No saved operator IDs found. \
               Use --skip-post to only generate the approval locally."
          ),
          [id] => Ok(id.clone()),
          _ if matches!(prompt, OperatorPrompt::Allow) && stdin_can_prompt() => Ok(prompts::select(
              "Select approving operator",
              saved_ids.iter().map(String::as_str).collect::<Vec<_>>(),
          )?
          .to_string()),
          _ => bail!(
              "--operator-id is required to post approval to API when multiple saved operator IDs are available"
          ),
      }
  }
```

daniilrrr (changes requested):
this is really nice, hard work that will make the `tvc` SDK much clearer to use and maintain. great job

i have some changes I would like to see throughout

daniilrrr:
PR description out of date, `InteractionMode` not in this PR

tvc/src/cli.rs
daniilrrr: having read the PR, you've done a nice job flattening the logic of each `run()` command into a (1) setup and config phase, then (2) post SDK API call phase. It is (1) that differs between interactive and non-interactive, while the rest is standard

To me, it's better for maintenance and discoverability to have a diamond shaped 
1. run()
  2a. setup_non_interactive(), or 
  2b. setup_interactive()
3. post()

let me know what you think

tvc/src/commands/login.rs
daniilrrr: not just here but this pattern of shared "core" logic in the other non/interactive commands is great

tvc/src/commands/deploy/approve.rs
richardpringle: @daniilrrr FYI, you can use ```rust to get syntax highlighting in comments

richardpringle: Part of the point of the whole PR is to remove the bool "can" prompt logic. I think it's a lot harder to maintain and I base that on my first pass at implementing this feature. There's conditional logic everywhere. That means that every helper function needs 2x the amount of tests to get full coverage. Instead, the ethos here is to flatten conditional logic as much as possible at the small cost of some duplication. IMO, the maintenance cost of the duplication is much much lower.

With all that said, you've made me realize that I'm missing a bunch of unit tests that would clearly emphasize my point a little better.

richardpringle:
```rust
      if let Some(id) = &args.operator_id {
          return Ok(id.clone());
      }
```

This is an interesting suggestion. Do I go overboard avoiding allocations? Absolutely. But I find it easier to write code where I minimize allocation by default. It makes code a lot easier to change when you do discover a hot-loop that's needlessly allocating. In a cli, optimizing for fewer allocations is almost certainly never going to happen, but if you generally avoid needlessly cloning, you never have to ask that question. I dont' think the cost is very high here (it's calling `post_approval_to_api` in each branch instead of once). 

If the function was more than 100 lines or something, some of these things could be factored out differently, but I think that's only worth it when the function starts getting longer.

daniilrrr: >  Instead, the ethos here is to flatten conditional logic as much as possible at the small cost of some duplication.

fair tradeoff. add this sentence to the PR to make it crystal clear what your goal is :) 

 > But I find it easier to write code where I minimize allocation by default. It makes code a lot easier to change when you do discover a hot-loop that's needlessly allocating.

that's a good pointer that I will try and incorporate in my future code, thanks
___
What I'm getting at is that in the other commands you have a super nice flattening of the commands into interactive / non interactive funcs to call and have removed the conditional logic, except here where we still hinge on `if let Some(id) = &args.operator_id { ... } else {...}`. I think it could be cleaner and more testable to to have `resolve_operator_id_interactive()` and `resolve_operator_id_noninteractive()` here. The LLM code did not correctly illustrate my example. I'll paste something better in a few min, hopefully

daniilrrr: pretty flat logic example, curious what you think

LLM code below: 
```rust
pub async fn run_interactive(args: Args) -> anyhow::Result<()> {
    let do_prompt_user = !args.dangerous_skip_interactive;

    // Guard: bail fast before fetching the manifest if review prompts are
    // required but the caller has no TTY to answer them.
    if do_prompt_user && !stdin_can_prompt() {
        bail_required_in_non_interactive("--dangerous-skip-interactive")?;
    }

    let (manifest, fetched_manifest_id) = load_manifest(&args).await?;

    if do_prompt_user {
        interactive_approve(&manifest)?;
    }

    if args.dry_run {
        println!("Dry run complete. No approval generated.");
        return Ok(());
    }

    let approval = sign_and_output(&args, &manifest).await?;

    if args.skip_post { return Ok(()); }

    let manifest_id = resolve_manifest_id(&args, fetched_manifest_id.as_deref())?;
    let operator_id = resolve_operator_id_interactive().await?;

    let plan = PostApprovalPlan {
        manifest_id: manifest_id.as_str(),
        operator_id: &operator_id,
    };

    post_approval_to_api(plan, &approval).await?;

    Ok(())
}

pub async fn run_non_interactive(args: Args) -> anyhow::Result<()> {
    if !args.dangerous_skip_interactive {
        bail_required_in_non_interactive("--dangerous-skip-interactive")?;
    }

    let (manifest, fetched_manifest_id) = load_manifest(&args).await?;

    if args.dry_run {
        println!("Dry run complete. No approval generated.");
        return Ok(());
    }

    let approval = sign_and_output(&args, &manifest).await?;

    if args.skip_post { return Ok(()); }

    let manifest_id = resolve_manifest_id(&args, fetched_manifest_id.as_deref())?;
    let operator_id = resolve_operator_id_noninteractive().await?;

    let plan = PostApprovalPlan {
        manifest_id: manifest_id.as_str(),
        operator_id: &operator_id,
    };

    post_approval_to_api(plan, &approval).await?;

    Ok(())
}
```
...
and the 2 versions
```rust
async fn resolve_operator_id_interactive() -> anyhow::Result<String> {
    let saved_ids = load_saved_operator_ids().await?;
    match &*saved_ids {
        [] => bail!(
            "--operator-id is required to post approval to API. \
             No saved operator IDs found. \
             Use --skip-post to only generate the approval locally."
        ),
        [id] => Ok(id.clone()),
        _ if stdin_can_prompt() => Ok(prompts::select(
            "Select approving operator",
            saved_ids.iter().map(String::as_str).collect::<Vec<_>>(),
        )?
        .to_string()),
        _ => bail!(
            "--operator-id is required to post approval to API when multiple saved operator IDs are available"
        ),
    }
}

async fn resolve_operator_id_noninteractive() -> anyhow::Result<String> {
    let saved_ids = load_saved_operator_ids().await?;
    match &*saved_ids {
        [] => bail!(
            "--operator-id is required to post approval to API. \
             No saved operator IDs found. \
             Use --skip-post to only generate the approval locally."
        ),
        [id] => Ok(id.clone()),
        _ => bail!(
            "--operator-id is required to post approval to API when multiple saved operator IDs are available"
        ),
    }
}
```

tvc/src/commands/deploy/init.rs
richardpringle: 🤔, I don't think this is overengineering, this is a good callout so we don't have to remember to add the checks to future commands

tvc/src/commands/login.rs
richardpringle: Now this is dumb LLM code. I can just use an `else-return` here.

tvc/src/config/app.rs
richardpringle: Yeah... but did "I" add unit tests haha

tvc/src/config/app.rs
richardpringle: >This is a nit tho

This one was hand-written and I definitely got the ick writing it haha. It was better than always having to check for `is_empty` every time though. 

Your name gets rid of the ick haha, thanks!

tvc/src/config/app.rs
richardpringle: I'll probably leave this one in this PR and do a single pass of "floating" numbers are strings in the code, everything that is a constant should be defined as such (effectively just named constants). There are a bunch more below too.

tvc/src/cli.rs
richardpringle: The goal here was to flatten everything out as much as possible. I even added the guards instead of the if-statements the LLM produced (matching on a bool will fail clippy 😉). I honestly prefer a single match statement with no nested logic, the nature if guards and wildcard matches means that order matters on every match statement. 

With that said, I don't mind adding the if-statements back. 

As far as your suggestion of `setup_(non_)interavtive`, I'm not sure I see how that changes things. I think this one is worth hopping on a call to talk about so I can understand the benefit of what you're suggesting

tvc/Cargo.toml
richardpringle: Yeah, it's already a dep in the workspace so I'm not actually adding anything new, just exposing it to `tvc`.

tvc/src/cli.rs
richardpringle: Just capturing what was said in our call: 
The setup fucntions should return the proper config and plan, that config and plan should get passed to the api calls, like `execute_login`. 

Also, setup can be the place that switches on interactive vs non-interactive

daniilrrr: yeah one version could be 

1. run(is_interactive)
  2a. setup_non_interactive() -> PostArgs 
  2b. setup_interactive() -> PostArgs
  3. post(PostArgs) 

where you declare the plumbing with the returned types

tvc/src/commands/deploy/init.rs
daniilrrr: last time I looked at wiring in "Setup" funcs in Rust as part of the type or interface, it was a little hairy and not super elegant. But I think we could have a `InteractiveWrapper` and `NonInteractiveWrapper` that auto-runs these so you can't forget them. This may be premature for the number of commands we have, BUT a nice way to never forget these checks for future functions. Maybe something for a follow-up ticket

LLM code:
```
// The core logic (implements the interface)
  pub struct CoreLogic;

  impl Service for CoreLogic {
      fn execute(&self) {
          println!("CoreLogic: Running main operation.");
      }
  }

  // Interactive wrapper — runs prechecks, then the wrapped function
  pub struct InteractiveWrapper<T: Service> {
      inner: T,
  }

  impl<T: Service> InteractiveWrapper<T> {
      pub fn new(inner: T) -> Self {
          Self { inner }
      }

      // Prechecks we run before the wrapped operation
      fn prechecks(&self) {
          let now = Instant::now();
          println!("InteractiveWrapper: Running prechecks at {:?}", now);
      }
  }

  impl<T: Service> Service for InteractiveWrapper<T> {
      fn execute(&self) {
          self.prechecks();       // Runs prechecks
          self.inner.execute();   // Runs the wrapped function
      }
  }

  // Non-interactive wrapper — same shape, different precheck behavior
  pub struct NonInteractiveWrapper<T: Service> {
      inner: T,
  }

  impl<T: Service> NonInteractiveWrapper<T> {
      pub fn new(inner: T) -> Self {
          Self { inner }
      }

      // Prechecks we run before the wrapped operation
      fn prechecks(&self) {
          let now = Instant::now();
          println!("NonInteractiveWrapper: Running prechecks at {:?}", now);
      }
  }

  impl<T: Service> Service for NonInteractiveWrapper<T> {
      fn execute(&self) {
          self.prechecks();       // Runs prechecks
          self.inner.execute();   // Runs the wrapped function
      }
  }

  fn main() {
      // Interactive path
      let interactive = InteractiveWrapper::new(CoreLogic);
      interactive.execute();

      // Non-interactive path
      let non_interactive = NonInteractiveWrapper::new(CoreLogic);
      non_interactive.execute();
  }
```

daniilrrr: likely there is a more elegant way to do this but "hooking in" setup funcs would be nice here to idiot-proof the SDK

tvc/src/config/app.rs
richardpringle: Created a ticket for this, resolving

tvc/src/cli.rs
richardpringle: done

tvc/src/commands/deploy/approve.rs
```diff
@@ -75,121 +76,229 @@ pub struct Args {
...
+    manifest_id: &'a str,
+    operator_id: &'b str,
+}
+
+struct PostTarget {
+    manifest_id: String,
+    operator_id: String,
+}
+
+struct ResolvedApproveInputs {
```
daniilrrr, line 89: naming nit - `Resolved` is not that informative and imo redundant here. I think `DeployApproveInputs` is slightly better

daniilrrr: also really happy with the look of this struct as input to the core logic, nice 👍

tvc/src/commands/deploy/approve.rs
```diff
@@ -75,121 +76,229 @@ pub struct Args {
...
+                    ),
+                }
+            }
+        };
+        Some(PostTarget {
+            manifest_id,
+            operator_id,
+        })
+    } else {
+        None
```
daniilrrr, line 151: non-blocking - if it were me I'd return `None` early prior to this condition to collapse logic, but I leave that to you

tvc/src/commands/deploy/approve.rs
```diff
@@ -75,121 +76,229 @@ pub struct Args {
...
-            }
-        }
-    };
+async fn load_saved_operator_ids() -> anyhow::Result<Vec<String>> {
+    let config = Config::load().await?;
+    Ok(config.get_last_operator_ids().unwrap_or_default())
+}
 
+async fn post_approval_to_api(
+    plan: PostApprovalPlan<'_, '_>,
```
daniilrrr, line 279: the `plan` fields are `.into`'d into Strings immediately so IMO storing as `&str` is not majorly beneficial here. I think you should change to `String`s over `&str` and therefore not have to deal with lifetimes

```rust
struct PostApprovalPlan {
    manifest_id: String,
    operator_id: String,
}
```

tvc/src/commands/deploy/approve.rs
```diff
@@ -75,121 +76,229 @@ pub struct Args {
...
+        post_approval_to_api(plan, &approval).await?;
+    }
+
+    Ok(())
+}
+
+async fn load_manifest(args: &Args) -> anyhow::Result<(Manifest, Option<String>)> {
+    match (&args.manifest, &args.deploy_id) {
+        (Some(path), _) => Ok((read_manifest_from_path(path).await?, None)),
+        (_, Some(deploy_id)) => {
```
daniilrrr, line 233: this bit is a smidge confusing at first glance but I think I'm grokking that, in priority order:
1. prefer a manifest file `path` and read from there
2. else if we have a `deploy_id`, fetch from there
3. fail 

^ I think the above as a comment on the func is helpful

tvc/src/commands/deploy/create.rs
```diff
@@ -101,182 +130,180 @@ pub struct Args {
...
     pub public_ingress_port: Option<u16>,
+}
 
-    /// Path to an unencrypted pivot container pull secret file.
-    ///
-    /// The content will be encrypted based on the active org's API environment and
-    /// override `pivotContainerEncryptedPullSecret` from the config file.
-    #[arg(long, value_name = "PATH", env = "TVC_PIVOT_PULL_SECRET")]
-    pub pivot_pull_secret: Option<PathBuf>,
+struct ResolvedDeployInputs {
```
daniilrrr, line 135: same naming nit, would prefer `DeployCreateInputs`

tvc/src/config/app.rs
```diff
@@ -155,32 +158,57 @@ impl AppConfig {
...
+            });
         }
+
         if self.manifest_set_id.is_none() && self.manifest_set_params.is_none() {
-            anyhow::bail!("Must specify either manifestSetId or manifestSetParams");
+            errors.push(AppConfigValidationError::MissingOneOf {
+                first: "manifestSetId",
+                second: "manifestSetParams",
+            });
         }
```
daniilrrr, line 191: not sure but maybe this could be clearer as a `match`? leave that to you to try and decide

daniilrrr (dismissed):
looks great! just had non-blocking comments. changing `PostApprovalPlan` to use `String` is my strongest take

daniilrrr (approved):

Avi-D-coder (approved):
Approving as a dep reviewer only.
I have not reviewed logic only lock file

PR #156, opened 2026-06-03
feat: breaking rename external connectivity in config to enable egress to match api

emostov:
Rename config field name to better align with api field name

### DIFF
--- a/tvc/src/config/app.rs
+++ b/tvc/src/config/app.rs
@@ -13,7 +13,7 @@ pub struct AppConfig {
     pub name: String,
     pub quorum_public_key: String,
     #[serde(default)]
-    pub external_connectivity: Option<bool>,
+    pub enable_egress: Option<bool>,
     #[serde(default)]
     pub manifest_set_id: Option<String>,
     #[serde(default)]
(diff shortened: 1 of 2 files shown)


tvc/src/config/app.rs
```diff
@@ -13,7 +13,7 @@ pub struct AppConfig {
     pub name: String,
     pub quorum_public_key: String,
     #[serde(default)]
-    pub external_connectivity: Option<bool>,
+    pub enable_egress: Option<bool>,
```
daniilrrr, line 16: i think `Option<bool>` is redundant, should just be `bool` and the absence of a value when deserializing the JSON -> `false` not `None`. This is what I did in my earlier PR with `dangerous_enable_debug_mode_deployments: bool` below. 

The question comes down to whether we there is a meaningful distinction between 3 values [`None`, `Some(false)`, `Some(true)`] or 2 [`false`, `true`], and I think it's the latter case

daniilrrr (approved):

tvc/src/config/app.rs
richardpringle: Agree

PR #157, opened 2026-06-03
chore(tvc): refactor now-unsafe test code

daniilrrr:
- in Rust 2024, `std::env::set_var` is now unsafe
- those tests were not strictly necessary and can be done via the `cmd` approach


richardpringle (approved):
🚀

PR #160, opened 2026-06-11
ENG-3890: Explain TVC deployment ports

richardpringle:
Linear: ENG-3890

Summary:
- Add shared guidance for how to choose publicIngressPort and healthCheckPort in the TVC deploy flow.
- Show the resolved port setup before creating a deployment, including whether health checks share the public ingress listener or use a separate listener.
- Improve port flag help text and cover the new guidance in help/init output tests.

New output:
```bash
➜  rust-sdk git:(richard/eng-3890-explain-ports-when-creating-a-new-deployment) ✗ tvc deploy init
Created deployment config template: deploy-2026-06-11-094428.json

  tvc deploy create --config-file deploy-2026-06-11-094428.json

Port guidance:
  publicIngressPort is the port inside your container that receives public app
  traffic.
  healthCheckPort is the port TVC probes to decide whether the deployment is
  healthy.
  Use the same port for both unless your binary exposes health checks on a
  separate listener. Generated configs default both to 3000.
  ```

### DIFF
--- a/tvc/src/commands/deploy/mod.rs
+++ b/tvc/src/commands/deploy/mod.rs
@@ -1,5 +1,7 @@
 //! Deploy commands.
 
+use crate::config::deploy::DeployConfig;
+
 pub mod approve;
 pub mod create;
 pub mod delete;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by daniilrrr** [comment]
ran it 
```
tvc deploy create -h
Create a new deployment from a config file

Usage: tvc deploy create [OPTIONS]

Options:
  -c, --config-file <PATH>
          Path to the deployment configuration file (JSON). Optional when all required fields are provided via flags or env [env: TVC_DEPLOY_CONFIG=]
      --non-interactive
          Disable interactive prompts and fail fast when required values are missing [env: TVC_NON_INTERACTIVE=]
      --pivot-pull-secret <PATH>
          Path to an unencrypted pivot container pull secret file [env: TVC_PIVOT_PULL_SECRET=]
      --dangerous-deploy-debug-mode
          Deploy in debug mode, which forwards secure enclave logs to the host and zeroes attestation PCRs. This defeats the purpose of a secure enclave, so it should only be used to debug non-prod applications and view application logs [env: TVC_DANGEROUS_DEPLOY_DEBUG_MODE=]
      --app-id <APP_ID>
          Override the appId field [env: TVC_APP_ID=]
      --qos-version <QOS_VERSION>
          Override the qosVersion field [env: TVC_QOS_VERSION=]
      --pivot-image-url <PIVOT_IMAGE_URL>
          Override the pivotContainerImageUrl field [env: TVC_PIVOT_IMAGE_URL=]
      --expected-pivot-digest <EXPECTED_PIVOT_DIGEST>
          Override the expectedPivotDigest field [env: TVC_EXPECTED_PIVOT_DIGEST=]
      --pivot-path <PIVOT_PATH>
          Override the pivotPath field [env: TVC_PIVOT_PATH=]
      --pivot-args <ARG>
          Override pivotArgs (replaces the file's list entirely; not appended) [env: TVC_PIVOT_ARGS=]
      --health-check-port <HEALTH_CHECK_PORT>
          Container port TVC probes to decide whether the deployment is healthy [env: TVC_HEALTH_CHECK_PORT=]
      --public-ingress-port <PUBLIC_INGRESS_PORT>
          Container port that receives public app traffic [env: TVC_PUBLIC_INGRESS_PORT=]
  -h, --help
          Print help (see more with '--help')

Port guidance:
  publicIngressPort is the port inside your container that receives public app
  traffic.
  healthCheckPort is the port TVC probes to decide whether the deployment is
  healthy.
  Use the same port for both unless your binary exposes health checks on a
  separate listener. Generated configs default both to 3000.

```

Looks nice to me

File: tvc/src/commands/deploy/mod.rs [hunk 1]
```diff
@@ -9,3 +11,68 @@ pub mod post_share;
...
+        );
+    }
+
+    format!(
+        r#"Ports:
+  Public ingress uses container port {public_port}.
+  Health checks use container port {health_port}.
+  Separate ports are for binaries that expose health checks on a separate listener."#
+    )
+}
```
daniilrrr, line 41: this seems excessive or redundant to the `after_help` and I think a simple `log.INFO()` statement for the 2 ports is fine. A user will set these purposefully so they will know if they have a binary that has separate listeners for health

tvc/src/commands/deploy/mod.rs
```diff
@@ -9,3 +11,68 @@ pub mod post_share;
 pub mod provisioning_details;
 pub mod restore;
 pub mod status;
+
+pub(crate) const PORT_GUIDANCE: &str = r#"Port guidance:
```
daniilrrr, line 15: just an idea
```suggestion
pub(crate) const DEPLOY_CREATE_PORT_HELP: &str = r#"Port guidance:
```

daniilrrr (approved):

tvc/src/commands/deploy/mod.rs
richardpringle: I think "excessive" is kind of helpful for LLMs though. For humans reading this, I agree, but I don't think it gets in the way

PR #161, opened 2026-06-16
Feat/tvc passkeys

emostov:
TODO
- [ ] test fully locally
- [ ] support popping open browser to make initial login smoother with bootstrapping authenticator

### DIFF
--- a/tvc/src/client.rs
+++ b/tvc/src/client.rs
@@ -1,9 +1,9 @@
 //! Client utilities for authenticated API calls.
 
-use crate::config::turnkey::{Config, StoredApiKey};
+use crate::config::turnkey::{Config, StoredApiKey, StoredPasskeySession};
 use anyhow::{Context, Result, anyhow, bail};
 use tracing::debug;
-use turnkey_api_key_stamper::TurnkeyP256ApiKey;
+use turnkey_api_key_stamper::{Stamp, StampHeader, StamperError, TurnkeyP256ApiKey};
 use turnkey_client::TurnkeyClient;
 
 /// Number of *required* auth env vars: org_id, api_key_public, api_key_private.
@@ -18,11 +18,29 @@ const DEFAULT_API_BASE_URL: &str = "https://api.turnkey.com";
 /// An authenticated Turnkey client with organization context.
 pub struct AuthenticatedClient {
     /// The Turnkey API client.
-    pub client: TurnkeyClient<TurnkeyP256ApiKey>,
+    pub client: TurnkeyClient<TvcStamper>,
     /// The organization ID for API calls.
     pub org_id: String,
     /// The API base URL for the active org. Used for environment-specific behavior.
     pub api_base_url: String,
+}
+
+/// Authentication mechanisms supported by TVC after config resolution.
+pub enum TvcStamper {
+    ApiKey(TurnkeyP256ApiKey),
+    PasskeySession(StoredPasskeySession),
+}
+
+impl Stamp for TvcStamper {
+    fn stamp(&self, body: &[u8]) -> std::result::Result<StampHeader, StamperError> {
+        match self {
+            Self::ApiKey(api_key) => api_key.stamp(body),
+            Self::PasskeySession(session) => Ok(StampHeader {
+                name: "X-Session".to_string(),
+                value: session.session.clone(),
+            }),
+        }
+    }
 }
 
 /// Build an authenticated Turnkey client.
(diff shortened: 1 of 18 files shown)


tvc/src/client.rs
```diff
@@ -67,28 +89,37 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
         "resolved active organization config"
     );
 
-    let api_key = StoredApiKey::load(org_config)
-        .await?
-        .ok_or_else(|| anyhow!("No API key found for org '{alias}'. Run `tvc login` first."))?;
+    if let Some(session) = StoredPasskeySession::load(alias, org_config).await? {
+        return Ok((
+            org_config.id.clone(),
+            org_config.api_base_url.clone(),
```
richardpringle, line 95: Do you need to clone here?

tvc/src/commands/auth.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Human-readable passkey label in Turnkey.
+    #[arg(long)]
+    pub name: String,
+    /// WebAuthn transport to use for user presence.
+    #[arg(long, value_enum, default_value_t = PasskeyTransport::Auto)]
+    pub passkey_transport: PasskeyTransport,
+}
+
+#[derive(Debug, ClapArgs)]
+pub struct ListArgs {}
```
richardpringle: Not sure why we don't have lint for this... 

```suggestion
pub struct ListArgs;
```

tvc/src/commands/auth.rs
```diff
@@ -0,0 +1,89 @@
...
+pub struct Args {
+    #[command(subcommand)]
+    pub command: AuthCommands,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum AuthCommands {
+    /// Manage passkey authenticators.
+    Passkey(PasskeyArgs),
+}
```
richardpringle, line 17: Might be worth mentioning other types of planned auth? Otherwise this seems over-abstracted. 

(This is a nit though)

tvc/src/commands/login.rs
```diff
@@ -49,6 +69,14 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
         "running login command"
     );
 
+    if args.passkey {
+        return run_passkey_login(args, is_non_interactive).await;
+    }
+
+    if args.create_org {
+        return run_dashboard_signup_handoff(args).await;
+    }
```
richardpringle, line 80: What happens if the passkey and create-org options are passed in? With the --create-org arg just get ignored? 

We should have proper validation before we get here

tvc/src/commands/login.rs
```diff
@@ -60,6 +88,109 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
     execute_login(config, plan).await
 }
 
+async fn run_passkey_login(args: Args, is_non_interactive: bool) -> Result<()> {
+    if is_non_interactive {
+        bail!("passkey authentication requires an interactive terminal; remove --non-interactive");
+    }
```
richardpringle, line 100: It's cleaner to do validation outside of the function instead of early returning immediately

tvc/src/commands/login.rs
```diff
@@ -60,6 +88,109 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+        );
+    };
+
+    let mut config = Config::load().await?;
+    let org_query = match args.org.clone().or_else(|| config.active_org.clone()) {
+        Some(org) => org,
+        None => bail!(
+            "Organization is required for passkey login. Pass --org or run `tvc login` first."
+        ),
+    };
```
richardpringle, line 119: Let-else?

tvc/src/commands/login.rs
```diff
@@ -60,6 +88,109 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+
+    let alias = match find_org(&config, &org_query) {
+        Some((alias, _)) => alias.clone(),
+        None => {
+            bail!("Organization '{org_query}' not found. Run `tvc login` without --passkey first.")
+        }
+    };
```
richardpringle, line 123: Another let-else

tvc/src/passkey/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub transports: Vec<String>,
+}
+
+/// A synchronous WebAuthn ceremony provider.
+pub trait PasskeyCeremony: Send + Sync {
+    fn assert(&self, challenge: &str) -> Result<WebAuthnAssertion>;
+
+    fn attest(&self, _challenge: &str, _name: &str) -> Result<WebAuthnAttestation> {
+        bail!("passkey registration is not implemented for this transport")
+    }
```
richardpringle: Why not use subtraits here?

tvc/src/passkey/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl<C> WebAuthnStamper<C> {
+    pub fn new(ceremony: C) -> Self {
+        Self { ceremony }
+    }
+}
+
+impl WebAuthnStamper<FixtureCeremony> {
+    /// Construct a deterministic stamper for CI tests without hardware.
+    pub fn new_for_tests(assertion: WebAuthnAssertion) -> Self {
```
richardpringle, line 101: Should this be behind a feature flag?

tvc/tests/auth_env.rs
```diff
@@ -116,3 +118,24 @@ fn auth_falls_back_to_disk_config_when_required_env_vars_are_unset() {
...
+}
+
+#[test]
+fn env_api_key_stamper_still_uses_x_stamp() {
+    let stamper = TvcStamper::ApiKey(TurnkeyP256ApiKey::generate());
+
+    let header = stamper.stamp(b"{}").unwrap();
+
+    assert_eq!(header.name, "X-Stamp");
+}
```
richardpringle: Why put these in a separate test file?

richardpringle (commented):
Not sure how useful this review is since this is in draft... 

Basically all nits

r-n-o (commented):
I'm not sure I follow the intent of this PR. Is the idea that we want to support popping open a browser / create a new org from the CLI? Or do we want to support true passkey auth where actions are authorized by a CLI-initiated passkey prompt? (and the user would then have to tap their yubikey or use touchID to allow activities like "CREATE_DEPLOYMENT" to be signed)

If we're going for easier onboarding I believe we should work with the dashboard APIs and make them more CLI friendly so that a passkey isn't required to create an org, potentially? Might be a medium/large project I'm not sure. And if we want to work on logging into the CLI for the first time (with an existing org), I think the challenge will be to prompt users with the right passkey. E.g. if a user creates their org with a passkey in Chrome vs. system passkey vs. 1password passkey we'll run into a serious lack of context if we try to pull the same passkey. Another challenge is the fact that we have non-resident keys; so you need to do email verification first, and then you can pull a list of credential IDs. Without the credential IDs from the server you can't prompt the user for a passkey signature.

emostov:
Closing this draft in favor of Linear tracking: [TVC-7](https://linear.app/turnkey/issue/TVC-7/cli-support-for-yubikey-webauthn-turnkey-authentication) for YubiKey/WebAuthn login, and [TVC-222](https://linear.app/turnkey/issue/TVC-222/cli-support-full-authenticator-enrollment-without-leaving-the-cli) for completing authenticator enrollment without leaving the CLI.

PR #164, opened 2026-06-17
[cancelled] feat(tvc): add deploy debug-logs command

daniilrrr:
This adds the rust-sdk side of server-streaming RPC support for the TVC debug-log endpoint and wires it into the TVC CLI.

Changes:
- Updates the rust-sdk proto inputs with contents from the mono WIP branch for TVC-6790 / 6790, including `GetEnclaveDebugLogs`, `LogLine`, and `LogEventType`.
- Updates codegen to model server-streaming responses with `RpcResponseKind::Streaming` and generate a `BoxStream` client method. Unary RPC behavior is unchanged; client-streaming and bidi-streaming RPCs remain unsupported.
- Adds client support for grpc-gateway streamed JSON response envelopes.
- Regenerates the checked-in Rust client/generated files from the synced protos.
- Adds `tvc deploy debug-logs` for streaming enclave debug logs.
- Wires the command to the generated `GetEnclaveDebugLogs` API.
- Formats streamed log lines with pod name and timestamp.
- Adds CLI parsing and output formatting tests.

Note: the proto files may need to be resynced with latest mono `main` branch contents by the time this rust-sdk change is actually ready to land.

Tested with:
- `make generate`
- `cargo fmt --check`
- `cargo check -p turnkey_client -p turnkey_codegen`
- `cargo test -p turnkey_codegen`
- `cargo test -p turnkey_client`
- `cargo check -p tvc`
- `cargo test -p tvc`

TODO: dev testing via the public endpoint


daniilrrr:
**Human comment**
If we don't want to introduce the new `futures-util` dep, we can hand roll our own Stream type. However, we would lose out on Stream functions and typical usage of a Rust stream. It is also already in the repo as a transitive dependency, just not a direct one. 

___
**AI Comment below:**

Design note on `futures-util` vs an SDK-owned stream type:

We considered avoiding a direct `futures-util` dependency by returning a custom SDK type, e.g. `TurnkeyResponseStream<T>`, with an inherent async method:

```rust
while let Some(item) = stream.next().await? {
    ...
}
```

That would be dependency-light and sufficient for the immediate `tvc` CLI use case, because the CLI only needs to read and print events sequentially.

The tradeoff is that a hand-rolled type would not participate in the standard async stream ecosystem unless we also implemented `Stream` for it. Callers would lose direct access to common stream combinators and integrations such as:

- `take`, `map`, `filter_map`
- `try_next`, `try_for_each`, `try_fold`, `try_collect`
- `boxed`
- passing the result into APIs that accept `impl Stream`
- composing it with other stream pipelines

Since server-streaming RPCs are naturally represented as async streams, this PR uses `futures_util::stream::BoxStream` and `StreamExt` instead. That makes the generated SDK method more idiomatic and more generally useful for SDK consumers, at the cost of adding `futures-util` as a direct public dependency.

`futures-util` is not part of Rust std/core, but it is the common crate used for stream extension traits and utilities in the Rust async ecosystem. Given this is generated SDK surface rather than only internal CLI plumbing, I think the composability is worth the direct dependency here.

daniilrrr:
closing in favor of #169 and dropping the streaming approach

PR #169, opened 2026-06-18
feat(tvc): add deploy debug-logs command

daniilrrr:
~Depends on https://github.com/tkhq/rust-sdk/pull/171 to merge first.~ Done

## Summary

Adds `tvc deploy debug-logs`, backed by the unary `GetTvcDeploymentDebugLogs` endpoint at `/public/v1/query/get_tvc_deployment_debug_logs`.
___
### Sample command:
`tvc deploy debug-logs --deploy-id $MY_DEPLOY_ID --since-seconds 3600 --poll`
___
The command fetches debug-mode TVC deployment logs and supports bounded history.

**The recommended history control is `--since-seconds`,** which returns logs newer than a relative time window and is the most predictable way to inspect recent output.

`--tail-lines` is also available for advanced scoping, **but it corresponds to raw pod log history per replica, so filtered output will contain fewer visible app lines than requested.**

By default, output omits the Kubernetes-prepended platform timestamp. Use `--include-platform-timestamp` to show it.

Note that debug mode deployments now only have 1 replica by default, so most users will see `replica 1/1` when they use this command.

## Notes

Poll mode is implemented client-side by repeatedly calling the unary endpoint. Poll requests use a small overlapping time window: by default the CLI polls every 2 seconds and requests a 2-second overlap to avoid missing late-arriving lines.

The CLI keeps a bounded cache of recently printed timestamped lines to suppress duplicates across overlap windows. The cache defaults to 1000 entries and can be adjusted with `--recent-line-capacity` for unusually high-volume logs. Dedupe is enabled by default; pass `--disable-dedupe` to print every returned line.

CLI defaults and numeric flag validation are handled by clap.

~Requires the corresponding mono debug-logs unary endpoint PR to land/deploy.~ Done.

## Testing

- Added CLI tests for help text, required flags, flag validation, debug-log command parsing, poll-mode dedupe behavior, and the dedupe disable path.
- `cargo fmt --check`
- `cargo test -p turnkey_client`
- `cargo test -p tvc debug_logs`

- Tested in dev by deploying a test image with extra logs, connecting to that deployment, then refreshing the `/time` page for additional logs. This sample was captured from an older 3-replica dev deployment.
```logs
tvc deploy debug-logs --deploy-id e4b4d6b9-97e3-4b23-b6fa-a425f049f310 --since-seconds 3600 --poll
Connected; polling for debug logs...
replica 2/3 2026-06-18T22:10:38.731409Z  INFO helloworld: Server listening on 0.0.0.0:3000
replica 2/3 2026-06-18T22:10:38.731431Z  WARN helloworld: (Dummy warn): Server listening on 0.0.0.0:3000
replica 2/3 2026-06-18T22:10:38.731435Z ERROR helloworld: (Dummy error): Server listening on 0.0.0.0:3000
```

### DIFF
--- a/tvc/src/cli.rs
+++ b/tvc/src/cli.rs
@@ -9,28 +9,29 @@ CLI for building with Turnkey Verifiable Cloud.
 
 Some commands accept multiple configuration input types.
 Configuration values are resolved in this order, highest priority first:
-  1. Command-line flag (e.g. --app-id)
-  2. Environment variable (e.g. TVC_APP_ID)
-  3. Config file value (--config-file)
-  4. Built-in default
+    1. Command-line flag (e.g. --app-id)
+    2. Environment variable (e.g. TVC_APP_ID)
+    3. Config file value (--config-file)
+    4. Built-in default
 
 Special rules (exceptions to the order above):
-  --pivot-args replaces the config file's list entirely (does not append)
-  Debug-mode flags (--dangerous-deploy-debug-mode and
+    --pivot-args replaces the config file's list entirely (does not append)
+
+    Debug-mode flags (--dangerous-deploy-debug-mode and
     --dangerous-enable-debug-mode-deployments) are opt-in only: the flag
     or its env var can turn debug mode ON, but its absence never turns OFF
     a config file that enables it. To disable debug mode, set it false in
     the config file (or omit it) and do not pass the flag.
 
 Authentication:
-  Local: run `tvc login` once; commands then read ~/.config/turnkey/.
-  CI:    set TVC_ORG_ID, TVC_API_KEY_PUBLIC, and TVC_API_KEY_PRIVATE
+    Local: run `tvc login` once; commands then read ~/.config/turnkey/.
+    CI:    set TVC_ORG_ID, TVC_API_KEY_PUBLIC, and TVC_API_KEY_PRIVATE
          to authenticate without files. Env vars take precedence over local
          config files. Setting some but not all three required vars will error.
 
 Interactive behavior:
-  By default, commands may prompt when stdin is a TTY. Use --non-interactive
-  or set TVC_NON_INTERACTIVE=true to disable prompts and fail fast instead.";
+    By default, commands may prompt when stdin is a TTY. Use --non-interactive
+    or set TVC_NON_INTERACTIVE=true to disable prompts and fail fast instead.";
 
 /// CLI command parsing and dispatch.
 #[derive(Debug, Parser)]
@@ -79,6 +80,7 @@ impl Cli {
                 DeployCommands::Init(args) => {
                     commands::deploy::init::run(args, non_interactive).await
                 }
+                DeployCommands::DebugLogs(args) => commands::deploy::debug_logs::run(args).await,
                 DeployCommands::Delete(args) => commands::deploy::delete::run(args).await,
                 DeployCommands::Restore(args) => commands::deploy::restore::run(args).await,
             },
@@ -160,6 +162,9 @@ enum DeployCommands {
     Create(commands::deploy::create::Args),
     /// Generate a template deployment configuration file.
     Init(commands::deploy::init::Args),
+    /// Fetch debug logs for a deployment.
+    #[command(long_about = commands::deploy::debug_logs::LONG_ABOUT)]
+    DebugLogs(commands::deploy::debug_logs::Args),
     /// Delete a deployment by marking it for deletion.
     Delete(commands::deploy::delete::Args),
     /// Restore a deleted deployment.
@@ -176,6 +181,7 @@ impl DeployCommands {
             DeployCommands::Status(_) => "deploy status",
             DeployCommands::Create(_) => "deploy create",
             DeployCommands::Init(_) => "deploy init",
+            DeployCommands::DebugLogs(_) => "deploy debug-logs",
             DeployCommands::Delete(_) => "deploy delete",
             DeployCommands::Restore(_) => "deploy restore",
         }
--- a/tvc/src/commands/deploy/debug_logs.rs
+++ b/tvc/src/commands/deploy/debug_logs.rs
@@ -13,7 +13,6 @@ use turnkey_client::generated::{
 
 const DEFAULT_FOLLOW_POLL_INTERVAL_SECONDS: i64 = 2;
 const FOLLOW_OVERLAP_SECONDS: i64 = 2;
-const RECENT_LOG_LINE_CAPACITY: usize = 1000;
 
 fn follow_poll_interval_seconds_parser() -> clap::builder::RangedI64ValueParser<i64> {
     clap::value_parser!(i64).range(1..=i64::MAX - FOLLOW_OVERLAP_SECONDS)
@@ -25,6 +24,10 @@ fn non_negative_i32_parser() -> clap::builder::RangedI64ValueParser<i32> {
 
 fn non_negative_i64_parser() -> clap::builder::RangedI64ValueParser<i64> {
     clap::value_parser!(i64).range(0..)
+}
+
+fn positive_usize_parser() -> clap::builder::RangedU64ValueParser<usize> {
+    clap::builder::RangedU64ValueParser::<usize>::new().range(1..)
 }
 
 pub(crate) const LONG_ABOUT: &str = r#"
@@ -94,6 +97,16 @@ pub struct Args {
     /// Include the platform timestamp prepended by the Kubernetes log stream.
     #[arg(long, env = "TVC_DEBUG_LOGS_INCLUDE_PLATFORM_TIMESTAMP")]
     pub include_platform_timestamp: bool,
+
+    /// Number of recently printed timestamped lines retained for follow-mode dedupe. This should
+    /// only be increased if high-volume logs still show duplicates across poll overlap windows.
+    #[arg(
+        long,
+        env = "TVC_DEBUG_LOGS_RECENT_LINE_CAPACITY",
+        default_value_t = 1000,
+        value_parser = positive_usize_parser()
+    )]
+    pub recent_line_capacity: usize,
 }
 
 /// Run the `deploy debug-logs` command.
@@ -108,6 +121,7 @@ pub async fn run(args: Args) -> anyhow::Result<()> {
         tail_lines: args.tail_lines,
         since_seconds: args.since_seconds,
         include_platform_timestamp: args.include_platform_timestamp,
+        recent_line_capacity: args.recent_line_capacity,
     };
 
     query_debug_logs(&auth.client, request).await
@@ -122,6 +136,7 @@ struct DebugLogQueryRequest {
     tail_lines: i32,
     since_seconds: i64,
     include_platform_timestamp: bool,
+    recent_line_capacity: usize,
 }
 
 impl From<DebugLogQueryRequest> for GetTvcDeploymentDebugLogsRequest {
@@ -149,8 +164,10 @@ async fn query_debug_logs(
     client: &turnkey_client::TurnkeyClient<TurnkeyP256ApiKey>,
     request: DebugLogQueryRequest,
 ) -> anyhow::Result<()> {
-    let mut log_printer =
-        DebugLogPrinter::new(request.include_platform_timestamp, RECENT_LOG_LINE_CAPACITY);
+    let mut log_printer = DebugLogPrinter::new(
+        request.include_platform_timestamp,
+        request.recent_line_capacity,
+    );
 
     let (current_request, followup_request) = if request.follow {
         (request.clone(), Some(request.into_followup_request()))
@@ -195,15 +212,13 @@ struct LogLineKey {
 }
 
 impl LogLineKey {
-    fn new(replica: &str, line: &LogLine) -> Option<Self> {
-        let ts = line.ts.as_ref()?;
-
-        Some(Self {
+    fn new(replica: &str, ts: Timestamp, content: String) -> Self {
+        Self {
             replica: replica.into(),
-            content: line.content.clone(),
-            seconds: ts.seconds.clone(),
-            nanos: ts.nanos.clone(),
-        })
+            content,
+            seconds: ts.seconds,
+            nanos: ts.nanos,
+        }
     }
 }
 
@@ -211,34 +226,33 @@ impl LogLineKey {
 struct RecentLogLines {
     seen: HashSet<LogLineKey>,
     order: VecDeque<LogLineKey>,
-    capacity: usize,
 }
 
 impl RecentLogLines {
     fn new(capacity: usize) -> Self {
+        assert!(capacity > 0, "recent log line capacity must be positive");
+
         Self {
-            seen: HashSet::new(),
-            order: VecDeque::new(),
-            capacity,
+            seen: HashSet::with_capacity(capacity),
+            order: VecDeque::with_capacity(capacity),
         }
     }
 
     fn insert(&mut self, key: LogLineKey) -> bool {
-        if self.capacity == 0 {
-            return true;
-        }
-
-        if !self.seen.insert(key.clone()) {
+        if self.seen.contains(&key) {
             return false;
         }
 
-        self.order.push_back(key);
-
-        while self.order.len() > self.capacity {
-            if let Some(oldest) = self.order.pop_front() {
-                self.seen.remove(&oldest);
-            }
+        if self.order.len() == self.order.capacity() {
+            let oldest = self
+                .order
+                .pop_front()
+                .expect("full recent log line cache should contain an oldest key");
+            self.seen.remove(&oldest);
         }
+
+        self.seen.insert(key.clone());
+        self.order.push_back(key);
 
         true
     }
@@ -247,15 +261,24 @@ impl RecentLogLines {
     ///
     /// NOTE: Mutating filter
     fn record_if_new(&mut self, replica: &str, line: &LogLine) -> bool {
-        match LogLineKey::new(replica, line) {
-            Some(key) => self.insert(key),
-            None => true,
-        }
+        let LogLine { content, ts } = line;
+        let Some(ts) = ts else {
+            return true;
+        };
+
+        let key = LogLineKey::new(replica, ts.clone(), content.clone());
+
+        self.insert(key)
     }
 
     #[cfg(test)]
     fn len(&self) -> usize {
         self.seen.len()
+    }
+
+    #[cfg(test)]
+    fn capacity(&self) -> usize {
+        self.order.capacity()
     }
 }
 
@@ -342,6 +365,10 @@ mod tests {
         }
     }
 
+    fn log_line_key(replica: &str, content: &str, seconds: &str, nanos: &str) -> LogLineKey {
+        LogLineKey::new(replica, timestamp(seconds, nanos), content.to_string())
+    }
+
     #[test]
     fn request_maps_to_unary_api_request() {
         let request = DebugLogQueryRequest {
@@ -352,6 +379,7 @@ mod tests {
             tail_lines: 10,
             since_seconds: 30,
             include_platform_timestamp: true,
+            recent_line_capacity: 1000,
         };
 
         let api_request: GetTvcDeploymentDebugLogsRequest = request.into();
@@ -372,6 +400,7 @@ mod tests {
             tail_lines: 100,
             since_seconds: 0,
             include_platform_timestamp: false,
+            recent_line_capacity: 1000,
         };
 
         let followup = request.clone().into_followup_request();
@@ -384,19 +413,20 @@ mod tests {
 
     #[test]
     fn log_line_key_matches_duplicate_lines() {
-        let line = log_line("hello", Some(timestamp("1710000000", "123456789")));
-
         assert_eq!(
-            LogLineKey::new("replica 1/3", &line),
-            LogLineKey::new("replica 1/3", &line)
+            log_line_key("replica 1/3", "hello", "1710000000", "123456789"),
+            log_line_key("replica 1/3", "hello", "1710000000", "123456789")
         );
     }
 
     #[test]
-    fn log_line_key_requires_timestamp() {
+    fn record_if_new_treats_missing_timestamp_as_new() {
         let line = log_line("hello", None);
+        let mut recent = RecentLogLines::new(1000);
 
-        assert_eq!(LogLineKey::new("replica 1/3", &line), None);
+        assert!(recent.record_if_new("replica 1/3", &line));
+        assert!(recent.record_if_new("replica 1/3", &line));
+        assert_eq!(recent.len(), 0);
     }
 
     #[test]
@@ -440,9 +470,9 @@ mod tests {
     }
 
     #[test]
-    fn recent_log_lines_treats_untimestamped_lines_as_new() {
+    fn recent_log_lines_treats_non_timestamped_lines_as_new() {
         let line = log_line("hello", None);
-        let mut recent = RecentLogLines::new(RECENT_LOG_LINE_CAPACITY);
+        let mut recent = RecentLogLines::new(1000);
 
         assert!(recent.record_if_new("replica 1/2", &line));
         assert!(recent.record_if_new("replica 1/2", &line));
@@ -452,11 +482,17 @@ mod tests {
     #[test]
     fn recent_log_lines_dedupes_timestamped_lines() {
         let line = log_line("hello", Some(timestamp("1710000000", "1")));
-        let mut recent = RecentLogLines::new(RECENT_LOG_LINE_CAPACITY);
+        let mut recent = RecentLogLines::new(1000);
 
         assert!(recent.record_if_new("replica 1/2", &line));
         assert!(!recent.record_if_new("replica 1/2", &line));
         assert_eq!(recent.len(), 1);
+    }
+
+    #[test]
+    #[should_panic(expected = "recent log line capacity must be positive")]
+    fn recent_log_lines_requires_positive_capacity() {
+        RecentLogLines::new(0);
     }
 
     #[test]
@@ -474,7 +510,7 @@ mod tests {
                 entry("replica 2/2", None),
             ],
         };
-        let mut printer = DebugLogPrinter::new(false, RECENT_LOG_LINE_CAPACITY);
+        let mut printer = DebugLogPrinter::new(false, 1000);
 
         printer.print_response(&response);
 
@@ -483,21 +519,9 @@ mod tests {
 
     #[test]
     fn recent_log_lines_evicts_oldest_entries() {
-        let first = LogLineKey::new(
-            "replica 1/2",
-            &log_line("first", Some(timestamp("1710000000", "1"))),
-        )
-        .unwrap();
-        let second = LogLineKey::new(
-            "replica 1/2",
-            &log_line("second", Some(timestamp("1710000000", "2"))),
-        )
-        .unwrap();
-        let third = LogLineKey::new(
-            "replica 1/2",
-            &log_line("third", Some(timestamp("1710000000", "3"))),
-        )
-        .unwrap();
+        let first = log_line_key("replica 1/2", "first", "1710000000", "1");
+        let second = log_line_key("replica 1/2", "second", "1710000000", "2");
+        let third = log_line_key("replica 1/2", "third", "1710000000", "3");
         let mut recent = RecentLogLines::new(2);
 
         assert!(recent.insert(first.clone()));
@@ -508,5 +532,21 @@ mod tests {
         assert_eq!(recent.len(), 2);
         assert!(recent.insert(first));
         assert_eq!(recent.len(), 2);
+    }
+
+    #[test]
+    fn recent_log_lines_does_not_grow_past_initial_capacity() {
+        let first = log_line_key("replica 1/2", "first", "1710000000", "1");
+        let second = log_line_key("replica 1/2", "second", "1710000000", "2");
+        let third = log_line_key("replica 1/2", "third", "1710000000", "3");
+        let mut recent = RecentLogLines::new(2);
+        let initial_capacity = recent.capacity();
+
+        assert!(recent.insert(first));
+        assert!(recent.insert(second));
+        assert!(recent.insert(third));
+
+        assert_eq!(recent.len(), 2);
+        assert_eq!(recent.capacity(), initial_capacity);
     }
 }
--- a/tvc/src/commands/deploy/debug_logs.rs
+++ b/tvc/src/commands/deploy/debug_logs.rs
@@ -49,6 +49,8 @@ Poll mode polls every 2 seconds by default and requests a 2-second overlap
 on each subsequent poll to avoid missing late-arriving log lines.
 By default, output omits the platform timestamp that Kubernetes prepends to each
 log line; pass `--include-platform-timestamp` to show it.
+The CLI dedupes timestamped lines with identical content across poll overlap windows
+by default; pass `--disable-dedupe` to print every returned line.
 See `tvc app create --help`, `tvc deploy create --help`, and `tvc --help`
 for more info."#;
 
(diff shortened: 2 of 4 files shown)


tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,423 @@
...
+use std::collections::HashSet;
+use std::time::Duration;
+use turnkey_client::TurnkeyP256ApiKey;
+use turnkey_client::generated::external::data::v1::{LogLine, Timestamp};
+use turnkey_client::generated::{
+    EnclaveDebugLogEntry, GetEnclaveDebugLogsRequest, GetEnclaveDebugLogsResponse,
+};
+
+const FOLLOW_POLL_INTERVAL: Duration = Duration::from_secs(2);
+const FOLLOW_SINCE_SECONDS: i64 = 5;
```
daniilrrr: lets make FOLLOW_POLL_INTERVAL optionally configurable and FOLLOW_SINCE_SECONDS (aka the overlap, right?) by default be (FOLLOW_POLL_INTERVAL+ 2) sec. So whichever poll you do, the since_seconds should be an additional 2 sec overlap.

daniilrrr: a default FOLLOW_POLL_INTERVAL of 2sec is fine

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,423 @@
...
+    seconds: String,
+    nanos: String,
+}
+
+impl LogLineKey {
+    fn new(replica: &str, line: &LogLine) -> Option<Self> {
+        let ts = line.ts.as_ref()?;
+
+        Some(Self {
+            replica: replica.to_string(),
```
daniilrrr:
```suggestion
            replica: replica.into(),
```

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,423 @@
...
+) {
+    let Some(line) = entry.line.as_ref() else {
+        return;
+    };
+
+    if let Some(key) = LogLineKey::new(&entry.replica, line) {
+        if !printed_lines.insert(key) {
+            return;
+        }
+    }
```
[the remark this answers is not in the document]
daniilrrr: this is a good idea however it opens us up to potential duplication of logs at the point of `.clear()`.  It would be ideal to "clear `printed_lines` except for the last 50 line," for which the ideal data structure would be a "sorted HashSet," and light Googling says that the 3rd party [OrderSet](https://docs.rs/ordermap/latest/ordermap/set/struct.OrderSet.html) is the best tool in Rust. However, this repo is sensitive to new deps, especially 3rd-party ones, so I don't think this edge case is worth handling via the dependency.

So I agree with the suggestion even if it introduces log line duplication every 10k lines. 

That said, we could mitigate simply by tracking a `last-seen timestamp` as a fallback when the `printed_lines HashSet` is empty. That could be a simple bandaid for this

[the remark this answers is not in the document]
daniilrrr: consulted with LLM and a simple HashSet + Vecdeque impl would work here for this FIFO set. advantageous cuz we can set a `capacity` and evict on that

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: Addressed in `aef00ee`: added `--follow-poll-interval-seconds` / `TVC_DEBUG_LOGS_FOLLOW_POLL_INTERVAL_SECONDS`, kept the default at 2 seconds, and follow-up requests now use `since_seconds = poll_interval_seconds + 2` for the overlap window.

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: Addressed in `aef00ee`: changed this to `replica.into()`.

tvc/src/commands/deploy/debug_logs.rs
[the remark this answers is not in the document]
daniilrrr: Addressed in `aef00ee`: replaced the unbounded `HashSet` with a bounded `RecentLogLines` helper backed by `HashSet + VecDeque`, with capacity 1000. The set remains the membership check and the deque is only eviction order, so this avoids a full-clear duplicate burst while keeping memory bounded.

daniilrrr:
Self-reminder to re-sync protos once backend PR in `mono` lands

tvc/src/cli.rs
```diff
@@ -8,29 +8,31 @@ const LONG_ABOUT: &str = "\
 CLI for building with Turnkey Verifiable Cloud.
 
 Some commands accept multiple configuration input types.
+f
 Configuration values are resolved in this order, highest priority first:
```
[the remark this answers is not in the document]
Avi-D-coder: yep this

proto/services/coordinator/public/v1/public_api.proto
```diff
@@ -2007,6 +2021,114 @@ service PublicApiService {
...
+    option (google.api.http) = {
+      post: "/public/v1/submit/create_mfa_policy"
+      body: "*"
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      description: "Create a new MFA policy for a user."
+      summary: "Create MFA policy"
+      tags: "MFA Policies"
+    };
+  }
```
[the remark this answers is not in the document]
daniilrrr: u right. I will sync protos in a separate PR.

daniilrrr:
will sync protos in a diff PR to reduce $GREEN_LINE_COUNT 

https://github.com/tkhq/rust-sdk/pull/171

proto/services/coordinator/public/v1/public_api.proto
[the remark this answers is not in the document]
daniilrrr: https://github.com/tkhq/rust-sdk/pull/171

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
+//! Deploy debug-logs command.
+
+use anyhow::{Context, bail};
+use chrono::{DateTime, SecondsFormat, Utc};
+use clap::Args as ClapArgs;
+use std::collections::{HashSet, VecDeque};
+use std::time::Duration;
+use turnkey_client::TurnkeyP256ApiKey;
+#[cfg(test)]
```
richardpringle: nit: pull this into the `mod tests` block

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+    )]
+    pub follow_poll_interval_seconds: Option<i64>,
+
+    /// Limit raw pod log history requested per replica. Filtered output may contain (many) fewer lines.
+    #[arg(long, env = "TVC_DEBUG_LOGS_TAIL_LINES", allow_hyphen_values = true)]
+    pub tail_lines: Option<i32>,
+
+    /// Return logs newer than this many seconds ago.
+    #[arg(long, env = "TVC_DEBUG_LOGS_SINCE_SECONDS", allow_hyphen_values = true)]
+    pub since_seconds: Option<i64>,
```
richardpringle: You should have `clap` do validation for you: 
```suggestion
   /// Seconds to wait between follow-mode polls.
    #[arg(
        long,
        env = "TVC_DEBUG_LOGS_FOLLOW_POLL_INTERVAL_SECONDS",
        allow_hyphen_values = true
    )]
    pub follow_poll_interval_seconds: Option<u16>,

    /// Limit raw pod log history requested per replica. Filtered output may contain (many) fewer lines.
    #[arg(long, env = "TVC_DEBUG_LOGS_TAIL_LINES", allow_hyphen_values = true)]
    pub tail_lines: Option<u16>,

    /// Return logs newer than this many seconds ago.
    #[arg(
        long,
        env = "TVC_DEBUG_LOGS_SINCE_SECONDS",
        allow_hyphen_values = true,
        value_parser = clap::value_parser!(u64).range(..=(i64::MAX - FOLLOW_OVERLAP_SECONDS) as u64)
    )]
    pub since_seconds: Option<u64>,
```

Shouldn't necessarily be exactly as in my suggestion, but I do think a u16 provides a sane max for the interval and tail lines. In any case, you should use the clap value-parser and just put maxes such that the type conversion is trivial and error proof

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+    deployment_id: String,
+    follow: bool,
+    follow_poll_interval_seconds: i64,
+    tail_lines: i32,
+    since_seconds: i64,
+    include_platform_timestamp: bool,
+}
+
+impl DebugLogQueryRequest {
+    fn to_api_request(&self) -> GetTvcDeploymentDebugLogsRequest {
```
richardpringle: This a bit of a nit, as it doesn't really matter here, but the pattern that I would prefer to see is forcing any cloning outside of function (this really makes a difference when you're debugging something on a hot path). I would also implement this as a standard `From` implementation since it's a pure type conversion
```rust
impl From<DebugLogQueryRequest> for GetTvcDeploymentDebugLogsRequest {
    fn from(req: DebugLogQueryRequest) -> Self {
        Self {
            organization_id: req.organization_id,
            deployment_id: req.deployment_id,
            tail_lines: req.tail_lines.try_into().unwrap(),
            since_seconds: req.since_seconds,
        }
    }
}
```
where you can just call `request.clone().into()` at the call-site. The key difference here is that cloning is explicit instead of implicit.

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+See `tvc app create --help`, `tvc deploy create --help`, and `tvc --help`
+for more info."#;
+
+/// Fetch debug logs for a deployment.
+#[derive(Debug, ClapArgs)]
+#[command(about, long_about = LONG_ABOUT)]
+pub struct Args {
+    /// ID of the deployment.
+    #[arg(short = 'd', long, env = "TVC_DEPLOY_ID")]
+    pub deploy_id: String,
```
richardpringle, line 63: Is this not necessarily a uuid? A uuid is generally `Copy` which would make this easier to deal with

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+        since_seconds,
+        include_platform_timestamp: args.include_platform_timestamp,
+    };
+
+    query_debug_logs(&auth.client, &request).await
+}
+
+#[derive(Clone, Debug)]
+struct DebugLogQueryRequest {
+    organization_id: String,
```
richardpringle, line 139: Is this not necessarily a uuid?

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+    }
+
+    current_request = request.followup_request();
+
+    let follow_poll_interval = Duration::from_secs(request.follow_poll_interval_seconds as u64);
+    loop {
+        tokio::time::sleep(follow_poll_interval).await;
+        let response = fetch_debug_logs(client, &current_request).await?;
+        log_printer.print_response(&response);
+    }
```
richardpringle: Obviously not for this PR, but it seems like it would be a good feature for the API to have cursor based pagination on logs instead of requiring the overlap. I assume there's a chance we have duplicate logs in here?

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+}
```
richardpringle, line 202: Along with some of my other comments, I would do something more along these lines: 

```suggestion
async fn query_debug_logs(
    client: &turnkey_client::TurnkeyClient<TurnkeyP256ApiKey>,
    request: DebugLogQueryRequest,
) -> anyhow::Result<()> {
    let mut log_printer =
        DebugLogPrinter::new(request.include_platform_timestamp, RECENT_LOG_LINE_CAPACITY);

    let (current_request, followup_request) = if request.follow {
        (request.clone(), Some(request.into_followup_request()))
    } else {
        (request, None)
    };

    let response = fetch_debug_logs(client, current_request).await?;

    if followup_request.is_some() {
        eprintln!("Connected; polling for debug logs...");
    }

    log_printer.print_response(&response);

    let Some(followup_request) = followup_request else {
        return Ok(());
    };

    let follow_poll_interval =
        Duration::from_secs(followup_request.follow_poll_interval_seconds as u64);
    loop {
        tokio::time::sleep(follow_poll_interval).await;
        let response = fetch_debug_logs(client, followup_request.clone()).await?;
        log_printer.print_response(&response);
    }
}
```

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+
+async fn fetch_debug_logs(
+    client: &turnkey_client::TurnkeyClient<TurnkeyP256ApiKey>,
+    request: &DebugLogQueryRequest,
+) -> anyhow::Result<GetTvcDeploymentDebugLogsResponse> {
+    client
+        .get_tvc_deployment_debug_logs(request.to_api_request())
```
richardpringle: Not for this PR, but the API here is really unfortunate (and not idiomatic). `get_tvc_deployment_debug_logs` takes an owned type (the request) then uses is as a borrowed value. Is there an easy way we can change this?

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+    let Some(tail_lines) = tail_lines else {
+        return Ok(0);
+    };
+
+    if tail_lines < 0 {
+        bail!("--tail-lines must be greater than or equal to 0");
+    }
+
+    Ok(tail_lines)
+}
```
richardpringle: You don't need this function if you use the clap value parser, you can just call `unwrap_or_default` on a `uXX` type.

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+    let Some(since_seconds) = since_seconds else {
+        return Ok(0);
+    };
+
+    if since_seconds < 0 {
+        bail!("--since-seconds must be greater than or equal to 0");
+    }
+
+    Ok(since_seconds)
+}
```
richardpringle: same here

tvc/src/commands/deploy/debug_logs.rs
richardpringle: Actually, even better, you should use `default_value_t`: https://docs.rs/clap/latest/clap/_derive/index.html

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+            nanos: ts.nanos.clone(),
+        })
+    }
+}
+
+#[derive(Debug)]
+struct RecentLogLines {
+    seen: HashSet<LogLineKey>,
+    order: VecDeque<LogLineKey>,
+    capacity: usize,
```
richardpringle:
```suggestion
```

Explanation below

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+}
+
+impl RecentLogLines {
+    fn new(capacity: usize) -> Self {
+        Self {
+            seen: HashSet::new(),
+            order: VecDeque::new(),
+            capacity,
+        }
+    }
```
richardpringle, line 247:
```suggestion
```suggestion
    fn new(capacity: usize) -> Self {
        Self {
            seen: HashSet::with_capacity(capacity),
            order: VecDeque::with_capacity(capacity),
        }
    }
```

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+
+    fn insert(&mut self, key: LogLineKey) -> bool {
+        if self.capacity == 0 {
+            return true;
+        }
```
richardpringle: I haven't seen where this is used, but my intuition tells me it would be better to have `Option<RecentLogLines>` than to have a capacity of 0. If we had an `Option` and I'm debugging and I see a `None`, it's clear why things aren't being logged as opposed to implicitly dropping the logs

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+
+        if !self.seen.insert(key.clone()) {
+            return false;
+        }
+
+        self.order.push_back(key);
```
richardpringle, line 263: I believe `VecDeque` functions the same way that a `Vec` does where it doubles in capacity every time you reach max capacity. It's therefore inefficient to push, then pop... and I shouldn't be lazy here's the source:
https://doc.rust-lang.org/stable/src/alloc/collections/vec_deque/mod.rs.html#2724-2738

You should just use the `VecDeque`'s capacity, compare it against its length. If it's full, pop the front *then* push to the back, otherwise just push to the back

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+
+        while self.order.len() > self.capacity {
+            if let Some(oldest) = self.order.pop_front() {
+                self.seen.remove(&oldest);
+            }
+        }
```
richardpringle: A `while` loop never made sense here since there's only ever one operation...

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+
+        true
+    }
+
+    /// Records timestamped lines and returns whether they are new; untimestamped lines always pass.
```
richardpringle:
```suggestion
    /// Records timestamped lines and returns whether they are new; lines missing timestamps always pass.
```

richardpringle: or "without timestamps"

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+    ///
+    /// NOTE: Mutating filter
+    fn record_if_new(&mut self, replica: &str, line: &LogLine) -> bool {
+        match LogLineKey::new(replica, line) {
+            Some(key) => self.insert(key),
+            None => true,
+        }
+    }
```
richardpringle, line 280: I don't love how the logic is hidden here. Why not make it explicit:

```suggestion
    fn record_if_new(&mut self, replica: &str, line: &LogLine) -> bool {
        let LogLine { content, ts } = line;
        let Some(ts) = ts else { return true };

        let key = LogLineKey::new(replica, ts.clone(), content.clone());

        self.insert(key)
    }
```

with 

```rust
impl LogLineKey {
    fn new(replica: &str, ts: Timestamp, content: String) -> Self {
        Self {
            replica: replica.into(),
            content,
            seconds: ts.seconds,
            nanos: ts.nanos,
        }
    }
}
```

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -0,0 +1,620 @@
...
+        response
+            .entries
+            .iter()
+            .filter_map(|entry| {
+                entry
+                    .line
+                    .as_ref()
+                    .map(|line| (entry.replica_label.as_str(), line))
+            })
+            .filter(|(replica, line)| recent_lines.record_if_new(replica, line))
```
richardpringle: I think this is fine for now, but relying on a timestamp for uniqueness doesn't seem fool proof. I don't know how things work on the other side, but I would expect the logs to be appended to a file and polled based on line number or something. 

Maybe this is a todo?

richardpringle (changes requested):

tvc/src/commands/deploy/debug_logs.rs
richardpringle: It's easy and a best practice, so I would encourage this nit

tvc/src/commands/deploy/debug_logs.rs
richardpringle: definitely think this this is the better way to handle input validation

tvc/src/commands/deploy/debug_logs.rs
richardpringle: This is a nit, and is unlikely to make a difference here as this is never going to be performance critical code, but cloning outside of functions is a best practice. If we (or the LLMs) write code this way, we don't have to put as much thought into optimizing performance critical code, we're already in the habit of using best practices when it comes to performance.

tvc/src/commands/deploy/debug_logs.rs
richardpringle: If it is a UUID, can fix in a follow up

tvc/src/commands/deploy/debug_logs.rs
richardpringle: This goes with the other cloning comment. It's a nit, but I would encourage a change.

tvc/src/commands/deploy/debug_logs.rs
richardpringle: I think we should do this before merging too as the clap validation and defaults makes the rest of the code a little cleaner

tvc/src/commands/deploy/debug_logs.rs
richardpringle: Could deal with this in a follow up

tvc/src/commands/deploy/debug_logs.rs
richardpringle: Could also do this in a follow up

tvc/src/commands/deploy/debug_logs.rs
richardpringle: I think you should fix this before we merge

tvc/src/commands/deploy/debug_logs.rs
richardpringle: I think you should fix this before we merge

tvc/src/commands/deploy/debug_logs.rs
richardpringle: nit... your choice

tvc/src/commands/deploy/debug_logs.rs
richardpringle: could do this in a follow up

richardpringle:
Okay, sorry for the review above that was all over the place, but I went through and commented on the comments that I think should be addressed now vs in follow ups. I made sure _all_ of my suggestions compile when they are all applied.

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: done

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: done

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: `tvc` package wide we just use `String` for id's like this. I will create a quick follow up ticket and have Valet do it for the whole package

daniilrrr: https://linear.app/turnkey/issue/SYS-103/refactor-rust-sdk-cliinternal-ids-to-uuid-where-applicable

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: https://linear.app/turnkey/issue/SYS-103/refactor-rust-sdk-cliinternal-ids-to-uuid-where-applicable

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: 1. yeah, `DebugLogPrinter.RecentLogLines` handles the dedupe
2. i deliberately chose not to do a more robust cursor approach or other and to just mirror `kubectl logs` usage, where we just query the pod logs at a given time and support follow. [K8s logging background.](https://kubernetes.io/docs/concepts/cluster-administration/logging/#how-nodes-handle-container-logs) K8s pod logs don't expose a stable cursor, so we would have had to build it out on the server-side and/or store the logs externally, which we explicitly don't want to do

So this "get current logs or follow (poll) for new logs and dedupe" is the 80/20 solution, in my opinion

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: i like it, done

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: great callout. 

this is a result of how codegen is done from the protos. we could probably do some discerning for GETs and Query's and make it more idiomatic, as well as any other improvements you can think of

daniilrrr: created https://linear.app/turnkey/issue/SYS-104/make-rust-sdk-generated-query-client-request-ownership-idiomatic

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: forgot about `default_value_t`, thank you. done

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: done. keeping `i32` and `i64` because those are the public API bounds via the protos. Note that they should not be different and should be a mirror of Kubernetes `PodLogOptions.TailLines`, and I let slip in an unnecessary mismatch. Should be `i64` all the way around and that was an oversight / sloppiness on my end. 

To not add more toil to this PR I will add a follow up ticket and do it there. `tail_lines` from `int32` -> `int64` will be a breaking but compatible change since the field is being widened

daniilrrr: https://linear.app/turnkey/issue/SYS-105/align-tvc-debug-log-tail-lines-type-with-kubernetes-int64

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: required `capacity` to be positive so this went away

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: done

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: done

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: yep fixed this, thank you for flagging

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: fixed

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: ahhh way better thank you

tvc/src/commands/deploy/debug_logs.rs
daniilrrr: [related to prev comment](https://github.com/tkhq/rust-sdk/pull/169#discussion_r3507854764)

we don't get a log cursor for free and don't want to build it nor store the logs outside of the running kube pod, so this is acceptable for now. Can revisit if those reqs ever change. 

Also, it should be very rare for timestamps to be absent because tvc-observer always requests kubelet timestamps from the running K8s `qos-enclave` container log stream

tvc/src/commands/deploy/debug_logs.rs
richardpringle: >Note that they should not be different and should be a mirror of Kubernetes

They need to be able to serialize and convert properly, but why do our internal binary representations need to mirror k8s?

daniilrrr: we don't **have** to match and can do whatever we want on the client-side, however I don't see the value of narrowing the bounds arbitrarily on the client. I would rather be consistent throughout and just match what we do in `mono`, which is to use the K8s API / Go SDK.

used here - https://github.com/tkhq/mono/blob/644b79d47f189d39165503ff53db1d89caeae1b0/src/go/coordinator/internal/public/tvc_deployment_debug_logs.go#L63-L69 

which is `corev1.PodLogOptions` under the hood here - 
https://github.com/tkhq/mono/blob/b6d6b775777c35703b27eba650ed8cd1c188871f/src/go/tvc-observer/internal/logquery/logs.go#L29

which has both `TailLines` and `SinceSeconds` as `int64` - https://pkg.go.dev/k8s.io/api/core/v1#PodLogOptions

tvc/src/commands/deploy/debug_logs.rs
richardpringle: So... the more I think about this, the more I think this is a bad idea. It's very possible that we get multiple log lines that have the same message and have the same timestamp (from concurrent requests for instance) and that information will be lost. If we can't properly stream logs, we shouldn't build an imperfect workaround into our cli, we should tell our clients to use `watch` while we work on a proper solution. 

If we eventually have enough developers that are all following logs at the same time, we could be adding a bunch of extra work for the k8s api-server. That could already be under heavy load from all the other stuff that we're doing. 

Maybe I'm being paranoid... but at the very least, I think it might make sense to get the larger group's opinion on this

tvc/src/commands/deploy/debug_logs.rs
richardpringle: What about concurrency? At many points in the past, I've relied on duplicate log lines (with the same timestamp) to determine how many concurrent requests were being handled when dealing with various bugs.

tvc/src/commands/deploy/debug_logs.rs
richardpringle: A feature flag could also give us the opportunity to load test this

daniilrrr:
@richardpringle 

Responding to the valid concerns around [dedupe](https://github.com/tkhq/rust-sdk/pull/169#discussion_r3508842244) and [concurrency](https://github.com/tkhq/rust-sdk/pull/169#discussion_r3508860517)

First, a bit of context
- Implementing this as a gRPC Stream on the API side is a more natural fit and was my original plan, however this would have been the very first streaming endpoint in the public API and would have therefore required significantly more time and uplift (middlewares, observability, other unknowns) in the coordinator to be at parity with the existing Unary endpoints, and this would have also been riskier. We chose to do a unary RPC approach for now as the 80/20 solution. 
- @narimonf and I have acknowledged that this endpoint and command will need to be revisited before GA to be made more robust and scalable, as you point out re: K8s api-server. Don't think there is a ticket, just made [ENG-4513](https://linear.app/turnkey/issue/ENG-4513/tvc-deployment-debug-mode-logs-scaling) to track

We know there are limitations with the current API and we hope to improve it in the near future, and we'd like customer usage to dictate the priority. We explicitly chose to do it via unary RPC to ship something usable more quickly. 

Your concern about the case of logs concurrent requests with identical [LogLineKey](https://github.com/tkhq/rust-sdk/blob/82bbb424d28d26dbfb5cd5531d467ff4e9f87cbc/tvc/src/commands/deploy/debug_logs.rs#L207)s (basically same timestamp and content) is valid. In this case a customer should turn off `--follow` and use `watch` and they would be successful. We could make that more clear in docs

Some ideas to address your concerns that we can do in this PR:
- naming: change `--follow` to `--poll` (maybe with configurable interval) to make it more explicit to users that we are not actually streaming logs under the hood, which better informs them of limitations
- we could make `dedupe_logs` default enabled but add a toggle to disable it (or vice versa) in case customers need exact lines, such as in the concurrency use case 
  - we could also remove dedupe altogether, although I think that for the majority of users it is a nice UX improvement

Let me know what you think

richardpringle (dismissed):

daniilrrr:
- renamed `--follow` to `--poll`
- made dedupe default enabled but added `--disable-dedupe` for advanced users / edge cases

richardpringle (approved):

PR #170, opened 2026-06-22
feat(tvc): JSON output mode (TVC-116)

daniilrrr:
## TVC-116: JSON output mode in the `tvc` CLI

Linear: https://linear.app/turnkey/issue/TVC-116/json-output-mode-in-tvc-sdk
Design reference: [cargo's `Shell` + `--message-format` + `Message` pattern](https://gist.github.com/richardpringle/46332d01d74036d4a4b1d1f9bfd7ab18)

### Summary

Adds a machine-readable JSON output mode to `tvc`, adapted from cargo's output architecture (central shell, message-format flag, tagged messages) — simplified for a sequential single-command CLI rather than copied wholesale. All user-facing output is routed through a central `Shell`; machine-relevant payloads are emitted as structured JSON messages with stable `reason` discriminators. Human-readable output remains the default, with
existing command presentation preserved; top-level errors now flow through `Shell` and respect the configured color policy.

This PR is the **foundation layer**: the plumbing, the contract types, and the four commands whose payloads are machine-relevant today. A follow-up branch PR https://github.com/tkhq/rust-sdk/pull/188 builds on this to make **every** command return a typed outcome and adds a `tvc schema` contract-introspection command.

### Architecture

1. **Global flags** (`cli.rs`) — global `--message-format` (`human` | `json`, default `human`) and `--color` (`auto` | `always` | `never`, default `auto`), parsed once at startup. The root long help documents the JSON Lines contract and that JSON mode implies non-interactive operation.

2. **`Ctx<W>` — one context, first parameter everywhere** (`output.rs`) — commands take `ctx: &mut Ctx<W>` as their first argument. This is cargo-inspired (its `gctx: &GlobalContext` convention) but deliberately diverges where it serves us: cargo passes a *shared* reference and wraps its shell in a `Mutex` because parallel build jobs print concurrently; `tvc` runs one command sequentially, so we use `&mut` — exclusivity is enforced by the compiler at zero runtime cost, with no lock to acquire or poison. `Ctx` bundles the `Shell<W>` with cross-cutting flags: `is_non_interactive()` merges `--non-interactive` with the JSON-mode override at construction, so commands check one thing.

3. **`Shell<W: Write>` — the output chokepoint** — generic over one writer type: `Box<dyn Write + Send>` in production, `Vec<u8>` in tests (`Shell::from_write` → `into_stdout()`/`into_stderr()`, no downcasting). Presentation output is explicitly namespaced behind `shell.human()` and is a silent no-op outside human mode. Commands use `shell_println!` / `shell_print!` / `shell_eprintln!` (format-args based, with no per-call allocation), while machine-relevant results use `ctx.shell().emit(&msg)`. Crate-level Clippy `print_stdout` / `print_stderr` denials prevent direct `println!`-family macros in both the library and binary, with actionable guidance for developers.

4. **`Message` trait** — each structured payload implements `reason() -> &'static str` (stable, kebab-case, external contract) and `human_message()`. In JSON mode `emit` writes exactly one newline-terminated JSON object to stdout with `reason` flattened in; in human mode it prints `human_message()`, preserving pre-existing text.

### JSON-mode semantics

- **JSON implies non-interactive**: enforced once in `Ctx::new`. A missing prompt-required input produces one structured `missing-required-input` error on **stdout** and a non-zero exit — no prompts, no hangs, no partial human text.
- **Errors are structured**: all command failures emit a single `command-error` (or `missing-required-input`) JSON object to stdout with exit code 1. Human mode keeps styled errors on stderr.
- **Presentation is suppressed**: `shell_println!`, `shell_print!`, `shell_eprintln!`, and methods on `shell.human()` are silent no-ops in JSON mode; only `emit()`-ed messages reach stdout.

### Stable `reason` strings in this PR

| reason | source |
|---|---|
| `deployment-config-created` | `deploy init` |
| `manifest-approval-generated` | `deploy approve` |
| `re-encrypted-share-generated` | `keys re-encrypt-share` |
| `quorum-key-share-posted` | `deploy post-share` |
| `command-error` | any command failure |
| `missing-required-input` | prompt-required input missing in non-interactive/JSON mode |

### Testing

- In-memory capture via `Shell::from_write(Vec::new(), …)` per the design gist.
- `tests/message_format.rs`: JSON mode emits exactly one parseable tagged line; human mode contains no `reason` envelope; structured error path.
- Existing pty / non-interactive / command suites unchanged and green — human output assertions were the regression net for the refactor.
- Full gates: `cargo build`, `cargo test`, `cargo clippy --all-targets` (zero warnings), `cargo fmt --check`.

### Sample commands (JSON mode)

```console
$ tvc deploy init --output deploy.json --message-format=json
{"reason":"deployment-config-created","command":"deploy init","path":"deploy.json","template":true,"interactive":false,"from_deployment":false,"needs_pull_secret":true}
```

```console
$ tvc deploy init --output deploy.json
Created deployment config template: deploy.json

  tvc deploy create --config-file deploy.json
```

```console
$ tvc login --message-format=json
{"reason":"missing-required-input","code":"missing_required_input","message":"--org is required in non-interactive mode (set --org or run in a TTY without --non-interactive / TVC_NON_INTERACTIVE=true)"}
$ echo $?
1
```

```console
$ tvc keys re-encrypt-share --quorum-key-metadata qkm.json --provision-bundle bundle.json --message-format=json
{"reason":"re-encrypted-share-generated","deploymentId":"…","ephemeralPublicKeyHex":"04…","reEncryptedShare":"…","shareApproval":{"signature":"…","member":{"alias":"operator-1","pubKey":"…"}}}
```


tvc/JSON_OUTPUT_MODE_PLAN_1.md
daniilrrr, line 1: remember to delete these plans

tvc/src/commands/app/create.rs
```diff
@@ -38,18 +40,21 @@ struct Overrides {
     pub dangerous_enable_debug_mode_deployments: bool,
 }
 
-pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
-    let config = if is_non_interactive {
+pub async fn run<W: Write>(ctx: &mut Ctx<W>, args: Args) -> Result<()> {
+    let config = if ctx.is_non_interactive() {
```
richardpringle, line 44: 👍  I like the way you use context here

tvc/src/commands/app/create.rs
richardpringle, line 1: 👍

tvc/src/commands/deploy/approve.rs
```diff
@@ -231,43 +245,69 @@ async fn run_with_resolved_inputs(inputs: ResolvedApproveInputs) -> anyhow::Resu
...
+    fn reason(&self) -> &'static str {
+        "manifest-approval-generated"
+    }
+
+    fn human_message(&self) -> String {
+        // Preserve the previous human behavior: pretty-printed approval JSON.
+        serde_json::to_string_pretty(self.approval)
+            .expect("serializing manifest approval should not fail")
+    }
+}
```
richardpringle, line 330: 👍

tvc/src/commands/deploy/init.rs
```diff
@@ -115,44 +118,73 @@ async fn execute(args: Args) -> Result<()> {
...
+        // The digest and debug mode were copied from the source manifest (the
+        // digest is image-coupled, so it must be recomputed if the image
+        // changes), and a pull secret (if the source used one) cannot be
+        // recovered and must be re-supplied.
+        if self.from_deployment {
+            message.push_str(
+                "\nSeeded from an existing deployment. Before deploying:\n  \
+                 - expectedPivotDigest and debug mode were copied from the source deployment.\n    \
+                 The digest is tied to the container image, so if you change\n    \
+                 pivotContainerImageUrl you MUST recompute the digest to match.\n",
```
richardpringle: nit (which I'm not supposed to be leaving on this review, but everything else looks so good, I gotta have some criticism 😅 )
Prefer `r#""#` instead of escaped newline characters here.

tvc/src/commands/deploy/init.rs
```diff
@@ -115,44 +118,73 @@ async fn execute(args: Args) -> Result<()> {
...
+            }
         }
 
-        println!();
-    }
+        if self.interactive {
+            message.push_str(&format!(
+                "\nRun: tvc deploy create --config-file {}\n\n{PORT_GUIDANCE}",
+                self.path
+            ));
```
richardpringle, line 183: Not suggesting you should do this, but what happens if you just called `message.push_str("")` or `message.push_str("\n")`? (I'll try to remember to reply to my own comment if/when I see the answer in another file)

tvc/src/cli.rs
```diff
@@ -47,68 +50,100 @@ pub struct Cli {
     )]
     non_interactive: bool,
 
+    /// Format user-facing output.
+    #[arg(long, global = true, value_enum, default_value_t = MessageFormat::Human)]
+    message_format: MessageFormat,
```
richardpringle, line 62: 👍

tvc/src/cli.rs
```diff
@@ -47,68 +50,100 @@ pub struct Cli {
...
+
+    /// Control ANSI color in user-facing output.
+    #[arg(long, global = true, value_enum, default_value_t = ColorChoice::Auto)]
+    color: ColorChoice,
```
richardpringle, line 66: 👍

tvc/JSON_OUTPUT_MODE_PLAN_1.md
```diff
@@ -0,0 +1,109 @@
...
+CLI context that owns it):
+
+- `Shell` owns stdout and stderr writers, message format, color policy, and
+  verbosity.
+- Command handlers receive the output dependency explicitly instead of using a
+  global or opening standard streams themselves.
+- Human and JSON rendering decisions are centralized in `Shell` and typed
+  `Message` implementations.
+- TVC command modules must not call `print!`, `println!`, `eprint!`, or
+  `eprintln!` for user-visible output.
```
richardpringle: A very quick google tells me that we can enforce this with Clippy
```rust
#![deny(clippy::print_stdout)]
#![deny(clippy::print_stderr)]
```

However, I didn't verify if this is true. We should do this, then put the `#[allow(...)]` rules in the one file where it makes sense to do that.

tvc/JSON_OUTPUT_MODE_PLAN_1.md
```diff
@@ -0,0 +1,109 @@
...
+- Interactive input is accessed through the invocation context or an injected
+  prompt policy so output mode and prompt behavior cannot diverge.
```
richardpringle: 🚀

tvc/JSON_OUTPUT_MODE_PLAN_1.md
```diff
@@ -0,0 +1,109 @@
...
+- Developer-only diagnostics continue to use `tracing`, not `Shell` messages.
+
+Tests should substitute in-memory writers so command output can be asserted
+without global stream redirection and without serializing test execution.
```
richardpringle: 🙌

tvc/JSON_OUTPUT_MODE_PLAN_1.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Confirmed interactivity policy
+
+JSON mode is always noninteractive, including when stdin is a TTY.
```
richardpringle: 👍

tvc/src/output.rs
```diff
@@ -0,0 +1,421 @@
...
+            }
+        }
+    }
+
+    pub fn line(&mut self, message: impl Display) -> Result<()> {
+        if matches!(self.message_format, MessageFormat::Human) {
+            writeln!(self.stdout, "{message}")?;
+        }
+        Ok(())
+    }
```
richardpringle, line 129: You _might_ not want to make this `pub`. Or you might want to give it a different name that makes it clear that this is a no-op if the message format isn't for humans

richardpringle: Another option would be to do something like 
```rust
shell.human.line("some message")
```
Just so everything is abundantly clear at the call site.

tvc/src/output.rs
```diff
@@ -0,0 +1,421 @@
...
+#[macro_export]
+macro_rules! shell_print {
+    ($shell:expr, $($arg:tt)*) => {
+        $shell.shell().print(format_args!($($arg)*))
+    };
+}
+
+#[macro_export]
+macro_rules! shell_err_line {
+    ($shell:expr, $($arg:tt)*) => {
```
richardpringle: the `$shell:expr` is really context, right?

richardpringle (commented):
Left a few comments, but nothing major. Love the direction. 

I know this is in "Draft", but it looks pretty ready to me!

tvc/src/commands/deploy/init.rs
richardpringle: I didn't see my answer

tvc/src/commands/deploy/init.rs
daniilrrr: done

tvc/src/output.rs
daniilrrr: yes, updated

tvc/src/commands/deploy/init.rs
daniilrrr: `""` would be no-op, `"\n"` would add a line. It all gets written out at `Shell.emit()` and `line()` adds a trailing newline via `writeln!()` 

https://github.com/tkhq/rust-sdk/blob/71ebbe12e9f754d2fa29715a5f7525feab8596e0/tvc/src/output.rs#L104

tvc/JSON_OUTPUT_MODE_PLAN_1.md
daniilrrr: gonna enforce this at `tvc` package level, and we won't need an `allow` because `output.rs` uses `writeln!()` which is fine

daniilrrr: i've also updated to use 
- `shell_println!`
- `shell_print!`
- `shell_eprintln!` 

to be a clearer substitute for what you're supposed to use

tvc/src/output.rs
daniilrrr: really like this! updated

daniilrrr:
hi @emostov or @tkhq/dependency-reviewers, I added `anstyle` as a new direct dep but it was already used transitively by Clap and test tooling, so there is no change to the dependency graph. Seeking your review 🙏

emostov (dismissed):
Dep review approval

tvc/src/output.rs
```diff
@@ -0,0 +1,392 @@
...
+pub trait Message: Serialize {
+    fn reason(&self) -> &'static str;
+
+    fn human_message(&self) -> String;
+
+    fn to_json_string(&self) -> String {
+        #[derive(Serialize)]
+        struct WithReason<'a, S: Serialize + ?Sized> {
+            reason: &'a str,
+            #[serde(flatten)]
```
Avi-D-coder, line 39: non blocking nit (more of a seeking others opinions): to me flatten makes parsing the message object harder since the message object can no longer be modeled as a reason less domain type.

Avi-D-coder (dismissed):
love this! one non blocking nit

tvc/src/output.rs
richardpringle: Depends on what you're doing with it. If you're doing something like using `jq` (which I often do with cargo to filter on files), it's actually a lot more convenient to deal with the flat object. 

I'm in favour of keeping this the way it is strictly because this is what cargo does and cargo is the goat IMO haha. 

I wouldn't be mad if didn't flatten though, so 🤷

tvc/src/commands/app/delete.rs
```diff
@@ -1,8 +1,11 @@
 //! App delete command - marks an app and all deployments for deletion.
 
 use crate::client::build_client;
+use crate::output::Ctx;
+use crate::shell_println;
```
richardpringle, line 5: Non blocking, but prefer merged imports. `rustfmt` has had this as an unstable feature forever 😞

tvc/src/output.rs
daniilrrr: I don't feel strongly either way. At this time I don't think we will be parsing the object much, just outputting it, but I am not sure. If we add more metadata beyond `reason` then maybe the domain type should be a separate level. 

```
  // Flat
  {
    "reason": "deployment-config-created",
    "path": "deploy.json",
    "template": true
  }

  // Nested
  {
    "reason": "deployment-config-created",
    "payload": {
      "path": "deploy.json",
      "template": true
    }
  }
```

Avi-D-coder: non blocking the jq point is a good one feel free to merge

tvc/src/output.rs
```diff
@@ -0,0 +1,392 @@
...
+
+        serde_json::to_string(&WithReason {
+            reason: self.reason(),
+            msg: self,
+        })
+        .expect("serializing TVC output message should not fail")
+    }
+}
+
+pub struct Shell<W: Write = Box<dyn Write + Send>> {
```
richardpringle, line 51: This type is over-constrained. It's rare that you should have to put trait bounds on type generics unless there's a custom `Drop` implementation among some other specific cases. But if the compiler doesn't tell you you have to, you shouldn't. A silly example is [`HashMap<K, V>`](https://doc.rust-lang.org/stable/src/std/collections/hash/map.rs.html#246-253). You can't do much with a map that doesn't have a key type that's `Hash`, but you can create one. 


```suggestion
pub struct Shell<W = Box<dyn Write + Send>> {
```

richardpringle, line 51: Leaving this as a separate comment, but you shouldn't need to use dynamic dispatch here. From what I can tell, the only reason that it's being used is because `Shell` is generic over a single type. Again I would say it's over constrained here. It should be possible to create a `Shell<Stdout, io::Empty>` if you don't care about the error messages. 

I would suggest using type aliases to prevent having to repeatedly write `Shell<Stdout, Stderr>` though

tvc/src/output.rs
```diff
@@ -0,0 +1,392 @@
...
+    stdout: W,
+    stderr: W,
+    color: ColorChoice,
+    use_color: bool,
+    message_format: MessageFormat,
+}
```
richardpringle, line 57: Leaving this as a separate comment, but you shouldn't need to use dynamic dispatch here. From what I can tell, the only reason that it's being used is because `Shell` is generic over a single type. Again I would say it's over constrained here. It should be possible to create a `Shell<Stdout, io::Empty>` if you don't care about the error messages. 

I would suggest using type aliases to prevent having to repeatedly write `Shell<Stdout, Stderr>` though

 
```suggestion
pub type StdShell = Shell<Stdout, Stderr>;

pub struct Shell<Out, Err> {
    stdout: Out,
    stderr: Err,
    color: ColorChoice,
    use_color: bool,
    message_format: MessageFormat,
}
```

tvc/src/output.rs
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    pub fn into_stdout(self) -> Vec<u8> {
+        self.stdout
+    }
+
+    pub fn into_stderr(self) -> Vec<u8> {
+        self.stderr
+    }
+}
```
richardpringle: This is only used in tests and should be behind `#[cfg(test)]`

tvc/src/main.rs
```diff
@@ -1,8 +1,17 @@
+// Keep synchronized with the corresponding lint policy in the other TVC crate root.
+// TVC has separate library and binary crates, so this attribute must appear in both.
+#![deny(
+    clippy::print_stdout,
+    clippy::print_stderr,
+    reason = "direct print macros bypass TVC's output protocol; use shell_println!, shell_print!, or shell_eprintln! for presentation output, Shell::emit for structured output, and tracing for diagnostics"
+)]
```
richardpringle, line 7: This is the old school way of defining crate level lints, we should put this in `Cargo.toml`

tvc/src/lib.rs
```diff
@@ -1,10 +1,19 @@
+// Keep synchronized with the corresponding lint policy in the other TVC crate root.
+// TVC has separate library and binary crates, so this attribute must appear in both.
+#![deny(
+    clippy::print_stdout,
+    clippy::print_stderr,
+    reason = "direct print macros bypass TVC's output protocol; use shell_println!, shell_print!, or shell_eprintln! for presentation output, Shell::emit for structured output, and tracing for diagnostics"
+)]
```
richardpringle, line 7: See comment in main.rs

tvc/src/cli.rs
```diff
@@ -47,68 +57,100 @@ pub struct Cli {
...
+                }
+                ExitCode::FAILURE
+            }
+        }
+    }
+}
 
-        match args.command {
+impl Commands {
+    async fn run<W: Write>(self, ctx: &mut Ctx<W>) -> anyhow::Result<()> {
```
richardpringle, line 107: So this has a bit of a knock on effect, but these functions are only ever called with a single concrete type of `Ctx`. It's better not to make them generic unless you decide to specifically write a test using a `Shell<Vec<u8>>` or `Shell<io::Empty>`.

richardpringle (dismissed):
Okay... since this PR is quite big, I think landing it then following up with https://github.com/tkhq/rust-sdk/pull/190 right after is the move here. 

Please see my comments regardless.

Good job, in general 🏆  - this was a pretty clean refactor

tvc/src/main.rs
daniilrrr: I chose this because  it has the really nice and clear `reason` when Clippy denies things while `Cargo.toml` doesn't, and imo it's not clear to a casual developer why we would ban a pretty normal `println!`, but I can refactor if you feel strongly

tvc/src/commands/app/delete.rs
daniilrrr: agree, but we'd need nightly `rustfmt` for this right? or a CI rule or a precommit hook or something external? 

gonna keep for now since my goto is just `cargo fmt && cargo clippy --all-targets --all-features -- -D warnings` like CI and that's what this produces. If you like you can make a ticket to cover this separately

tvc/src/output.rs
daniilrrr: you've addressed this on https://github.com/tkhq/rust-sdk/pull/190 so I will keep

tvc/src/output.rs
daniilrrr: done, put inside `mod tests{ }`

tvc/src/cli.rs
daniilrrr: right, you covered it in https://github.com/tkhq/rust-sdk/pull/190 so I will keep

emostov (approved):

tvc/src/commands/app/delete.rs
richardpringle: I've worked on a repo that used nightly... it was honestly a pain to always remember to run `cargo +nightly fmt` every time when that wasn't the normal flow elsewhere. 

I think we just do it on a best effort bases. If you're making changes anyway, can tell the LLMs to do it or whatever since it doesn't really matter.

tvc/src/main.rs
richardpringle: Well that's dumb... Let's keep this

richardpringle (approved):

PR #171, opened 2026-06-26
chore(rust-sdk): pin activity versions, sync protos

daniilrrr:
Note: @moeodeh3 discovered a gap for Activity version capping that I have implemented in #172 . That PR should be merged ahead of this one
___
Sync recently updated protos 

Made 1 Manual decision here: https://github.com/tkhq/rust-sdk/pull/171/changes#diff-3f7e738903cc8204a4e3b1c8ab4db5812898e6bb5ad8b6c7b36df42e29481a8fR333

```rust
                ActivityStatus::ConsensusNeeded | ActivityStatus::AuthenticatorsNeeded => {
                    return Err(TurnkeyClientError::ActivityRequiresApproval(activity.id));
                }
```
New `ActivityStatus::AuthenticatorsNeeded` was added but it doesn't have an explicit `TurnkeyClientError` nor message. I lumped it with the existing one
___
Also set the following `Activity` version caps, matching [tkhq/sdk/.../codegen.js](https://github.com/tkhq/sdk/blob/main/packages/sdk-server/scripts/codegen.js#L16)

```json
"pins": {
    "ACTIVITY_TYPE_CREATE_AUTHENTICATORS": "ACTIVITY_TYPE_CREATE_AUTHENTICATORS_V2",
    "ACTIVITY_TYPE_CREATE_API_KEYS": "ACTIVITY_TYPE_CREATE_API_KEYS_V2",
    "ACTIVITY_TYPE_CREATE_POLICY": "ACTIVITY_TYPE_CREATE_POLICY_V3",
    "ACTIVITY_TYPE_CREATE_PRIVATE_KEYS": "ACTIVITY_TYPE_CREATE_PRIVATE_KEYS_V2",
    "ACTIVITY_TYPE_CREATE_SUB_ORGANIZATION": "ACTIVITY_TYPE_CREATE_SUB_ORGANIZATION_V8",
    "ACTIVITY_TYPE_CREATE_USERS": "ACTIVITY_TYPE_CREATE_USERS_V4",
    "ACTIVITY_TYPE_SIGN_RAW_PAYLOAD": "ACTIVITY_TYPE_SIGN_RAW_PAYLOAD_V2",
    "ACTIVITY_TYPE_SIGN_TRANSACTION": "ACTIVITY_TYPE_SIGN_TRANSACTION_V2",
    "ACTIVITY_TYPE_EMAIL_AUTH": "ACTIVITY_TYPE_EMAIL_AUTH_V3",
[...]

PR #172, opened 2026-06-26
feat(codegen): explicit activity-version pinning

daniilrrr:
## Summary

Adds an explicit activity-version pinning mechanism to Rust SDK codegen, mirroring the TypeScript SDK's `VERSIONED_ACTIVITY_TYPES` map.

## Problem

Today, proto sync can auto-bump generated activity submission methods to whatever `ACTIVITY_TYPE_*` literal is present in the proto annotation, even if that activity version has not been released to production yet.

## Mechanism

- Adds `codegen/activity_version_caps.json` with an explicit `pins` map.
- The map keys are base activity families, e.g. `ACTIVITY_TYPE_SIGN_TRANSACTION`.
- The map values are the exact `ACTIVITY_TYPE_*` literals codegen should emit, e.g. `ACTIVITY_TYPE_SIGN_TRANSACTION_V2`.
- Codegen strips trailing `_V<number>` suffixes to derive the base family, checks for a pin, and overrides the emitted activity type before looking up intent/result types in `proto/activities.json`.
- Pin targets are validated against `activities.json`, and pins that cross base activity families are rejected.

## No-op default

This PR seeds the config with empty pins, so it is mechanism-only and generated client output remains unchanged. The actual pins should be set in #171.

## How to add a pin

Add an entry to `codegen/activity_version_caps.json`:

```json
{
  "pins": {
    "ACTIVITY_TYPE_SIGN_TRANSACTION": "ACTIVITY_TYPE_SIGN_TRANSACTION_V2"
  }
}
```

The target activity version must already exist in `proto/activities.json` and must already be released in production.

## Ordering

This should merge before #171 so that PR can add/bump pins using this mechanism.

## Verification

- `PATH="/workspace/.cargo/bin:$PATH" make generate`
- `git diff --stat -- client/src/generated` produced no output
- `PATH="/workspace/.cargo/bin:$PATH" make lint`
- `PATH="/workspace/.cargo/bin:$PATH" make test`

### DIFF
--- a/codegen/activity_version_caps.json
+++ b/codegen/activity_version_caps.json
@@ -1,0 +1,4 @@
+{
+  "$comment": "Explicit activity-version pins for codegen. Maps a base ACTIVITY_TYPE_* (the family identity) to the EXACT ACTIVITY_TYPE_* literal codegen should emit, overriding the version from the proto annotation. An activity NOT listed here emits whatever the proto annotation specifies (latest). Adding/bumping a pin is a deliberate, reviewed opt-in to a specific activity version that MUST already be released in production. Mirrors the TS SDK VERSIONED_ACTIVITY_TYPES map.",
+  "pins": {}
+}
--- a/codegen/activity_version_caps.json
+++ b/codegen/activity_version_caps.json
@@ -1,4 +1,3 @@
 {
-  "$comment": "Explicit activity-version pins for codegen. Maps a base ACTIVITY_TYPE_* (the family identity) to the EXACT ACTIVITY_TYPE_* literal codegen should emit, overriding the version from the proto annotation. An activity NOT listed here emits whatever the proto annotation specifies (latest). Adding/bumping a pin is a deliberate, reviewed opt-in to a specific activity version that MUST already be released in production. Mirrors the TS SDK VERSIONED_ACTIVITY_TYPES map.",
   "pins": {}
 }
--- a/codegen/src/main.rs
+++ b/codegen/src/main.rs
@@ -1,7 +1,7 @@
 use regex::Regex;
 use std::collections::{HashMap, HashSet};
 use std::fs;
-use std::io::Write;
+use std::io::{ErrorKind, Write};
 use std::path::PathBuf;
 use walkdir::WalkDir;
 
@@ -11,6 +11,7 @@ const PUBLIC_API_PROTO_PATH: &str = "proto/services/coordinator/public/v1/public
 const EXTERNAL_ACTIVITY_PROTO_PATH: &str = "proto/external/activity/v1/activity.proto";
 const INCLUDE_PROTO_PATH: &str = "proto";
 const ACTIVITIES_MAPPING_PATH: &str = "proto/activities.json";
+const ACTIVITY_VERSION_CAPS_PATH: &str = "codegen/activity_version_caps.json";
 const GENERATED_CLIENT_DIR: &str = "client/src/generated";
 const CLIENT_TEMPLATE_PATH: &str = "codegen/src/client.template.rs";
 
@@ -30,6 +31,12 @@ struct ActivityDetails {
 #[derive(Debug, serde::Deserialize)]
 struct ActivitiesFile {
     activities: HashMap<String, ActivityDetails>,
+}
+
+#[derive(Debug, Default, serde::Deserialize)]
+struct ActivityVersionCaps {
+    #[serde(default)]
+    pins: HashMap<String, String>,
 }
 
 #[derive(Debug, PartialEq, Eq)]
@@ -96,6 +103,7 @@ fn main() {
         fs::read_to_string(ACTIVITIES_MAPPING_PATH).expect("cannot read activities.json");
     let parsed_activities: ActivitiesFile =
         serde_json::from_str(&activities_mapping_data).expect("cannot parse activities.json");
+    let activity_version_caps = load_activity_version_caps(ACTIVITY_VERSION_CAPS_PATH);
 
     let external_activity_proto = fs::read_to_string(EXTERNAL_ACTIVITY_PROTO_PATH)
         .expect("Failed to read external activity proto file");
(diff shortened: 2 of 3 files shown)


daniilrrr:
More detail, from AI: 

## Prior art / reference

This mirrors the existing version-capping pattern in the TypeScript SDK repo `tkhq/sdk`. In the TS SDK, the capping map lives at the top of each codegen script as a const named `VERSIONED_ACTIVITY_TYPES`:

- `tkhq/sdk: packages/sdk-types/scripts/codegen.js`
- `tkhq/sdk: packages/sdk-server/scripts/codegen.js`
- `tkhq/sdk: packages/core/scripts/codegen.js`
- `tkhq/sdk: packages/sdk-browser/scripts/codegen.js`

Concrete references:

- `tkhq/sdk: packages/sdk-server/scripts/codegen.js#L16` — the map (string form): e.g. `ACTIVITY_TYPE_SIGN_TRANSACTION: "ACTIVITY_TYPE_SIGN_TRANSACTION_V2"`, and notably `ACTIVITY_TYPE_ETH_SEND_TRANSACTION: "ACTIVITY_TYPE_ETH_SEND_TRANSACTION"` (deliberately pinned to V1).
- `tkhq/sdk: packages/sdk-server/scripts/codegen.js#L311` — the consumer: `type: "${versionedActivityType ?? unversionedActivityType}"` (use the pinned version if present, else the raw/unversioned value).
- `tkhq/sdk: packages/core/scripts/codegen.js#L578-L582` — same idea with an explicit comment: "Use the capped version if one exists, otherwise use the raw swagger value" → `VERSIONED_ACTIVITY_TYPES[unversionedActivityType]?.[0] ?? rawActivityType`.
- `tkhq/sdk: packages/sdk-types/scripts/codegen.js#L21-L122` — the 3-element array form `[activityType, intentType, resultType]` (sdk-types also pins intent/result type names).
- Git history precedent: commit `56c730b41` ("Lock ethsendtransaction version") — the exact precedent for deliberately holding an activity at an older version.

### How the Rust approach differs

Our Rust approach mirrors the same idea — an explicit `base activity family → exact activity type` map applied as an override before intent/result lookup — but consolidates it to a **single source file** (`codegen/activity_version_caps.json`) instead of the TS SDK's duplicated `VERSIONED_ACTIVITY_TYPES` map across 4 codegen scripts. A pin therefore only has to be set once. The downstream `proto/activities.json` lookup keys off the (possibly pinned) activity type, so the correct intent/result types come along automatically rather than being pinned separately.

codegen/src/main.rs
```diff
@@ -154,9 +162,23 @@ fn main() {
...
+                let proto_activity_type = request_to_activity_type
                         .get(short_req_type)
                         .unwrap_or_else(|| panic!("no activity type annotation found for {short_req_type} in external activity proto"));
+                let activity_type = resolve_activity_type(
+                    proto_activity_type,
+                    &activity_version_caps.pins,
+                    &parsed_activities.activities,
+                )
+                .unwrap_or_else(|err| panic!("{err}"));
+                if activity_type != *proto_activity_type {
```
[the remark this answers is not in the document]
daniilrrr: this is fine and the code compiles 

> This compiles because Rust comparison operators borrow their operands for PartialEq; it does not move out of proto_activity_type.

Declaring the variable like 
`let moved = *proto_activity_type;` 
would fail

codegen/src/main.rs
```diff
@@ -312,6 +334,48 @@ fn main() {
...
+            return Err(format!(
+                "activity version cap for {base} points to {target}, but base families differ"
+            ));
+        }
+        if !activities.contains_key(target) {
+            return Err(format!(
+                "activity version cap for {base} points to {target}, which was not found in activities.json"
+            ));
+        }
+    }
```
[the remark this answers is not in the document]
daniilrrr: this is a minor quibble because the pins map will be tiny, however I split out `validate` from `resolve/lookup` for extra clarity

codegen/src/main.rs
```diff
@@ -556,4 +620,82 @@ message BazRequest {
...
+            activity_base("ACTIVITY_TYPE_SIGN_TRANSACTION"),
+            "ACTIVITY_TYPE_SIGN_TRANSACTION"
+        );
+        assert_eq!(
+            activity_base("ACTIVITY_TYPE_SIGN_TRANSACTION_V10"),
+            "ACTIVITY_TYPE_SIGN_TRANSACTION"
+        );
+        assert_eq!(
+            activity_base("ACTIVITY_TYPE_EXPORT_WALLET_VAULT"),
+            "ACTIVITY_TYPE_EXPORT_WALLET_VAULT"
```
richardpringle, line 640: 👍 nice

codegen/src/main.rs
```diff
@@ -312,6 +332,50 @@ fn main() {
...
+    }
+
+    Ok(())
+}
+
+fn resolve_activity_type(activity_type: &str, pins: &HashMap<String, String>) -> String {
+    pins.get(activity_base(activity_type))
+        .cloned()
+        .unwrap_or_else(|| activity_type.to_string())
+}
```
richardpringle: I know performance isn't crucial here, but it's simple enough to avoid the clones. 

```suggestion
fn resolve_activity_type<'a>(activity_type: &'a str, pins: &'a HashMap<String, String>) -> &'a str {
    pins.get(activity_base(activity_type))
        .map(String::as_str)
        .unwrap_or(activity_type)
}
```

Will require removing one other `as_str` call on the result

codegen/activity_version_caps.json
```diff
@@ -0,0 +1,4 @@
+{
+  "$comment": "Explicit activity-version pins for codegen. Maps a base ACTIVITY_TYPE_* (the family identity) to the EXACT ACTIVITY_TYPE_* literal codegen should emit, overriding the version from the proto annotation. An activity NOT listed here emits whatever the proto annotation specifies (latest). Adding/bumping a pin is a deliberate, reviewed opt-in to a specific activity version that MUST already be released in production. Mirrors the TS SDK VERSIONED_ACTIVITY_TYPES map.",
+  "pins": {}
+}
```
richardpringle, line 3: Why not `activity_version_caps.rs`? There wouldn't be any need for the `$comment` hack, you could have actual documentation and instead of a `HashMap` that you have to resolve at runtime, you could simply have a `match` statement that's correct by design at compile time.

Is the answer because of validation? Is that because `proto/activities.json` is generated? 

I guess validation requires you to do things at runtime here, but maybe we can create a ticket to fix this and move as much as possible to compile time.

richardpringle (changes requested):
Bit of a nit, but a really easy/quick fix

codegen/src/main.rs
daniilrrr: done, thanks

codegen/activity_version_caps.json
daniilrrr: I would like to keep it as `.json` so that we align our process with how the other SDKs are pinned (although I acknowledge that the others are JS so JSON is more natural there). 

I have removed the `$comment` and moved the text to the `codegen/README` which is a better home for it

richardpringle (approved):

PR #173, opened 2026-06-27
feat(tvc): flexible --operator-seed source (env/stdin/file)

prasanna-anchorage:
## What
`tvc deploy approve` (and `tvc keys re-encrypt-share`) currently take `--operator-seed <PATH>` — the operator master seed must be written to a file on disk. This makes `--operator-seed` (env `TVC_OPERATOR_SEED`) accept a **source spec** so the seed needn't touch disk:

- `env:NAME` — read the hex seed from environment variable `NAME`
- `stdin` (or `-`) — read the hex seed from stdin
- `file://<path>` — read from a file
- bare `<path>` — read from a file (**back-compatible** with today's behavior)

Unset still falls back to the logged-in org operator key. Resolution happens in one place (`operator_key::load_operator_pair`) via a small `read_seed_source` helper.

## Why
For CI/automation, requiring a file means materializing the operator seed onto disk (then trying to scrub it). Passing it via **stdin** keeps the secret off disk *and* out of `argv`/shell history; `env:` is convenient where a file is undesirable. This is a small quality-of-life + secret-hygiene improvement.

## Notes
- **No new dependencies** — `std` only (`from_hex_seed` already existed on `LocalPair`).
- Back-compatible: a bare path is still treated as a seed file.
- `cargo fmt` + `cargo clippy -p tvc` clean; `cargo check -p tvc` passes.
- Caveat: use `stdin` with non-interactive approval (`--dangerous-skip-interactive`), since stdin is consumed for the seed.

### DIFF
--- a/tvc/src/operator_key.rs
+++ b/tvc/src/operator_key.rs
@@ -1,15 +1,20 @@
-//! Resolve an operator's QOS key pair from CLI args or the active org config.
+//! Resolve an operator's QOS key pair from a seed SOURCE or the active org config.
 
 use crate::config::turnkey::{Config, StoredQosOperatorKey};
 use crate::pair::LocalPair;
-use anyhow::anyhow;
-use std::path::Path;
+use anyhow::{Context, anyhow};
+use std::io::Read;
 
-/// Load the operator's QOS key pair, preferring an explicit seed file path
-/// and falling back to the operator key stored under the active org config.
-pub async fn load_operator_pair(operator_seed: Option<&Path>) -> anyhow::Result<LocalPair> {
-    match operator_seed {
-        Some(path) => LocalPair::from_master_seed(path).await,
+/// Load the operator's QOS key pair from a seed `source` spec, or fall back to
+/// the operator key stored under the active org config when `source` is `None`.
+///
+/// The source is flexible so the seed needn't be written to disk:
+///   - `env:NAME`             read the hex seed from environment variable `NAME`
+///   - `stdin` (or `-`)       read the hex seed from stdin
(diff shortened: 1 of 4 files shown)


tvc/src/operator_key.rs
```diff
@@ -29,3 +34,24 @@ pub async fn load_operator_pair(operator_seed: Option<&Path>) -> anyhow::Result<
...
+/// previous `--operator-seed <PATH>` behavior).
+fn read_seed_source(spec: &str) -> anyhow::Result<String> {
+    if spec == "stdin" || spec == "-" {
+        let mut s = String::new();
+        std::io::stdin()
+            .read_to_string(&mut s)
+            .context("read operator seed from stdin")?;
+        Ok(s)
+    } else if let Some(name) = spec.strip_prefix("env:") {
+        std::env::var(name).with_context(|| format!("operator seed env var '{name}' is not set"))
```
emostov, line 49: The TVC_OPERATOR_SEED env var should be used when the cli arg is not specified. I don't see a problem with this as is though, just maybe slightly inconsistent behavior

emostov:
I think this makes sense. Would just like to see some updated e2e smoke tests. @prasanna-anchorage is it ok if I push to to this?

prasanna-anchorage:
> I think this makes sense. Would just like to see some updated e2e smoke tests. @prasanna-anchorage is it ok if I push to to this?

sure, I didn't have access to run any workflows, but feel free to push to this. I'll resolve some of the merge conflicts

richardpringle:
@prasanna-anchorage we just merged #195 so I'm going to close this

PR #179, opened 2026-07-02
feat(tvc): seed `deploy init` config from an existing deployment

richardpringle:
Closes ENG-3887.

## What

Adds `tvc deploy init --from-deployment <DEPLOY_ID>` (env `TVC_FROM_DEPLOYMENT`) to scaffold a new deployment config seeded from an existing deployment, reusing the existing **init → edit → create** flow. The user gets a pre-filled, editable `DeployConfig` JSON to feed to the unchanged `tvc deploy create --config-file`.

## Field mapping (`TvcDeployment` → `DeployConfig`)

| DeployConfig field | Source |
|---|---|
| `app_id`, `qos_version` | deployment top-level |
| `pivot_container_image_url` | `pivot_container.container_url` |
| `pivot_path`, `pivot_args` | `pivot_container` |
| `health_check_type` / `health_check_port` / `public_ingress_port` | `pivot_container` (u32→u16 checked) |
| `expected_pivot_digest` | **`<FILL_IN…>` placeholder** — forced re-entry |
| `dangerous_deploy_debug_mode` | **`false`** — safe default |
| `pivot_container_encrypted_pull_secret` | `Some(PLACEHOLDER)` iff `has_pull_secret`, else `None` |

## Why the digest/debug handling

The digest and debug mode aren't server-derivable (they live only in the manifest), so the digest is left blank to force re-entry, debug mode resets to off, and a loud warning is printed. The pull-secret plaintext is unrecoverable, so a private-image source writes a placeholder to prompt re-supply via `--pivot-pull-secret`.

## Changes

- `config/deploy.rs` — `DeployConfig::from_deployment(&TvcDeployment)` pure mapping leaf + unit tests.
- `client.rs` — `fetch_tvc_deployment` helper; `deploy get-status` refactored to reuse it.
- `commands/deploy/init.rs` — `--from-deployment` flag; composes with `--interactive` / `--non-interactive`; prints digest/image-coupling warning + pull-secret note.
- `cli.rs` — updated `deploy init` help.
- `tests/deploy_init_help.rs` — arg-parse/help test.

## Out of scope (possible follow-ups)

- `create --from-deployment` one-shot recreate.
- Recovering the exact digest/debug-mode by parsing the manifest.

## Testing

Ran through the quickstart to create a deployment, then used that deployment as the `--from-deployment` seed to create another.

### DIFF
--- a/tvc/src/cli.rs
+++ b/tvc/src/cli.rs
@@ -160,7 +160,8 @@ enum DeployCommands {
         after_help = commands::deploy::PORT_GUIDANCE
     )]
     Create(commands::deploy::create::Args),
-    /// Generate a template deployment configuration file.
+    /// Generate a deployment configuration file
+    #[command(long_about = commands::deploy::init::LONG_ABOUT)]
     Init(commands::deploy::init::Args),
     /// Fetch debug logs for a deployment.
     #[command(long_about = commands::deploy::debug_logs::LONG_ABOUT)]
--- a/tvc/src/commands/deploy/get_status.rs
+++ b/tvc/src/commands/deploy/get_status.rs
@@ -27,7 +27,7 @@ pub async fn run(args: Args) -> anyhow::Result<()> {
         deploy_id: deployment_id,
     } = args;
 
-    // TODO:
+    // TODO (TVC-154):
     // this is a little backwards, all the variables that are needed to build the client
     // are resolved INSIDE the `build-client`. Instead, all the necessary data for the client
     // should be resolved, then passed into an infallible constructor
--- a/tvc/src/commands/deploy/get_status.rs
+++ b/tvc/src/commands/deploy/get_status.rs
@@ -2,11 +2,15 @@
 
 use anyhow::{Context, anyhow};
 use clap::Args as ClapArgs;
-use turnkey_client::generated::external::data::v1::{AppStatus, DeploymentStatus};
-use turnkey_client::generated::{GetAppStatusRequest, GetTvcDeploymentRequest};
+use turnkey_client::generated::{
+    GetAppStatusRequest,
+    external::data::v1::{AppStatus, DeploymentStatus},
+};
 
-use crate::client::fetch_tvc_app;
-use crate::commands::display::format_egress_enabled;
+use crate::{
+    client::{fetch_tvc_app, fetch_tvc_deployment},
+    commands::display::format_egress_enabled,
+};
 
 /// Get the live status of a deployment from the app status API.
 #[derive(Debug, ClapArgs)]
@@ -19,25 +23,21 @@ pub struct Args {
 
 /// Run the deploy get-status command.
 pub async fn run(args: Args) -> anyhow::Result<()> {
+    let Args {
+        deploy_id: deployment_id,
+    } = args;
+
+    // TODO:
+    // this is a little backwards, all the variables that are needed to build the client
+    // are resolved INSIDE the `build-client`. Instead, all the necessary data for the client
+    // should be resolved, then passed into an infallible constructor
     let auth = crate::client::build_client().await?;
+    let org_id = auth.org_id.clone();
 
-    let deployment_request = GetTvcDeploymentRequest {
-        organization_id: auth.org_id.clone(),
-        deployment_id: args.deploy_id.clone(),
-    };
-
-    let deployment_response = auth
-        .client
-        .get_tvc_deployment(deployment_request)
-        .await
-        .context("failed to fetch deployment")?;
-
-    let deployment = deployment_response
-        .tvc_deployment
-        .ok_or_else(|| anyhow!("deployment not found: {}", args.deploy_id))?;
+    let deployment = fetch_tvc_deployment(&auth, org_id.clone(), deployment_id.clone()).await?;
 
     let app_request = GetAppStatusRequest {
-        organization_id: auth.org_id.clone(),
+        organization_id: org_id.clone(),
         app_id: deployment.app_id.clone(),
     };
 
@@ -59,13 +59,13 @@ pub async fn run(args: Args) -> anyhow::Result<()> {
     println!("{}", format_egress_enabled(app.enable_egress));
     println!(
         "Is Targeted Deployment: {}",
-        if app_status.targeted_deployment_id == args.deploy_id {
+        if app_status.targeted_deployment_id == deployment_id {
             "yes"
         } else {
             "no"
         }
     );
-    if let Some(deployment_status) = find_deployment_status(&app_status, &args.deploy_id) {
+    if let Some(deployment_status) = find_deployment_status(&app_status, &deployment_id) {
         println!(
             "{}",
             crate::commands::app_status::format_replica_status(deployment_status)
--- a/tvc/src/commands/deploy/init.rs
+++ b/tvc/src/commands/deploy/init.rs
@@ -77,8 +77,8 @@ async fn execute(args: Args) -> Result<()> {
     // Seed the config either from an existing deployment or a blank template.
     let mut config = match from_deployment {
         Some(deploy_id) => {
-            // TODO:
-            // see other TODO. TL;DR split fetching the data/resources separately
+            // TODO (TVC-154):
+            // TL;DR split fetching the data/resources separately
             // from building the client
             let auth = build_client().await?;
             let org_id = auth.org_id.clone();
--- a/tvc/src/commands/deploy/init.rs
+++ b/tvc/src/commands/deploy/init.rs
@@ -1,13 +1,23 @@
 //! Deploy init command - generates a template config file.
 
 use super::PORT_GUIDANCE;
-use crate::config::deploy::DeployConfig;
-use crate::config::turnkey;
-use crate::prompts::{bail_interactive_conflicts_with_non_interactive, ensure_stdin_is_tty};
+use crate::{
+    client::{build_client, fetch_tvc_deployment},
+    config::{deploy::DeployConfig, turnkey},
+    prompts::{bail_interactive_conflicts_with_non_interactive, ensure_stdin_is_tty},
+};
 use anyhow::{Context, Result, bail};
 use chrono::Local;
 use clap::Args as ClapArgs;
 use std::path::PathBuf;
+
+pub(crate) const LONG_ABOUT: &str = r#"
+Generate a deployment config file to edit, then pass to `tvc deploy create`.
+
+--from-deployment <DEPLOY_ID> copies every field from that deployment, including
+the expected pivot digest and debug mode (read from its manifest); a private-image
+pull secret cannot be recovered and must be re-supplied. Without the flag, a blank
+placeholder template is written."#;
 
 /// Generate a template deployment configuration file.
 #[derive(Debug, ClapArgs)]
@@ -16,6 +26,15 @@ pub struct Args {
     /// Output file path.
     #[arg(short, long, value_name = "PATH", env = "TVC_DEPLOY_CONFIG_OUT")]
     pub output: Option<PathBuf>,
+
+    /// Seed the config from an existing deployment instead of a blank template.
+    ///
+    /// Fetches the deployment and copies every recoverable field, including the
+    /// expected pivot digest and debug mode (read from its manifest). The digest
+    /// is tied to the container image, so recompute it if you change the image. A
+    /// pull secret, if the source used one, must be re-supplied.
+    #[arg(long, value_name = "DEPLOY_ID", env = "TVC_FROM_DEPLOYMENT")]
+    pub from_deployment: Option<String>,
 
     /// Walk through prompts for each field and write a filled config instead
     /// of a placeholder template.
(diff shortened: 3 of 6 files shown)


tvc/src/commands/deploy/init.rs
```diff
@@ -17,6 +19,15 @@ pub struct Args {
     #[arg(short, long, value_name = "PATH", env = "TVC_DEPLOY_CONFIG_OUT")]
     pub output: Option<PathBuf>,
 
+    /// Seed the config from an existing deployment instead of a blank template.
+    ///
+    /// Fetches the deployment and copies every recoverable field. The expected
+    /// pivot digest is intentionally left blank (it is tied to the container
+    /// image and must be recomputed if the image changes), and debug mode is
+    /// reset to off. A pull secret, if the source used one, must be re-supplied.
```
richardpringle: Do we want to turn debug-mode off or keep it as is if it is on?

tvc/src/cli.rs
```diff
@@ -160,7 +160,8 @@ enum DeployCommands {
         after_help = commands::deploy::PORT_GUIDANCE
     )]
     Create(commands::deploy::create::Args),
-    /// Generate a template deployment configuration file.
+    /// Generate a deployment configuration file, blank or seeded from an
+    /// existing deployment via `--from-deployment <DEPLOY_ID>`.
```
daniilrrr: nit - confusing comment, can remove imo and explain better in the CLI help text

tvc/src/config/deploy.rs
```diff
@@ -36,6 +38,66 @@ pub struct DeployConfig {
...
+            health_check_type,
+            health_check_port,
+            public_ingress_port,
+        } = pivot_container
+            .ok_or_else(|| anyhow!("deployment {} has no pivot container spec", id))?;
+
+        let health_check_port = u16::try_from(health_check_port)
+            .context("deployment health check port does not fit in u16")?;
+        let public_ingress_port = u16::try_from(public_ingress_port)
+            .context("deployment public ingress port does not fit in u16")?;
```
daniilrrr, line 90: port numbers max at 2^16 so we were wrong here to use u32 for those vals in `TvcContainerSpec`. annoying impossible error path, oh well

tvc/src/commands/deploy/get_status.rs
```diff
@@ -19,25 +23,21 @@ pub struct Args {
 
 /// Run the deploy get-status command.
 pub async fn run(args: Args) -> anyhow::Result<()> {
+    let Args {
+        deploy_id: deployment_id,
+    } = args;
+
+    // TODO:
```
daniilrrr: can you create a ticket here pls

tvc/src/commands/deploy/init.rs
daniilrrr: An App must have debug_mode enabled for Deployments with debug_mode to succeed. To regain nonzero PCRs and TEE capabilities for running in Prod, you would need to create a fresh app.

So,
- If we copy a debug deployment into the SAME App, that App has already been opted in to debug mode. So nothing is lost
- If we copy a debug deployment into a NEW non-debug app, the deployment will just fail. 

So either way, the user can't "accidentally" enter debug mode through these paths and would need to opt in or opt out elsewhere.

Do with this what you will, I think either is fine.

If it were up to me, I would copy debug mode as-is. I can see a flow where a user is onboarding and wants to stay in debug mode as they iterate on code changes / build new images / create fresh deployments. So having to flip `debug_mode` -> `true` each time would be slightly more annoying.

tvc/src/commands/deploy/init.rs
```diff
@@ -17,6 +19,15 @@ pub struct Args {
     #[arg(short, long, value_name = "PATH", env = "TVC_DEPLOY_CONFIG_OUT")]
     pub output: Option<PathBuf>,
```
daniilrrr, line 28: noticed that the variable name diverges from the ENV name format of `TVC_$variable_name`, can we fix this in this PR?

tvc/src/commands/deploy/init.rs
```diff
@@ -47,31 +64,82 @@ async fn execute(args: Args) -> Result<()> {
         bail!("File already exists: {}", output.display());
```
daniilrrr, line 72: I think this check should come before creating the `output` PathBuf above. currently reads a bit confusingly to me to, in  the absent case, create a fresh PathBuf that definitely doesn't exist and then check if it exists

tvc/src/commands/deploy/init.rs
```diff
@@ -47,31 +64,82 @@ async fn execute(args: Args) -> Result<()> {
...
+    // Seed the config either from an existing deployment or a blank template.
+    let mut config = match from_deployment {
+        Some(deploy_id) => {
+            // TODO:
+            // see other TODO. TL;DR split fetching the data/resources separately
+            // from building the client
+            let auth = build_client().await?;
+            let org_id = auth.org_id.clone();
+            let deployment = fetch_tvc_deployment(&auth, org_id, deploy_id).await?;
+            DeployConfig::try_from(deployment)?
```
daniilrrr, line 86: if this fails for some reason, like selecting an invalid deployment, I think we should error to the user that they used a bad `DEPLOYMENT_ID` but then fall back to the infallible None case so that they get at least some JSON config for their trouble

tvc/src/commands/deploy/init.rs
```diff
@@ -47,31 +64,82 @@ async fn execute(args: Args) -> Result<()> {
...
     } else {
         println!("Created deployment config template: {}", output.display());
+    }
+    println!();
+
+    // Print the reminders specific to seeding a config from an existing deployment:
+    // the image-coupled digest was left blank on purpose, and a pull secret (if the
+    // source used one) cannot be recovered and must be re-supplied.
+    if is_from_deployment {
+        println!("Seeded from an existing deployment. Before deploying:");
```
daniilrrr, line 130: nit - would be nice to include "Seeded from an existing deployment with ID: $deployment_id"

tvc/src/commands/deploy/init.rs
```diff
@@ -47,31 +64,82 @@ async fn execute(args: Args) -> Result<()> {
...
+
+        if needs_pull_secret {
+            println!(
+                "  - The source deployment used a private image. Its pull secret cannot be\n    \
+                 recovered; pass `--pivot-pull-secret <PATH>` to `tvc deploy create` (or\n    \
+                 remove pivotContainerEncryptedPullSecret if the new image is public)."
+            );
+        }
+
+        println!("  - Debug mode was reset to off; re-enable it explicitly if you need it.");
```
daniilrrr: reminder to update this depending on your decision re copying debug mode elsewhere

tvc/src/cli.rs
daniilrrr: speaking of, I think this command would benefit from a LONG_ABOUT

daniilrrr (commented):
looks good! have some suggestions

tvc/src/commands/deploy/init.rs
daniilrrr: i would change the variable name to `pub deploy_config_path` or similar

tvc/src/commands/deploy/init.rs
richardpringle: >If it were up to me, I would copy debug mode as-is.

Yeah, I think this makes sense too, will change

richardpringle: 57eba4a

tvc/src/cli.rs
richardpringle: 46a4211

tvc/src/commands/deploy/get_status.rs
richardpringle: Already done: https://linear.app/turnkey/issue/TVC-154

daniilrrr: nice. 

I meant to tag the issue name in-code like 

```rust
// TODO (TVC-154)
```

so we don't miss it

tvc/src/commands/deploy/init.rs
richardpringle: I would rather do it in a separate PR. Can stick an LLM on verifying that all options conform to the convention. Created a ticket here: https://linear.app/turnkey/issue/TVC-165

tvc/src/commands/deploy/init.rs
richardpringle: You're suggesting I do something like
```rust
let output = match output {
    Some(path) if path.exists() => bail!("..."),
    Some(path) => path,
    None => PathBuf::from("..."),
};
```
?

I don't I'm not opposed, but I don't think this change is out of scope of this PR. All I did was rename the variable here.

tvc/src/commands/deploy/init.rs
richardpringle: I think it would be confusing/annoying to call `tvc deploy init --from-deployment <id>` and have the _blank_ template created. It should fail, then I should be able to make the decision whether I want to inspect why it's failing, maybe try to figure out what's wrong, then create a new deployment. I think just creating a file on failure would lead to creating files that the user might not know exactly what they want to do with right away. Remember, they have to explicitly opt-in to using `--from-deployment` and they have probably put some effort into finding the id of the deployment they want to copy. The only failure cases are when there's no manifest or pivot container information, which shouldn't ever be the case. 

Imagine a scenario where a user doesn't realize they're not logged into the correct org. They would just be creating blank templates, instead of getting an error message which might push them into looking into why things are failing

tvc/src/commands/deploy/init.rs
richardpringle: This is what it looks like to run the command. I really don't think it's necessary to print the id again. 
```
$ tvc deploy init --from-deployment 0a4ebd2d-4e90-420f-8750-bd6ba97a984c --output seed.json
Created deployment config template: seed.json

Seeded from an existing deployment. Before deploying:
  - expectedPivotDigest and debug mode were copied from the source deployment.
    The digest is tied to the container image, so if you change
    pivotContainerImageUrl you MUST recompute the digest to match.

  tvc deploy create --config-file seed.json

Port guidance:
  publicIngressPort is the port inside your container that receives public app
  traffic.
  healthCheckPort is the port TVC probes to decide whether the deployment is
  healthy.
  Use the same port for both unless your binary exposes health checks on a
  separate listener. Generated configs default both to 3000.a
```

I do think the message could use some iteration, but I would rather get the feature out first (I think the message is a little verbose).

richardpringle:
Pushed changes above (timeline). Going to squash commits after (can use `compare` feature to verify no changes).

richardpringle:
whoops... change above is just pulling in changes in main

tvc/src/cli.rs
```diff
@@ -160,7 +160,9 @@ enum DeployCommands {
         after_help = commands::deploy::PORT_GUIDANCE
     )]
     Create(commands::deploy::create::Args),
-    /// Generate a template deployment configuration file.
+    /// Generate a deployment configuration file either from an existing deployment
+    /// with `--from-deployment` or from a predefined template.
```
daniilrrr: i would revert this change to the comment, think this level of detail doesn't belong here and is already covered elsewhere

tvc/src/commands/deploy/init.rs
daniilrrr: yep what you wrote is clear. I'll leave it to you whether to do in this PR

tvc/src/commands/deploy/init.rs
```diff
@@ -47,31 +72,82 @@ async fn execute(args: Args) -> Result<()> {
...
 
-    // Generate template (optionally walking prompts to fill it in)
-    let mut config = DeployConfig::template(last_app_id.as_deref());
-    if args.interactive {
-        config.fill_interactively(last_app_id.as_deref())?;
+    // Seed the config either from an existing deployment or a blank template.
+    let mut config = match from_deployment {
+        Some(deploy_id) => {
+            // TODO:
+            // see other TODO. TL;DR split fetching the data/resources separately
```
daniilrrr: "other TODO" is not enough context. add the ticket number, which I think is TVC-154

tvc/src/commands/deploy/init.rs
daniilrrr: that makes sense to me. maybe add a crystal clear `.with_context()` here? 

```suggestion
                        DeployConfig::try_from(deployment).with_context(|| format!("Config failed to be created from this existing deployment: {deploy_id}"))?

```

daniilrrr (dismissed):
non-blocking nits

tvc/src/commands/deploy/init.rs
richardpringle: I forced an error here and checkout how it displays:
```
$ tvc deploy init --from-deployment 0a4ebd2d-4e90-420f-8750-bd6ba97a984c --output seed.json
Error: Config failed to be created from this existing deployment: 0a4ebd2d-4e90-420f-8750-bd6ba97a984c

Caused by:
    0: deployment public ingress port does not fit in u16
    1: out of range integral type conversion attempted
```

It's redundant to display the ID when it's a required argument to the command. If this were a running application, I would agree with you, but when you look at the only way to produce the error, displaying the ID again is redundant

richardpringle: I do think that we should go through and sanity check these (happy paths and sad paths)
[TVC-169](https://linear.app/turnkey/issue/TVC-169)

daniilrrr: a little redundant yes but I think it's beneficial to be explicit if we want this SDK to be consumed/used by agents eventually. However I won't block

daniilrrr (approved):

PR #184, opened 2026-07-06
Am/feat/remove delete org profiles

am-hernandez:
## Summary

Adds `tvc login delete` to remove a saved login profile.

## What it does

- Deletes the profile's config entry and its API + operator key files on disk.
- Resolves the target by alias or org ID via `--org`; prompts with a picker when no arg is given.
- Requires confirmation; `--yes` to confirm in non-interactive mode.

## Example

Run `tvc login delete --org <YOUR_ORG_ALIAS>` to target a specific alias and you will see the following:

```
WARNING: This permanently deletes login profile <YOUR_ORG_ALIAS> (<YOUR_ORG_ID>).
  - Removes the local config entry and deletes the API and operator key
    files from disk. This cannot be undone.
  - It does NOT touch the Turnkey dashboard (https://app.turnkey.com). If this API
    key is registered there, it stays valid until you remove it
    (instructions are printed after deletion).

? Permanently delete profile <YOUR_ORG_ALIAS> (<YOUR_ORG_ID>) and its key files? (y/N)
```

Answering the last line in the affirmative to continue with deletion:

```
> Permanently delete profile <YOUR_ORG_ALIAS> (<YOUR_ORG_ID>) and its key files? Yes

Deleted login profile <YOUR_ORG_ALIAS> (<YOUR_ORG_ID>).

Removed key files:
  ~/.config/turnkey/orgs/<YOUR_ORG_ALIAS>/api_key.json
  ~/.config/turnkey/orgs/<YOUR_ORG_ALIAS>/operator.json

IMPORTANT: This did not touch the Turnkey dashboard. If this API key is still registered
there, it remains valid until you remove it. If you haven't already:
  1. Go to https://app.turnkey.com/dashboard/v2/users and click your user
  2. Delete the API key with public key (if present):
       <API_PUBLIC_KEY>

```

Closes [TVC-158](https://linear.app/turnkey/issue/TVC-158/add-login-delete-command-to-tvc-cli)

## Safety / clarity

- Loud pre-confirm `WARNING`: the deletion is permanent and does **not** touch the Turnkey dashboard.
- Post-delete reminder to revoke the API key in the dashboard — links the env-specific dashboard (prod/preprod/dev) and names the key's public key. Wording is hedged since the command never inspects dashboard state.


tvc/src/commands/login.rs
```diff
@@ -26,6 +26,20 @@ pub struct Args {
...
+/// key files on disk.
+#[derive(Debug, ClapArgs)]
+#[command(about, long_about = None)]
+pub struct DeleteArgs {
+    /// Organization alias or ID of the profile to delete.
+    /// If not provided, will prompt interactively.
+    #[arg(long, value_name = "ORG")]
+    pub org: Option<String>,
+    /// Skip the confirmation prompt (required to delete in non-interactive mode).
+    #[arg(long)]
```
richardpringle: We should also add a short option

tvc/src/commands/login.rs
```diff
@@ -61,6 +75,175 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+        prompts::confirm_or_bail(
+            &format!("Permanently delete profile '{alias}' ({org_id}) and its key files?"),
+            "deletion",
+        )?;
+    }
+
+    let removed = config
+        .remove_org(&alias)
+        .expect("alias was resolved from config");
+    config.save().await?;
```
richardpringle: Should wait to save. If there are any issues with loading the api-key, we'll end up with a weird state.

tvc/src/commands/login.rs
```diff
@@ -61,6 +75,175 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+        .await
+        .ok()
+        .flatten()
+        .map(|key| key.public_key);
+
+    let mut deleted = Vec::new();
+    for path in [&removed.api_key_path, &removed.operator_key_path] {
+        match tokio::fs::remove_file(path).await {
+            Ok(()) => deleted.push(path.display().to_string()),
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
```
richardpringle: Should emit a warning here

tvc/src/commands/login.rs
```diff
@@ -61,6 +75,175 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+/// with a picker when no query is given. Bails if no profiles are configured or
+/// the query matches none.
+fn resolve_profile_alias(
+    config: &Config,
+    org: Option<String>,
+    is_non_interactive: bool,
+) -> Result<String> {
+    if config.orgs.is_empty() {
+        bail!("No login profiles are configured. Run `tvc login` to add one.");
+    }
```
richardpringle: This function is only used in one place for profile deletion. This message doesn't make sense in that context. Was the plan to use this elsewhere?

tvc/src/commands/login.rs
```diff
@@ -61,6 +75,175 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+            Some((alias, _)) => Ok(alias.clone()),
+            None => bail!(
+                "Login profile '{query}' not found. \
+                 Run `tvc login` to see configured profiles."
+            ),
+        },
+        None => {
+            if is_non_interactive {
+                return Err(error_required_in_non_interactive("--org"));
+            }
```
richardpringle: It's a better practice to validate your inputs _before_ performing any business logic. It makes things a lot easier to change and reason about if validation happens before anything else. Validation should also be clearly separate from business logic.

tvc/src/commands/login.rs
richardpringle: Further, `find_org` already handles this, so this is redundant

tvc/src/commands/login.rs
```diff
@@ -61,6 +75,175 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+        .map(|org| org.id.clone())
+        .unwrap_or_default();
+
+    if !args.yes {
+        if is_non_interactive {
+            bail!(
+                "Refusing to delete profile '{alias}' ({org_id}) without confirmation. \
+                 Pass --yes to confirm in non-interactive mode."
+            );
+        }
```
richardpringle: Validate your inputs before performing any business logic ([explanation](https://github.com/tkhq/rust-sdk/pull/184/changes#r3537894256), hopefully the links work, if not, I left another comment below)

tvc/src/commands/login.rs
```diff
@@ -61,6 +75,175 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+    }
+
+    // A local delete does not touch or inspect the dashboard-registered API key.
+    // We can't tell whether one is still registered (the user may have already
+    // removed it), so surface the steps without asserting its state.
+    let dashboard_url = dashboard_base_url(&removed.api_base_url);
+    println!();
+    println!(
+        "IMPORTANT: This did not touch the Turnkey dashboard. If this API key is still registered"
```
```suggestion
        "IMPORTANT: The API key may still be registered on the dashboard."
```
Then something like "it will remain valid until it is manually removed" (with the link, etc).

tvc/src/commands/login.rs
```diff
@@ -61,6 +75,175 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+    );
+    println!("there, it remains valid until you remove it. If you haven't already:");
+    println!("  1. Go to {dashboard_url}/dashboard/v2/users and click your user");
+    match api_public_key {
```
richardpringle, line 186: Side note: we should have a discovery endpoint on the api server so that we aren't manually writing these paths.

richardpringle: Ticket created: https://linear.app/turnkey/issue/TVC-160/serve-dashboard-route-map-from-a-single-endpoint-instead-of-hardcoding

tvc/src/config/turnkey/mod.rs
```diff
@@ -188,6 +188,24 @@ impl Config {
...
+    pub fn remove_org(&mut self, alias: &str) -> Option<OrgConfig> {
+        let removed = self.orgs.remove(alias)?;
+        debug!(org_alias = alias, "removing organization config");
+        self.last_created_app_id.remove(alias);
+        self.last_operator_ids.remove(alias);
+        if self.active_org.as_deref() == Some(alias) {
+            self.active_org = None;
+        }
+        Some(removed)
+    }
```
richardpringle, line 207: 👍

tvc/src/cli.rs
```diff
@@ -105,15 +105,28 @@ impl Cli {
                     commands::keys::re_encrypt_share::run(args).await
                 }
             },
-            Commands::Login(args) => commands::login::run(args, non_interactive).await,
+            Commands::Login { command, args } => match command {
+                None => commands::login::run(args, non_interactive).await,
+                Some(LoginCommands::Delete(delete_args)) => {
+                    commands::login::run_delete(delete_args, non_interactive).await
+                }
```
richardpringle: I don't think `tvc login delete` makes sense to be honest. I think profile management should be done separately. Would it be a huge lift to move this to `tvc profile delete` or `tvc org delete`?

tvc/src/commands/login.rs
```diff
@@ -61,6 +75,175 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+            alias: alias.to_string(),
+        }
+    }
+}
+
+impl std::fmt::Display for ProfileChoice {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.display)
+    }
+}
```
richardpringle: This is an anti-pattern. The purpose of the `Display` trait is so that you can effectively lazily allocate display messages

A better way to do this is as follows: 
```suggestion
struct ProfileChoice<'a> {
    alias: &'a str,
    org_id: &'a str,
    is_active: bool,
}

impl<'a> std::fmt::Display for ProfileChoice<'a> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        let suffix = if self.is_active { " (active)" } else { "" };

        write!(f, "{} {}{suffix}", self.alias, self.org_id)
    }
}
```

Note, I removed the `new` constructor, since it's actually more clear to just set the variables and we don't have to rely on positional arguments

tvc/src/commands/login.rs
```diff
@@ -61,6 +75,175 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+    // dashboard-revocation reminder below can name exactly which key to remove.
+    // Best-effort: a missing or unreadable key file just omits the value.
+    let api_public_key = StoredApiKey::load(&removed)
+        .await
+        .ok()
+        .flatten()
+        .map(|key| key.public_key);
+
+    let mut deleted = Vec::new();
+    for path in [&removed.api_key_path, &removed.operator_key_path] {
```
richardpringle: Looking at the way we store things by default, the operator key and and api-keys are stored in the directory by default. We should just delete the directory then update the config. We can make the config self-healing on load, if there are missing directories from the config, we delete them from the config and save the config before we finish loading.

If custom paths were used for the api-key and operator key, we should not delete them at all. We should leave a warning for the user.

richardpringle (changes requested):
So, I think we need to do a little more thinking about how we could end up in a broken state and if there are any broken states from which we can self-heal. 

I understand the motivation here, but I'm not sure that this is so important to automate right now (vs manually deleting directories and editing the config)

tvc/src/cli.rs
am-hernandez: I agree and had considered this previously but ultimately went with the login subcommand because it also surfaced orgs. 

I do prefer `tvc profile delete` over `org delete` and the current implementation. One org can have many aliases, and deleting an org by ID would delete an associated alias at random (see [TVC-159](https://linear.app/turnkey/issue/TVC-159/fix-login-alias-matching-for-orgs-with-multiple-aliases)). Though `org delete` would also take an alias I think this behvior needs to be adjusted to where `org` means org ID and `profile`, or even `alias`, means alias. 

If sticking with current terminology I think `tvc alias delete` may be best here.

tvc/src/commands/login.rs
am-hernandez: this is great!

tvc/src/commands/login.rs
am-hernandez: definitely, will fix this.

tvc/src/commands/login.rs
am-hernandez: I am adding this as `-o`, and `-y`, but I think we need another PR to standardize these options. 

I found `-o` used for both `--output` and `--approval-out` in `deploy init` and `deploy approve` respectively. Those long options should both be `--output` and I think `-org` should be short for `--organization`.  Created a ticket for this, [TVC-170](https://linear.app/turnkey/issue/TVC-170/standardize-tvc-cli-shortlong-option-names-o-org).

am-hernandez: resolved here: 34b50b0431cf4a4523b108f740aaf74df783c5a7

tvc/src/commands/login.rs
am-hernandez: resolved here: 73e4b3d63328db8353213f31d8d29316564477a5

tvc/src/commands/login.rs
am-hernandez: resovled here: f0058e1cc713a0dca508e3cfa0d35b656ba87aeb

tvc/src/commands/login.rs
am-hernandez: resolved here: b39c00393a8ad552f6d3a5747c9d816dcab9c707

tvc/src/commands/login.rs
am-hernandez: resolved here: 8b526d4fc2dac57aa5215d79767f95e4a782062e

tvc/src/commands/login.rs
am-hernandez: resolved here: 53a86de6d995b198134df838eef06670b5c27d54

tvc/src/commands/login.rs
am-hernandez: Thanks for flagging and explaining this! Resolved here: bdf85fc0bba6e4a8a1133ff378f988e93370d12a

tvc/src/commands/login.rs
am-hernandez: Resolved here: 983b457787f68d0a73ec6870a94d8b1b3b8a6943

Created a ticket for the self-healing behavior: [TVC-171](https://linear.app/turnkey/issue/TVC-171/self-heal-tvc-config-on-load-prune-org-entries-whose-key-directory-is), for a followup.

tvc/src/cli.rs
am-hernandez: thought some more and looking through the code, `login profile delete` looks best here so I implemented that. 

Resolved here: 027685dfdc73a4663a4ecd3c64264793857de7f8

tvc/src/commands/login.rs
```diff
@@ -61,6 +75,172 @@ pub async fn run(args: Args, is_non_interactive: bool) -> Result<()> {
...
+    // Validate inputs before any business logic: non-interactive mode cannot
+    // prompt, so it requires --org (which profile) and --yes (confirmation).
+    if is_non_interactive {
+        if args.org.is_none() {
+            return Err(error_required_in_non_interactive("--org"));
+        }
+        if !args.yes {
+            return Err(error_required_in_non_interactive("--yes"));
+        }
+    }
```
richardpringle, line 97: 👍

richardpringle (approved):

PR #188, opened 2026-07-11
feat(tvc): typed Outcome for every command

daniilrrr:
## TVC-116 (part 2): every command emits a typed Outcome

Builds on #170 (merged) and is rebased onto current `main` — including #190's `StdCtx` cleanup and the operator/hosted-key refactors (versioned operator registry, the `local_*` key-command renames, and `keys create-quorum-key`). Command entry points take `&mut StdCtx` and return a typed result that the CLI emits centrally.

### What this adds

Before this work, most commands succeeded **silently** in `--message-format=json` — e.g. `tvc app init --message-format=json` wrote the file, printed nothing, exited 0. Now every command emits a typed, tagged outcome.

- **`Outcome` enum** (`src/outcome.rs`): one variant per command, ordered to mirror the CLI tree. Every command's `run()` returns `Result<Outcome>`, so a command with no outcome doesn't compile. `cli.rs` emits it centrally, symmetric with the existing error path. On success the CLI fails open — if writing the outcome fails (e.g. a broken pipe into `jq`), it warns on stderr but still exits `0`, since the operation already succeeded.
- **Envelope/payload split**: per-command payload structs implement `Display` (human rendering) + `Serialize` (JSON fields) only. `Message` — the "emittable" trait — is implemented **solely** by the top-level envelopes: `Outcome` (terminal), `DebugLogLine` (streaming), and the error types. A bare payload therefore can't be emitted; it must be wrapped in an `Outcome`. `Outcome::reason()` is one explicit match — the single source of truth for the reason vocabulary (24 terminal reasons, plus the `debug-log-line` streaming reason and 2 error reasons).
- **Wide-event model**: success = exactly one tagged JSON line on stdout (`json.loads(stdout)` always works). Sole exception: `deploy debug-logs` streams NDJSON `debug-log-line` messages — which also fixes it previously printing *nothing* in JSON mode, including the log lines themselves.
- **`keys create-quorum-key`** (new on `main`) is wired into the `Outcome` model like every other command.
- **Contract fix**: `deployment-config-created` emits `fromDeployment`/`needsPullSecret` (camelCase).
- **Tests**: reason uniqueness + kebab-case checks; XOR-branch serialization tests for the multi-shape `deploy approve` payloads; and human-output goldens that lock the existing human rendering byte-for-byte. (Moving reason-uniqueness to compile-time enforcement via a proc-macro is tracked as a follow-up, TVC-207.)
- **Conventions**: message DTOs derive `Default` (4 `ActivityStatus`-bearing structs need manual impls — the generated enum lacks `Default` upstream); each file's tests live in its single `mod tests` (no `#[cfg(test)]` scatter in `src`).
- **Repo guidance**: `tvc/AGENTS.md` (with a `tvc/CLAUDE.md` symlink) codifies the conventions used here — raw string literals for multi-line output, prefer moving over cloning, exhaustive destructuring of generated types, and short `use` imports over fully-qualified paths.

> An earlier revision of this PR also added a `tvc schema` command (plus a `schemars` dependency and a schema-catalog snapshot test). That has been dropped — the approach was poorly fitted, and machine-readable self-onboarding is better served by a future `mcp serve`.

### Samples

```console
$ tvc app init --output app.json --message-format=json     # was silent before this PR
{"reason":"app-config-created","command":"app init","path":"app.json","template":true,"interactive":false}
```

```console
$ tvc deploy approve --deploy-id deploy-123 --operator-id operator-456 --dangerous-skip-interactive --message-format=json
{"reason":"manifest-approval-posted","approval":{"signature":"30440220...","member":{"alias":"operator-primary","pubKey":"04ab..."}},"manifestId":"manifest-789","operatorId":"operator-456","approvalIds":["approval-abc"],"quorumReached":true}
```

The posted outcome identifies the manifest and operator, returns the approval IDs created by the API, and reports whether the deployment's manifest-approval quorum is now satisfied. Without `--approval-out` it also includes the signed approval inline; with `--approval-out approval.json`, that payload is replaced by `"writtenTo":"approval.json"`.

## Reviews packet

For structured review: https://reviews.figitaki.dev/r/lYghULlT

### DIFF
--- a/tvc/AGENTS.md
+++ b/tvc/AGENTS.md
@@ -19,6 +19,13 @@ Default guidance for coding-agent runs in this repository.
   `From<T>` over `From<&T>`. When a clone is genuinely needed, make it explicit
   at the call site — e.g. `value.clone().into()` or
   `items.iter().cloned().map(Into::into)`.
+- Prefer short (imported) names over fully-qualified paths. Add a `use` and
+  write `impl Display for Foo { fn fmt(&self, f: &mut Formatter<'_>) … }` rather
+  than `impl std::fmt::Display for Foo { … std::fmt::Formatter … }`. Only keep a
+  longer/module-qualified form when it disambiguates from another in-scope name —
+  e.g. `fmt::Result` stays qualified (via `use std::fmt::{self, Display, Formatter}`)
+  so it doesn't collide with `anyhow::Result`, and `std::fmt::Write` may need
+  `as _` where `std::io::Write` is also in scope.
 - When converting from an external/generated type (e.g. the API's `TvcApp`,
   `TvcDeployment`, `AppStatus`) into one of our own structs, destructure it
   exhaustively — `let Foo { a, b, c: _ } = value;` with no trailing `..` —
--- a/tvc/AGENTS.md
+++ b/tvc/AGENTS.md
@@ -1,0 +1,30 @@
+# AGENTS.md
+
+Default guidance for coding-agent runs in this repository.
+
+## Coding style
+
+- Prefer raw string literals (`r#"..."#`) over escaped quotation marks (`\"`) or
+  escaped newlines (`\n`) for any nontrivial string or multi-line output (e.g.
+  `human_message` bodies, help text, JSON fixtures, test goldens). Lay the text
+  out on real lines so it reads as it renders. If a line ends in significant
+  trailing whitespace, still use a raw literal but add a comment noting the
+  trailing whitespace so editors/formatters don't silently strip it.
+- Prefer moving owned values over cloning them. If you already own a value and
+  this is its last use (e.g. building the return value at the end of a
+  function), move it out — use `.into_iter()` or partial field moves instead of
+  `.clone()`.
+- Don't hide clones inside functions. A function or `From`/`TryFrom` impl should
+  take an owned value (`T`, not `&T`) rather than clone internally; prefer
+  `From<T>` over `From<&T>`. When a clone is genuinely needed, make it explicit
+  at the call site — e.g. `value.clone().into()` or
+  `items.iter().cloned().map(Into::into)`.
+- When converting from an external/generated type (e.g. the API's `TvcApp`,
+  `TvcDeployment`, `AppStatus`) into one of our own structs, destructure it
+  exhaustively — `let Foo { a, b, c: _ } = value;` with no trailing `..` —
+  rather than reading fields with `value.a`. Bind the fields you use and
+  `_`-bind the ones you don't. This way, when the upstream type gains a field, the destructure
+  fails to compile and forces a deliberate decision about whether the new field
+  belongs in our output — instead of it being silently dropped. Skip this only
+  where it adds noise for no value, e.g. reading one or two fields off a large
+  API response result.
--- a/tvc/CLAUDE.md
+++ b/tvc/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/tvc/src/cli.rs
+++ b/tvc/src/cli.rs
@@ -1,6 +1,7 @@
 //! CLI parsing and dispatch.
 
 use crate::commands;
+use crate::outcome::Outcome;
 use crate::output::{ColorChoice, Ctx, ErrorMessage, MessageFormat, Shell, StdCtx};
 use clap::{ArgAction, Parser, Subcommand, builder::BoolishValueParser};
 use std::io::Write;
@@ -85,7 +86,16 @@ impl Cli {
         let mut ctx = Ctx::new(shell, args.non_interactive);
         let result = args.command.run(&mut ctx).await;
         match result {
-            Ok(()) => ExitCode::SUCCESS,
+            Ok(outcome) => {
+                // Fail open: the command itself already succeeded, so a failure
+                // to deliver the outcome message should not flip the exit code.
+                // Make a best-effort warning on stderr and still exit successfully.
+                if let Err(emit_error) = ctx.shell().emit(&outcome) {
+                    let mut stderr = std::io::stderr();
+                    let _ = writeln!(stderr, "warning: failed to write CLI output: {emit_error}");
+                }
+                ExitCode::SUCCESS
+            }
             Err(error) => {
                 let shell = ctx.shell();
                 let emit_result = if shell.message_format().is_json() {
@@ -104,7 +114,7 @@ impl Cli {
 }
 
 impl Commands {
-    async fn run(self, ctx: &mut StdCtx) -> anyhow::Result<()> {
+    async fn run(self, ctx: &mut StdCtx) -> anyhow::Result<Outcome> {
         match self {
             Commands::Deploy { command } => match command {
                 DeployCommands::Approve(args) => commands::deploy::approve::run(ctx, args).await,
--- a/tvc/src/commands/app/create.rs
+++ b/tvc/src/commands/app/create.rs
@@ -7,13 +7,13 @@ use crate::{
         turnkey::{self, StoredQosOperatorKey},
     },
     outcome::Outcome,
-    output::{Ctx, Message, StdCtx},
+    output::{Ctx, StdCtx},
     prompts, shell_println,
 };
 use anyhow::{Context, Result, anyhow, bail};
 use clap::Args as ClapArgs;
 use serde::Serialize;
-use std::fmt::Write as _;
+use std::fmt::{self, Display, Formatter};
 use std::time::{SystemTime, UNIX_EPOCH};
 use std::{
     io,
@@ -281,13 +281,10 @@ pub struct AppCreated {
     config_path: String,
 }
 
-impl Message for AppCreated {
-    fn reason(&self) -> &'static str {
-        "app-created"
-    }
-
-    fn human_message(&self) -> String {
-        let mut message = format!(
+impl Display for AppCreated {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
             r#"
 App created successfully!
 
@@ -295,26 +292,24 @@ App ID: {}
 Name: {}
 Manifest Set ID: {}"#,
             self.app_id, self.name, self.manifest_set_id
-        );
+        )?;
 
         if !self.manifest_set_operator_ids.is_empty() {
-            let _ = write!(
-                message,
+            write!(
+                f,
                 "\nManifest Set Operator IDs: {}",
                 self.manifest_set_operator_ids.join(", ")
-            );
+            )?;
         }
 
-        let _ = write!(
-            message,
+        write!(
+            f,
             r#"
 Config: {}
 
 Use one of the Manifest Set Operator IDs above with `tvc deploy approve --operator-id`"#,
             self.config_path
-        );
-
-        message
+        )
     }
 }
 
--- a/tvc/src/commands/app/create.rs
+++ b/tvc/src/commands/app/create.rs
@@ -6,11 +6,14 @@ use crate::{
         app::{AppConfig, AppConfigValidationErrors, OperatorSetParams},
         turnkey::{self, StoredQosOperatorKey},
     },
-    output::{Ctx, StdCtx},
+    outcome::Outcome,
+    output::{Ctx, Message, StdCtx},
     prompts, shell_println,
 };
 use anyhow::{Context, Result, anyhow, bail};
 use clap::Args as ClapArgs;
+use serde::Serialize;
+use std::fmt::Write as _;
 use std::time::{SystemTime, UNIX_EPOCH};
 use std::{
     io,
@@ -51,7 +54,7 @@ struct Overrides {
     pub dangerous_enable_debug_mode_deployments: bool,
 }
 
-pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
+pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
     let config = if ctx.is_non_interactive() {
         build_app_config_non_interactive(&args).await?
     } else {
@@ -233,7 +236,7 @@ async fn load_saved_operator_public_key() -> Option<String> {
     Some(operator_key.public_key)
 }
 
-async fn run_with_config(ctx: &mut StdCtx, args: Args, app_config: AppConfig) -> Result<()> {
+async fn run_with_config(ctx: &mut StdCtx, args: Args, app_config: AppConfig) -> Result<Outcome> {
     shell_println!(ctx, "Creating app '{}'...", app_config.name)?;
 
     let auth = build_client().await?;
@@ -259,27 +262,60 @@ async fn run_with_config(ctx: &mut StdCtx, args: Args, app_config: AppConfig) ->
     config.set_last_operator_ids(&operator_ids)?;
     config.save().await?;
 
-    shell_println!(ctx)?;
-    shell_println!(ctx, "App created successfully!")?;
-    shell_println!(ctx)?;
-    shell_println!(ctx, "App ID: {app_id}")?;
-    shell_println!(ctx, "Name: {}", app_config.name)?;
-    shell_println!(ctx, "Manifest Set ID: {}", result.result.manifest_set_id)?;
-    if !operator_ids.is_empty() {
-        shell_println!(
-            ctx,
-            "Manifest Set Operator IDs: {}",
-            operator_ids.join(", ")
-        )?;
-    }
-    shell_println!(ctx, "Config: {}", args.config_file.display())?;
-    shell_println!(ctx)?;
-    shell_println!(
-        ctx,
-        "Use one of the Manifest Set Operator IDs above with `tvc deploy approve --operator-id`"
-    )?;
+    Ok(Outcome::AppCreate(AppCreated {
+        app_id,
+        name: app_config.name,
+        manifest_set_id: result.result.manifest_set_id,
+        manifest_set_operator_ids: operator_ids,
+        config_path: args.config_file.display().to_string(),
+    }))
+}
 
-    Ok(())
+#[derive(Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AppCreated {
+    app_id: String,
+    name: String,
+    manifest_set_id: String,
+    manifest_set_operator_ids: Vec<String>,
+    config_path: String,
+}
+
+impl Message for AppCreated {
+    fn reason(&self) -> &'static str {
+        "app-created"
+    }
+
+    fn human_message(&self) -> String {
+        let mut message = format!(
+            r#"
+App created successfully!
+
+App ID: {}
+Name: {}
+Manifest Set ID: {}"#,
+            self.app_id, self.name, self.manifest_set_id
+        );
+
+        if !self.manifest_set_operator_ids.is_empty() {
+            let _ = write!(
+                message,
+                "\nManifest Set Operator IDs: {}",
+                self.manifest_set_operator_ids.join(", ")
+            );
+        }
+
+        let _ = write!(
+            message,
+            r#"
+Config: {}
+
+Use one of the Manifest Set Operator IDs above with `tvc deploy approve --operator-id`"#,
+            self.config_path
+        );
+
+        message
+    }
 }
 
 fn build_create_tvc_app_intent(app_config: &AppConfig) -> CreateTvcAppIntent {
--- a/tvc/src/commands/app/delete.rs
+++ b/tvc/src/commands/app/delete.rs
@@ -2,10 +2,11 @@
 
 use crate::client::build_client;
 use crate::outcome::Outcome;
-use crate::output::{Message, StdCtx};
+use crate::output::StdCtx;
 use anyhow::{Context, Result};
 use clap::Args as ClapArgs;
 use serde::Serialize;
+use std::fmt::{self, Display, Formatter};
 use std::time::{SystemTime, UNIX_EPOCH};
 use turnkey_client::generated::{ActivityStatus, DeleteTvcAppAndDeploymentsIntent};
 
@@ -65,13 +66,10 @@ impl Default for AppDeleted {
     }
 }
 
-impl Message for AppDeleted {
-    fn reason(&self) -> &'static str {
-        "app-deleted"
-    }
-
-    fn human_message(&self) -> String {
-        format!(
+impl Display for AppDeleted {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
             r#"App delete accepted.
 App and deployments marked for deletion.
 
--- a/tvc/src/commands/app/delete.rs
+++ b/tvc/src/commands/app/delete.rs
@@ -1,12 +1,13 @@
 //! App delete command - marks an app and all deployments for deletion.
 
 use crate::client::build_client;
-use crate::output::StdCtx;
-use crate::shell_println;
+use crate::outcome::Outcome;
+use crate::output::{Message, StdCtx};
 use anyhow::{Context, Result};
 use clap::Args as ClapArgs;
+use serde::Serialize;
 use std::time::{SystemTime, UNIX_EPOCH};
-use turnkey_client::generated::DeleteTvcAppAndDeploymentsIntent;
+use turnkey_client::generated::{ActivityStatus, DeleteTvcAppAndDeploymentsIntent};
 
 /// Delete a TVC application and all of its deployments.
 #[derive(Debug, ClapArgs)]
@@ -18,7 +19,7 @@ pub struct Args {
 }
 
 /// Run the app delete command.
-pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
+pub async fn run(_ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
     let auth = build_client().await?;
 
     let intent = DeleteTvcAppAndDeploymentsIntent {
@@ -36,11 +37,48 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
         .await
         .context("failed to delete TVC app and deployments")?;
 
-    shell_println!(ctx, "App delete accepted.")?;
-    shell_println!(ctx, "App and deployments marked for deletion.")?;
-    shell_println!(ctx, "App ID: {}", result.result.app_id)?;
-    shell_println!(ctx, "Activity ID: {}", result.activity_id)?;
-    shell_println!(ctx, "Activity Status: {:?}", result.status)?;
+    Ok(Outcome::AppDelete(AppDeleted {
+        app_id: result.result.app_id,
+        activity_id: result.activity_id,
+        activity_status: result.status,
+    }))
+}
 
-    Ok(())
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AppDeleted {
+    app_id: String,
+    activity_id: String,
+    /// Stable proto status name, e.g. `ACTIVITY_STATUS_COMPLETED`.
+    activity_status: ActivityStatus,
+}
+
+/// Manual because the generated `ActivityStatus` does not implement
+/// `Default`; the zero value is the enum's `Unspecified` variant.
+impl Default for AppDeleted {
+    fn default() -> Self {
+        Self {
+            app_id: String::default(),
+            activity_id: String::default(),
+            activity_status: ActivityStatus::Unspecified,
+        }
+    }
+}
+
+impl Message for AppDeleted {
+    fn reason(&self) -> &'static str {
+        "app-deleted"
+    }
+
+    fn human_message(&self) -> String {
+        format!(
+            r#"App delete accepted.
+App and deployments marked for deletion.
+
+App ID: {}
+Activity ID: {}
+Activity Status: {}"#,
+            self.app_id, self.activity_id, self.activity_status.as_str_name()
+        )
+    }
 }
--- a/tvc/src/commands/app/delete.rs
+++ b/tvc/src/commands/app/delete.rs
@@ -78,7 +78,9 @@ App and deployments marked for deletion.
 App ID: {}
 Activity ID: {}
 Activity Status: {}"#,
-            self.app_id, self.activity_id, self.activity_status.as_str_name()
+            self.app_id,
+            self.activity_id,
+            self.activity_status.as_str_name()
         )
     }
 }
--- a/tvc/src/commands/app/init.rs
+++ b/tvc/src/commands/app/init.rs
@@ -2,11 +2,12 @@
 
 use crate::config::app::AppConfig;
 use crate::config::turnkey::{Config, StoredQosOperatorKey};
-use crate::output::StdCtx;
+use crate::outcome::Outcome;
+use crate::output::{Message, StdCtx};
 use crate::prompts::{bail_interactive_conflicts_with_non_interactive, ensure_stdin_is_tty};
-use crate::shell_println;
 use anyhow::{Context, Result, bail};
 use clap::Args as ClapArgs;
+use serde::Serialize;
 use std::path::PathBuf;
 
 /// Generate a template app configuration file.
@@ -30,7 +31,7 @@ pub struct Args {
 }
 
 /// Run the app init command.
-pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
+pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
     if args.interactive {
         if ctx.is_non_interactive() {
             bail_interactive_conflicts_with_non_interactive()?;
@@ -38,10 +39,10 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
             ensure_stdin_is_tty()?;
         }
     }
-    execute(ctx, args).await
+    execute(args).await
 }
 
-async fn execute(ctx: &mut StdCtx, args: Args) -> Result<()> {
+async fn execute(args: Args) -> Result<Outcome> {
     // Check if file already exists
     if args.output.exists() {
         bail!("File already exists: {}", args.output.display());
@@ -62,30 +63,45 @@ async fn execute(ctx: &mut StdCtx, args: Args) -> Result<()> {
     std::fs::write(&args.output, json)
         .with_context(|| format!("failed to write file: {}", args.output.display()))?;
 
-    if args.interactive {
-        shell_println!(ctx, "Created app config: {}", args.output.display())?;
-        shell_println!(ctx)?;
-        shell_println!(
-            ctx,
-            "Run: tvc app create --config-file {}",
-            args.output.display()
-        )?;
-    } else {
-        shell_println!(
-            ctx,
-            "Created app config template: {}",
-            args.output.display()
-        )?;
-        shell_println!(ctx)?;
-        shell_println!(ctx, "Edit the file to fill in your values, then run:")?;
-        shell_println!(
-            ctx,
-            "  tvc app create --config-file {}",
-            args.output.display()
-        )?;
+    Ok(Outcome::AppInit(AppConfigCreated {
+        command: "app init",
+        path: args.output.display().to_string(),
+        template: !args.interactive,
+        interactive: args.interactive,
+    }))
+}
+
+#[derive(Default, Serialize)]
+pub struct AppConfigCreated {
+    command: &'static str,
+    path: String,
+    template: bool,
+    interactive: bool,
+}
+
+impl Message for AppConfigCreated {
+    fn reason(&self) -> &'static str {
+        "app-config-created"
     }
 
-    Ok(())
+    fn human_message(&self) -> String {
+        if self.interactive {
+            format!(
+                r#"Created app config: {}
+
+Run: tvc app create --config-file {}"#,
+                self.path, self.path
+            )
+        } else {
+            format!(
+                r#"Created app config template: {}
+
+Edit the file to fill in your values, then run:
+  tvc app create --config-file {}"#,
+                self.path, self.path
+            )
+        }
+    }
 }
 
 /// Load the operator public key from the active org's config
--- a/tvc/src/commands/app/init.rs
+++ b/tvc/src/commands/app/init.rs
@@ -3,11 +3,12 @@
 use crate::config::app::AppConfig;
 use crate::config::turnkey::{Config, StoredQosOperatorKey};
 use crate::outcome::Outcome;
-use crate::output::{Message, StdCtx};
+use crate::output::StdCtx;
 use crate::prompts::{bail_interactive_conflicts_with_non_interactive, ensure_stdin_is_tty};
 use anyhow::{Context, Result, bail};
 use clap::Args as ClapArgs;
 use serde::Serialize;
+use std::fmt::{self, Display, Formatter};
 use std::path::PathBuf;
 
 /// Generate a template app configuration file.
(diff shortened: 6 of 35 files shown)


tvc/src/commands/app/create.rs
```diff
@@ -150,27 +154,52 @@ async fn run_with_config(ctx: &mut StdCtx, args: Args, app_config: AppConfig) ->
...
-    shell_println!(ctx)?;
-    shell_println!(
-        ctx,
-        "Use one of the Manifest Set Operator IDs above with `tvc deploy approve --operator-id`"
-    )?;
 
-    Ok(())
+    fn human_message(&self) -> String {
+        let mut message = format!(
+            "\nApp created successfully!\n\nApp ID: {}\nName: {}\nManifest Set ID: {}",
```
richardpringle: I think it's a lot easier to read these messages using raw string literals

tvc/src/commands/app/delete.rs
```diff
@@ -18,7 +20,7 @@ pub struct Args {
 }
 
 /// Run the app delete command.
-pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
+pub async fn run(_ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
```
richardpringle, line 23: We should add a trait enforce the pattern here. It's not going to improve the existing code, but it'll enforce the pattern. 

I'll think about it and put it in a ticket. We can iterate a little bit on the design in the ticket itself before anyone goes to implement anything. 

I also think it'll be useful to prevent breaking changes. Changing an output message is technically breaking for a cli. Post GA we'll have to be careful, especially when our docs say to install via `cargo install tvc`. For CI, we should actually have a whole deprecation policy etc. I'm getting ahead of myself here in this stream of consciousness, but I'm convincing myself more an more that all this structure will eventually be super beneficial

tvc/src/commands/app/delete.rs
```diff
@@ -36,11 +38,44 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
...
+}
+
+#[derive(Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct AppDeleted {
+    app_id: String,
+    activity_id: String,
+    /// Stable proto status name, e.g. `ACTIVITY_STATUS_COMPLETED`.
+    #[schemars(with = "String")]
+    activity_status: ActivityStatus,
```
richardpringle: 🤔 what are we doing with json-schema here?

tvc/src/commands/app/delete.rs
```diff
@@ -36,11 +38,44 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
...
+
+impl Message for AppDeleted {
+    fn reason(&self) -> &'static str {
+        "app-deleted"
+    }
 
-    Ok(())
+    fn human_message(&self) -> String {
+        format!(
+            "App delete accepted.\nApp and deployments marked for deletion.\nApp ID: {}\nActivity ID: {}\nActivity Status: {:?}",
```
richardpringle: 🤔 I don't think that a `Debug` implementation should ever end up in user-facing logs. We should have a Display implementation for this

tvc/src/commands/app/list.rs
```diff
@@ -36,16 +38,9 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> anyhow::Result<()> {
...
-        return Ok(());
-    }
-
-    for app in &apps {
-        render_app(ctx, app)?;
-    }
-
-    Ok(())
+    Ok(Outcome::AppList(AppsListed {
+        apps: apps.iter().map(AppSummary::from).collect(),
```
richardpringle: `apps` isn't after this, it can be consumed

```suggestion
        apps: apps.into_iter().map(AppSummary::from).collect(),
```

tvc/src/commands/app/list.rs
```diff
@@ -54,24 +49,72 @@ fn filter_by_name(apps: &mut Vec<TvcApp>, name: Option<&str>) {
...
+        Self {
+            id: app.id.clone(),
+            name: app.name.clone(),
+            quorum_public_key: app.quorum_public_key.clone(),
+            live_deployment_id: app.live_deployment_id.clone(),
+            egress_enabled: app.enable_egress,
+            public_domain: (!app.public_domain.is_empty()).then(|| app.public_domain.clone()),
+        }
+    }
+}
```
richardpringle: I would do something more like this (checkout the inline comments, they aren't meant to be committed):

```suggestion
#[derive(Serialize, JsonSchema)]
#[serde(rename_all = "camelCase")]
struct AppSummary {
    id: String,
    name: String,
    quorum_public_key: String,
    live_deployment_id: Option<String>,
    egress_enabled: bool,
    // could do something with schemars about the string not being empty
    #[serde(skip_serializing_if = "String::is_empty")]
    public_domain: String,
}

impl From<TvcApp> for AppSummary {
    fn from(app: TvcApp) -> Self {
        let TvcApp {
            id,
            name,
            live_deployment_id,
            public_domain,
            quorum_public_key,
            // can alternatively not specify all of these with `..`
            // but doing it this way forces us to think about 
            // whether we want to include any new fields or not
            organization_id: _,
            enable_egress: egress_enabled,
            manifest_set: _,
            share_set: _,
            created_at: _,
            updated_at: _,
            enable_debug_mode_deployments: _,
        } = app;

        Self {
            id,
            name,
            quorum_public_key,
            live_deployment_id,
            egress_enabled,
            public_domain,
        }
    }
}
```

tvc/src/commands/app/list.rs
```diff
@@ -54,24 +49,72 @@ fn filter_by_name(apps: &mut Vec<TvcApp>, name: Option<&str>) {
...
         format!("Quorum Public Key: {}", app.quorum_public_key),
         format!("Live Deployment: {live}"),
-        format_egress_enabled(app.enable_egress),
+        format_egress_enabled(app.egress_enabled),
     ];
 
-    if !app.public_domain.is_empty() {
-        lines.push(format!("Public Domain: {}", app.public_domain));
+    if let Some(domain) = &app.public_domain {
+        lines.push(format!("Public Domain: {domain}"));
```
richardpringle: With my comment above, I would do
```suggestion
    if !app.public_domain.is_emtpy() {
        lines.push(format!("Public Domain: {}", app.public_domain));
```

tvc/src/commands/app/list.rs
```diff
@@ -54,24 +49,72 @@ fn filter_by_name(apps: &mut Vec<TvcApp>, name: Option<&str>) {
...
+
+        self.apps
+            .iter()
+            .map(render_app)
+            .collect::<Vec<_>>()
+            .join("\n")
+    }
+}
+
+fn render_app(app: &AppSummary) -> String {
```
richardpringle: I would just implement `Display` here and maybe rename the type to `AppDisplay` so it's clear that's the only purpose of this type.

tvc/src/commands/app/init.rs
```diff
@@ -62,30 +64,40 @@ async fn execute(ctx: &mut StdCtx, args: Args) -> Result<()> {
...
+    interactive: bool,
+}
+
+impl Message for AppConfigCreated {
+    fn reason(&self) -> &'static str {
+        "app-config-created"
     }
 
-    Ok(())
+    fn human_message(&self) -> String {
```
richardpringle: Just thinking out loud here... what if this method took `self` by value and output `std::fmt::Arguments<'static>`? 

100% not necessary, but it means you could still call `human_message` and have what is effectively lazy evaluation. It also kind of signals to someone writing this that the function may or may not get executed (for example if someone passed in a `-q` for "quiet" aka no logs)

tvc/src/outcome.rs
```diff
@@ -0,0 +1,199 @@
...
+        let mut reasons: Vec<&str> = all_terminal_shapes().iter().map(Message::reason).collect();
+        reasons.extend(NON_TERMINAL_REASONS);
+
+        let unique: HashSet<&str> = reasons.iter().copied().collect();
+        assert_eq!(
+            unique.len(),
+            reasons.len(),
+            "duplicate reason strings in: {reasons:?}"
+        );
+    }
```
richardpringle, line 190: 🤔 

I'm not saying that we delete this, but this test is kind of flawed. Let's say I add an outcome and give it a duplicate reason, this test won't catch that. 

A better way to achieve this would be with a proc-macro. But at the end of the day, I'm not sure it's really even valuable to test. At best you can prevent regression.

tvc/src/outcome.rs
```diff
@@ -0,0 +1,199 @@
...
+        for outcome in all_terminal_shapes() {
+            let reason = outcome.reason();
+            assert!(
+                reason
+                    .chars()
+                    .all(|c| c.is_ascii_lowercase() || c.is_ascii_digit() || c == '-'),
+                "reason `{reason}` is not kebab-case"
+            );
+        }
+    }
```
richardpringle, line 203: This is something else that would be better enforced programatically with a proc-macro at compile time instead of with a unit test. 

You can't prevent someone from _not_ using the macro, but programatically generating these things gives us a much higher certainty that than a test that can't guarantee that it's actually testing all cases

tvc/src/output.rs
```diff
@@ -93,9 +94,20 @@ impl<W, W2> Shell<W, W2> {
...
+    /// mode prints nothing (JSON mode still emits the message).
     pub fn emit<M: Message>(&mut self, message: &M) -> Result<()> {
         match self.message_format {
-            MessageFormat::Human => self.human().line(message.human_message()),
+            MessageFormat::Human => {
+                let text = message.human_message();
+                if text.is_empty() {
+                    return Ok(());
+                }
+                self.human().line(text)
```
richardpringle, line 108: 🤔  again, just thinking out lout, if `human_message()` returned `fmt::Arguments<'static>`, you could do this:

```suggestion
                let Some(text) = message.human_message().as_str() else  {
                    return Ok(());
                }
                
                self.human().lint(text)
```

The difference is pretty subtle and it's not about line count or anything, it's more about being forced to handle both of these paths instead of relying on an extra check

tvc/src/output.rs
```diff
@@ -278,7 +290,7 @@ impl Display for MissingRequiredInput {
 
 impl Error for MissingRequiredInput {}
 
-#[derive(Serialize)]
+#[derive(Serialize, JsonSchema)]
```
richardpringle: More a note for myself as I assume you're going to answer this somewhere else (or already have), but I'm still not quite sure why we need to output the schema for these messages or how one would use it

tvc/src/output.rs
```diff
@@ -394,6 +406,44 @@ mod tests {
...
+    #[test]
+    fn shell_emit_json_still_emits_message_with_empty_human_message() {
+        let mut shell = TestShell::with_json_formatter();
+
+        shell.emit(&MachineOnlyMessage { value: "ok" }).unwrap();
+
+        let output = String::from_utf8(shell.into_stdout()).unwrap();
+        assert_eq!(
+            output,
+            "{\"reason\":\"machine-only-message\",\"value\":\"ok\"}\n"
```
richardpringle: prefer raw string literals over escaped quotation marks

tvc/tests/snapshots/schema.json
richardpringle, line 1: Interesting that I'm seeing this in the `tests` directory

tvc/JSON_OUTPUT_MODE_PLAN_0.md
richardpringle, line 1: remember to delete

tvc/JSON_OUTPUT_MODE_PLAN_1.md
richardpringle, line 1: remember to delete

tvc/OUTCOME_PLAN.md
richardpringle, line 1: remember to delete

tvc/OUTCOME_PLAN.md
```diff
@@ -0,0 +1,311 @@
...
+  }
+  ```
+
+- Scoped form: `tvc schema deploy approve`.
+- `formatVersion` for drift detection, à la `cargo metadata --format-version`;
+  prior art: `terraform providers schema -json`.
+- Cost: `schemars` dependency + one derive per message struct. Lands after
+  Phases 1–6: it derives from the structs created in Phases 1–5, and Phase 6's
+  reason-registry test should be green first so the schema is generated from a
+  verified contract.
```
richardpringle: I think this is overkill. I would prefer that we instead had a `tvc mcp serve` that turned the cli into an mcp. Instead of burning tokens on the shape of the responses and the reponses (json is supposed to be self describing, I'm not quire sure how json schema for outputs really helps).

I think we should split this part out and discuss before committing to it.

tvc/QUESTIONS.md
richardpringle, line 1: remember to delete

Cargo.lock
```diff
@@ -3251,10 +3251,23 @@ checksum = "82d20c4491bc164fa2f6c5d44565947a52ad80b9505d8e36f8d54c27c739fcd0"
...
+name = "schemars_derive"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "33d020396d1d138dc19f1165df7545479dcd58d93810dc5d646a16e55abefa80"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "serde_derive_internals",
+ "syn 2.0.100",
+]
```
richardpringle: new dep
(reminder for myself or another dep reviewer)

Cargo.lock
```diff
@@ -3368,6 +3381,17 @@ dependencies = [
...
+[[package]]
+name = "serde_derive_internals"
+version = "0.29.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "18d26a20a969b9e3fdf2fc2d9f21eda6c40e2de84c9408bb5d3b05d499aae711"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.100",
+]
```
richardpringle: new dep

tvc/src/cli.rs
```diff
@@ -85,7 +86,14 @@ impl Cli {
         let mut ctx = Ctx::new(shell, args.non_interactive);
         let result = args.command.run(&mut ctx).await;
         match result {
-            Ok(()) => ExitCode::SUCCESS,
+            Ok(outcome) => {
+                if let Err(emit_error) = ctx.shell().emit(&outcome) {
+                    let mut stderr = std::io::stderr();
+                    let _ = writeln!(stderr, "error: failed to write CLI output: {emit_error}");
+                    return ExitCode::FAILURE;
+                }
```
richardpringle, line 96: I'm leaning towards this being the wrong way to handle log failures. I think that we should fail-open and make a best effort here, because the actual operation has successfully completed in this case. 

if you running `tvc <some subcommand> | jq` and `jq` fails to parse an earlier message for whatever reason, trying to write to stdout will fail. In this case, the overall command will exit with a non-zero exit code, but the tvc command should still have exited successfully (with an attempt at warning on stderr). 

In other words, we should fail open if we can't deliver messages, or at least this outcome message

tvc/src/commands/schema.rs
```diff
@@ -0,0 +1,316 @@
...
+    let scope = args.command.join(" ");
+    let mut messages = message_catalog();
+
+    if !scope.is_empty() {
+        let matches_scope =
+            |command: &str| command == scope || command.starts_with(&format!("{scope} "));
+
+        if !messages
+            .iter()
+            .any(|message| message.command != "*" && matches_scope(message.command))
```
richardpringle: nit: 
I think 
```suggestion
        if messages
            .iter()
            .find(|message| matches_scope(message.command))
            .is_none()
```
is a little more obvious that there are no matches 

I actually think the logic that you have here might also be wrong? What if you're scope is `*` which should be valid, right? It's going to error. You should be able to do that I'm pretty sure

tvc/src/commands/schema.rs
```diff
@@ -0,0 +1,316 @@
...
+            "quorum-key-share-posted",
+            "deploy post-share",
+            Terminal,
+        ),
+        entry::<deploy::status::DeploymentStatusReport>(
+            "deployment-status",
+            "deploy status",
+            Terminal,
+        ),
+        entry::<deploy::create::DeploymentCreated>("deployment-created", "deploy create", Terminal),
```
richardpringle: What happens if we use an alias? Or what if we change the command name? Manually entering commands like this seems fragile and it seems like we're kind of re-inventing the wheel for something that clap already handles, it just doesn't do it in a json format like this. It would be better if we expanded `clap::Subcommand` with an internal
```rust
pub(crate) trait Subcommand: clap::Subcommand {
    type Outcome: schemars::JsonSchema;
}
```
or something along these lines. 

So... I know that the nature of the ticket was that it would make the `tvc` cli easier to read/use for agents, but I would much prefer having an mcp-layer and have ideas on how to do that with minimal maintenance, using the clap types/traits.

richardpringle (changes requested):
Not sure what key combination I hit, but I didn't mean to leave the last review with as a comment review (I didn't mean to submit it at all, I had more comments)

tvc/src/outcome.rs
richardpringle: Actually, a better way to make all the reasons unique would be to split the `Message` trait up into `Reason` and `Message`. Then only implement `reason` on `Outcome`. You could still do something like `trait Reason: Message` if it doesn't make sense for something to have a reason, but no message. In any case, the trait components don't quite constrain this the way we want (I don't think). The inner outcome type should never be emitted on its own, right?

Oh! I think the proper thing to do is to implement `Display` for all these outcome types, then do this:
```rust
impl Message for Outcome {
    fn reason(&self) -> 'static str {
        match self {
            Self::Login(_) => "login"
        }
    }

    fn human_message(&self) -> 'static str {
        match self {
            Self::Login(outcome) => format!("{outcome}")
        }
    }
}

impl Serialize for Outcome {
    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
        match self {
            Self::Login(outcome) => outcome.serialize(serializer),
        }
    }
}
```

This way, only the top level outcome can only be emitted. The reason is easy to keep unique because it's implemented on the top level outcome. The message is still defined in the subcommand's module. 

This seems like it puts the proper constraints on the problem

tvc/src/commands/app_status.rs
```diff
@@ -13,14 +15,45 @@ pub fn sanitize_app_status(mut app_status: AppStatus) -> AppStatus {
...
+}
+
+impl From<&Timestamp> for TimestampPayload {
+    fn from(timestamp: &Timestamp) -> Self {
+        Self {
+            seconds: timestamp.seconds.clone(),
+            nanos: timestamp.nanos.clone(),
+        }
+    }
+}
```
richardpringle: This an anti pattern. Clone outside functions as much as possible.

```suggestion
impl From<Timestamp> for TimestampPayload {
    fn from(timestamp: Timestamp) -> Self {
        Self {
            seconds: timestamp.seconds,
            nanos: timestamp.nanos,
        }
    }
}
```
Then you call `timestamp.clone().into()` or `timstamps.iter().cloned().map(Into::into)`

tvc/src/commands/app_status.rs
```diff
@@ -13,14 +15,45 @@ pub fn sanitize_app_status(mut app_status: AppStatus) -> AppStatus {
...
+    format_replica_counts(
+        deployment_status.ready_replicas,
+        deployment_status.desired_replicas,
     )
 }
 
+/// Render ready/desired replica counts with the shared label.
+pub fn format_replica_counts(ready: i32, desired: i32) -> String {
+    format!("Healthy / Desired Replicas: {ready}/{desired}")
+}
```
richardpringle, line 45: Why have both these functions. It's the same logic with two different calling patterns

richardpringle (commented):
To be honest, I didn't actually finish the review here, there are enough comments already and it's annoying that I didn't put them all in the same review. 

Leaving this as a comment review to stay consistent.

richardpringle:
I kind of dove into this without looking at size of the diff. I would love if this could be split up into "every run command produces an `Outcome::<variant>`" and then we can talk about the schema stuff. 

I think my suggestion about implementing `Message` for the top-level outcome would remove the necessary tests for uniqueness and some of those things that would be better solved through proc macros. 

I can't remember if it was on slack or if I mentioned it in here, but it seems like it would be a good idea to have a `Run` trait that forces implementing new commands with the `context` as the first argument pattern. I **don't** think we should do that yet, just do the `Outcome` stuff.

daniilrrr:
this PR is way too big, that is my bad. Separately I am going to remove the Schema usage from this PR as it seems poorly fitted to the problem and might be better served by an `mcp serve`

tvc/src/commands/app/delete.rs
daniilrrr: killed `schemars` throughout PR. will resolve comments referencing schema

tvc/tests/snapshots/schema.json
daniilrrr: killed `schemars` throughout PR. will resolve comments referencing schema

tvc/OUTCOME_PLAN.md
daniilrrr: killed `schemars` throughout PR. will resolve comments referencing schema

Cargo.lock
daniilrrr: removed

tvc/src/commands/schema.rs
daniilrrr: file removed

tvc/src/commands/app/create.rs
daniilrrr: updated to string literals everywhere for non-trivial / multi-line outputs

tvc/src/commands/app/list.rs
daniilrrr: done

tvc/src/commands/app/delete.rs
daniilrrr: great catch, i have removed it in 4 places. 

can't have a Display impl here because `ActivityStatus` is foreign and coming from generated `turnkey_client`, so i'll just name it explicitly

daniilrrr: are you ok with debug for std Collections? currently in [approve.rs](https://github.com/tkhq/rust-sdk/blob/ba9746022072e521721caced089d62785f7c32b2/tvc/src/commands/deploy/approve.rs#L207) and [status.rs](https://github.com/tkhq/rust-sdk/blob/ba9746022072e521721caced089d62785f7c32b2/tvc/src/commands/deploy/status.rs#L123)

tvc/src/commands/app_status.rs
daniilrrr: fixed throughout

tvc/src/commands/app/delete.rs
daniilrrr: agree, punting on new `Trait` to a future ticket

tvc/src/cli.rs
daniilrrr: agreed

tvc/src/commands/app/list.rs
daniilrrr: nice, I read this the other day too https://antoine.vandecreme.net/blog/exhaustive-destructuring-praise/

daniilrrr: done here and elsewhere where relevant and funcs are working on individual types

tvc/src/commands/app/list.rs
daniilrrr: done

tvc/src/commands/app/init.rs
daniilrrr: gonna punt, could do in follow-up https://linear.app/turnkey/issue/TVC-207/enforce-outcome-reason-uniqueness-at-compile-time-via-a-proc-macro if you feel strongly

tvc/src/outcome.rs
daniilrrr: done

tvc/src/outcome.rs
daniilrrr: punting to https://linear.app/turnkey/issue/TVC-207/enforce-outcome-reason-uniqueness-at-compile-time-via-a-proc-macro

tvc/src/output.rs
daniilrrr: punting on `Arguments<'static>`

daniilrrr:
hey @richardpringle, thank you for the comprehensive partial review. I truly appreciate the feedback

- I've addressed all of the style and patterns comments you raised, and folded the relevant and repeated ones into `tvc/AGENTS.md` to avoid agents retreading similar ground in the future
- I have done the `Message` refactor you suggested, and now `human_message()` is implemented by each Payload's `Display impl`. I think this is great and I can go one level further -> see [comment](https://github.com/tkhq/rust-sdk/pull/188/changes#r3626460543)
- Rebased on latest main 
- We've discussed doing a `Run trait` in a follow-up PR, however I think the command forms are currently constrained sufficiently by the `match` statement [here](https://github.com/tkhq/rust-sdk/blob/74b9fa253c79436abca4d0589f72c7e791fc2683/tvc/src/cli.rs#L87-L88) so it's ok to leave as-is
- I'm experimenting with the Reviews tool and created a packet for you here https://reviews.figitaki.dev/r/lYghULlT to help navigate the large PR, in case it is useful 

Planning to follow up with [TVC-207](https://linear.app/turnkey/issue/TVC-207/enforce-outcome-reason-uniqueness-at-compile-time-via-a-proc-macro) to implement the proc macro for`Outcome` enforcement

tvc/src/output.rs
daniilrrr, line 30: @richardpringle 

I could go one level further in the refactor and kill `human_message()` here by having `Display` as a supertrait and use the mandatory `.to_string()` instead of `.human_message()`. Personally I like the explicitness of `human_message()` but it doesn't add anything at this moment. And `.to_json_string()` is a good counterpart to `.to_string()` from `Display`. What do you think? 

___
LLM output below:
```rust
Here's the shape of it. The idea: make Display the one human-render trait, and shrink Message to just "reason + JSON," requiring Display as a supertrait.

The trait

// now
pub trait Message: Serialize {
    fn reason(&self) -> &'static str;
    fn human_message(&self) -> String;
    fn to_json_string(&self) -> String { /* default: serialize self + flatten reason */ }
}

// supertrait version
pub trait Message: Serialize + Display {
    fn reason(&self) -> &'static str;
    fn to_json_string(&self) -> String { /* unchanged */ }
}

...

// now
MessageFormat::Human => {
    let text = message.human_message();
    if text.is_empty() { return Ok(()); }   // machine-only convention preserved
    self.human().line(text)
}

// after
MessageFormat::Human => {
    let text = message.to_string();          // Display
    if text.is_empty() { return Ok(()); }
    self.human().line(text)
}

...

Outcome — instead of a human_message that delegates to the payload's Display, it just is Display (delegating), and its Message impl keeps only reason:

// after
impl Display for Outcome {
        with_message!(self, |msg| write!(f, "{msg}"))   // payload's Display
    }
}
impl Message for Outcome {
    fn reason(&self) -> &'static str { match self { … } }   // unchanged
}

That removes the current Outcome::human_message → to_string() indirection hop you noticed.
```

tvc/src/outcome.rs
daniilrrr: see https://github.com/tkhq/rust-sdk/pull/188/changes#r3626460543 for one follow up

tvc/src/commands/app/list.rs
```diff
@@ -54,24 +49,97 @@ fn filter_by_name(apps: &mut Vec<TvcApp>, name: Option<&str>) {
...
+            live_deployment_id,
+            public_domain,
+            enable_egress: egress_enabled,
+            enable_debug_mode_deployments: debug_mode_deployments_enabled,
+            organization_id: _,
+            manifest_set: _,
+            share_set: _,
+            created_at: _,
+            updated_at: _,
+        } = app;
```
richardpringle, line 89: 🥇 

It feels dumb now... but just wait until you add a field, get a compiler error, and realize you want to print that field too. 

I started doing this after not seeing some fields in prometheus metrics and flying blind while  waited for the fix to propagate to prod

tvc/src/commands/deploy/approve.rs
```diff
@@ -105,6 +107,135 @@ struct PostApprovalPlan<'a> {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.write_str(&approval_payload_human_message(
+            &self.approval,
+            &self.written_to,
+        ))
+    }
+}
+
+#[derive(Default, Serialize)]
+pub struct ApprovalDryRun {}
```
richardpringle, line 231: I generally prefer 
```suggestion
pub struct ApprovalDryRun;
```
It's what they use in The Book: https://doc.rust-lang.org/book/ch05-01-defining-structs.html#defining-unit-like-structs

tvc/src/commands/deploy/approve.rs
```diff
@@ -105,6 +107,135 @@ struct PostApprovalPlan<'a> {
...
+
+#[derive(Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ApprovalPosted {
+    /// Present when the approval was printed inline (no `--approval-out`).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    approval: Option<Approval>,
+    /// Present when the approval was written to a file via `--approval-out`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    written_to: Option<String>,
```
richardpringle, line 171:
```suggestion
    #[serde(flatten)]
    approval_or_path: ApprovalOrPath
```
with 
```rust
#[derive(Serialize)]
enum ApprovalOrPath {
    WrittenTo(String),
    Approval(Approval)
}
```

You might need to wrap that in an `Option` if it's possible for both to be `None`.

If you do this, it will be impossible for the invalid scenario to exist.

tvc/src/commands/deploy/approve.rs
```diff
@@ -105,6 +107,135 @@ struct PostApprovalPlan<'a> {
...
+#[derive(Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ApprovalGenerated {
+    /// Present when the approval was printed inline (no `--approval-out`).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    approval: Option<Approval>,
+    /// Present when the approval was written to a file via `--approval-out`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    written_to: Option<String>,
+}
```
richardpringle, line 219: Same thing here, you can do something like:
```rust
pub struct ApprovalGenerated(Option<ApprovalOrPath>);
```

tvc/src/commands/deploy/approve.rs
```diff
@@ -105,6 +107,135 @@ struct PostApprovalPlan<'a> {
...
+    approval: &Option<Approval>,
+    written_to: &Option<String>,
+) -> String {
+    match (approval, written_to) {
+        (Some(approval), _) => serde_json::to_string_pretty(approval)
+            .expect("serializing manifest approval should not fail"),
+        (None, Some(path)) => format!("Approval written to: {path}"),
+        (None, None) => String::new(),
+    }
+}
```
richardpringle, line 161: With my comments below, you won't need this free function

tvc/src/commands/deploy/approve.rs
```diff
@@ -246,30 +378,53 @@ async fn build_inputs_non_interactive(
...
+        .approval_out
+        .as_ref()
+        .map(|path| path.display().to_string());
+    // The approval payload rides in the outcome only when it was not written
+    // to a file; otherwise the outcome carries the file path.
+    let inline_approval = if written_to.is_none() {
+        Some(approval.clone())
+    } else {
+        None
+    };
```
richardpringle, line 403: Instead of doing this here, you can make a constructor for `ApprovalOrPath` (see comments above) that you can call in both branches so you don't have to clone the approval

tvc/src/commands/deploy/approve.rs
```diff
@@ -300,36 +457,11 @@ async fn sign_and_output(
...
         let json = serde_json::to_string_pretty(&approval)?;
         write_file(path, &json).await?;
-        shell_println!(ctx, "Approval written to: {}", path.display())?;
-    } else {
-        // Emit the approval as the machine-relevant payload. In human mode this
-        // prints the pretty JSON exactly as before; in JSON mode it is wrapped
-        // in a stable `reason` envelope so the approval survives suppression.
-        ctx.shell().emit(&ManifestApprovalGenerated {
-            approval: &approval,
-        })?;
```
richardpringle, line 310: 👍 

Nice... this definitely doesn't belong here!

tvc/src/commands/deploy/approve.rs
```diff
@@ -378,47 +510,37 @@ async fn post_approval_to_api(
         .await
         .context("failed to post manifest approval")?;
 
-    shell_println!(ctx)?;
-    shell_println!(ctx, "Approval posted successfully!")?;
-    shell_println!(ctx)?;
-    shell_println!(ctx, "Approval IDs: {:?}", result.result.approval_ids)?;
-    shell_println!(ctx, "Manifest ID: {}", plan.manifest_id)?;
-    shell_println!(ctx, "Operator ID: {}", plan.operator_id)?;
```
richardpringle, line 386: Yeah, this great for maintenance and extension 🥇

tvc/src/commands/keys/init_local_quorum_key.rs
```diff
@@ -37,25 +39,33 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
...
 
-    Ok(())
+impl Display for QuorumKeyConfigCreated {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        // Constraints inherited from qos_crypto::shamir::shares_generate.
+        write!(
+            f,
+            r#"Created quorum key config template: {}
+
+Constraints (see qos_crypto/src/shamir.rs):
```
richardpringle, line 60: Not for this PR, but is this already a dependency? We might be able to pull these values out and display them here instead of manually doing it this way.

tvc/src/outcome.rs
```diff
@@ -0,0 +1,204 @@
...
+            Outcome::AppList(_) => "apps-listed",
+            Outcome::AppCreate(_) => "app-created",
+            Outcome::AppInit(_) => "app-config-created",
+            Outcome::AppSetLiveDeploy(_) => "live-deployment-set",
+            Outcome::AppDelete(_) => "app-deleted",
+            Outcome::KeysCreateQuorumKey(_) => "quorum-key-created",
+            Outcome::KeysGenerateQuorumKey(_) => "quorum-key-generated",
+            Outcome::KeysInitQuorumKey(_) => "quorum-key-config-created",
+            Outcome::KeysReEncryptShare(_) => "re-encrypted-share-generated",
+        }
```
richardpringle, line 114: This looks a lot nicer

tvc/src/outcome.rs
```diff
@@ -0,0 +1,204 @@
...
+    };
+    use std::collections::HashSet;
+
+    /// One zero-value instance per terminal shape, in `Outcome`'s declaration
+    /// order (command-local shapes expanded in their own declaration order).
+    /// The registry tests only read reason strings, so `Default` fixtures
+    /// suffice; realistic payloads are exercised by each command's own tests.
+    fn all_terminal_shapes() -> Vec<Outcome> {
+        vec![
+            Outcome::Login(login::LoggedIn::default()),
```
richardpringle, line 138: you're just using these in tests, right? You should add a `#[cfg(test)] impl Default for X { ... }` to all the implementations as I don't think they're meant to be used outside of tests

tvc/src/outcome.rs
```diff
@@ -0,0 +1,204 @@
...
+            Outcome::KeysCreateQuorumKey(keys::create_quorum_key::QuorumKeyCreated::default()),
+            Outcome::KeysGenerateQuorumKey(
+                keys::generate_local_quorum_key::QuorumKeyGenerated::default(),
+            ),
+            Outcome::KeysInitQuorumKey(
+                keys::init_local_quorum_key::QuorumKeyConfigCreated::default(),
+            ),
+            Outcome::KeysReEncryptShare(
+                keys::re_encrypt_local_share::ReEncryptedShareGenerated::default(),
+            ),
```
richardpringle, line 169: I think I already left this comment but we can use `strum` for this: https://docs.rs/strum/latest/strum/derive.EnumIter.html

Just add `strum` as a dev-dependency and use `#[cfg_attr(test, derive(EnumIter))]` then filter the iterator for non-terminal reasons and call `collect` 🤙 

Can do this in a follow up.

tvc/src/output.rs
richardpringle: yeah, this change looks good, can do in a follow up

tvc/src/commands/app/list.rs
daniilrrr: not dumb at all, it's a little boilerplate-y but for versioned types I think it's pretty sick. In fact it surfaced the fact that I wasn't publishing `debug_mode` info in the summaries that I wanted to have. So already it's useful

tvc/CLAUDE.md
richardpringle, line 1: 😂

tvc/AGENTS.md
```diff
@@ -0,0 +1,37 @@
...
+  trailing whitespace, still use a raw literal but add a comment noting the
+  trailing whitespace so editors/formatters don't silently strip it.
+- Prefer moving owned values over cloning them. If you already own a value and
+  this is its last use (e.g. building the return value at the end of a
+  function), move it out — use `.into_iter()` or partial field moves instead of
+  `.clone()`.
+- Don't hide clones inside functions. A function or `From`/`TryFrom` impl should
+  take an owned value (`T`, not `&T`) rather than clone internally; prefer
+  `From<T>` over `From<&T>`. When a clone is genuinely needed, make it explicit
+  at the call site — e.g. `value.clone().into()` or
```
richardpringle, line 20: ❤️

tvc/AGENTS.md
```diff
@@ -0,0 +1,37 @@
...
+  `as _` where `std::io::Write` is also in scope.
+- When converting from an external/generated type (e.g. the API's `TvcApp`,
+  `TvcDeployment`, `AppStatus`) into one of our own structs, destructure it
+  exhaustively — `let Foo { a, b, c: _ } = value;` with no trailing `..` —
+  rather than reading fields with `value.a`. Bind the fields you use and
+  `_`-bind the ones you don't. This way, when the upstream type gains a field, the destructure
+  fails to compile and forces a deliberate decision about whether the new field
+  belongs in our output — instead of it being silently dropped. Skip this only
+  where it adds noise for no value, e.g. reading one or two fields off a large
+  API response result.
```
richardpringle, line 37: 🚀

tvc/AGENTS.md
richardpringle, line 1: I have a few things to add but this is a fantastic start

richardpringle (approved):
Check out my comments. I think a few more (relatively small) changes can be made in a follow up. 

This really cleans up the code though, I love it

tvc/AGENTS.md
daniilrrr: that's the idea! we can def add as we go

tvc/src/outcome.rs
daniilrrr: im planning to kill this test and use the proc-macro instead in follow up so I didn't pay much mind to it

tvc/src/commands/deploy/approve.rs
daniilrrr: done in #227

tvc/src/commands/deploy/approve.rs
daniilrrr: already fixed

tvc/src/commands/deploy/approve.rs
daniilrrr: already fixed

tvc/src/outcome.rs
daniilrrr: you already did this in your PR, nice

tvc/src/commands/keys/init_local_quorum_key.rs
daniilrrr: `qos_crypto` doesn't export the values so we cant directly ref them but I will update the comments in #227

tvc/src/commands/deploy/approve.rs
daniilrrr: looks like you're covering this in #208  (thanks!) so i wont fix

tvc/src/commands/deploy/approve.rs
daniilrrr: you're covering this in #208

PR #189, opened 2026-07-13
feat(tvc): validate app-id at deploy create

richardpringle:
## Summary

Validate `app-id` at `tvc deploy create` so a leftover placeholder or a wrong/nonexistent app UUID fails fast instead of creating a deployment that can't be approved.

## What it does

- Rejects a malformed `app-id` (not a well-formed UUID) before sending, in `DeployConfig::validate()`. Accepts any UUID version since Turnkey IDs mix v4 and v7.
- Verifies the app exists via `get_tvc_app` before creating the deployment; errors with `app not found: <id>`.

The `<FILL_IN_APP_ID>` placeholder was already rejected by `validate()`. Server-side validation lives in `tkhq/mono` and is out of scope here.

Closes [ENG-4459](https://linear.app/turnkey/issue/ENG-4459/validate-app-id-at-tvc-deploy-create-reject-placeholder-nonexistent)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ qos_p256 = { version = "0.12.1", default-features = false }
 base64 = { version = "0.22.0", default-features = false, features = ["std"] }
 bs58 = { version = "0.5.0", features = ["std", "check"], default-features = false }
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
+uuid = { version = "1.13.1", default-features = false }
 serde = { version = "1.0.219", default-features = false, features = ["std", "derive"] }
 serde_json = { version = "1.0.140", default-features = false, features = ["std"] }
 serde_bytes = { version = "0.11", default-features = false }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -27,6 +27,7 @@ qos_p256 = { version = "0.12.1", default-features = false }
 base64 = { version = "0.22.0", default-features = false, features = ["std"] }
 bs58 = { version = "0.5.0", features = ["std", "check"], default-features = false }
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
+uuid = { version = "1", default-features = false }
```
richardpringle: Need exact version like the other dependencies

Cargo.toml
```diff
@@ -27,6 +27,7 @@ qos_p256 = { version = "0.12.1", default-features = false }
 base64 = { version = "0.22.0", default-features = false, features = ["std"] }
 bs58 = { version = "0.5.0", features = ["std", "check"], default-features = false }
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
+uuid = { version = "1.13.1", default-features = false }
```
richardpringle, line 30: https://github.com/tkhq/qos/blob/main/Cargo.lock#L3365

emostov (approved):

PR #190, opened 2026-07-15
Remove generics that aren't necessary

richardpringle:
I should have caught this on my initial review. I started by wanting to see how much of a pain my suggestion would be, then I ended up just taking on the pain myself. I made a couple other small changes like ~~adding the lint rule to Cargo.toml (still needs to be removed from main.rs and lib.rs) and~~ instead of fallibly converting byte arrays to strings in tests, I infallibly convert strings to byte arrays. I also messed around with some type aliases. 

My changes should be pretty straighforward.

### DIFF
--- a/tvc/src/cli.rs
+++ b/tvc/src/cli.rs
@@ -1,7 +1,7 @@
 //! CLI parsing and dispatch.
 
 use crate::commands;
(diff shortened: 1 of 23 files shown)


daniilrrr (approved):
lgtm, appreciate you doing this in your own PR

PR #191, opened 2026-07-15
tvc: surface consensus-needed clearly (non-zero exit, no duplicate activities)

emostov:
Part 2 of 3 for TVC-183 (consensus-needed handling in the tvc CLI).

## Problem

In orgs requiring root-quorum consensus, `tvc deploy create` / `tvc deploy approve` submit an activity and the server returns HTTP 200 with `ACTIVITY_STATUS_CONSENSUS_NEEDED` — a normal response, not an error. But the generated client maps ConsensusNeeded to `Err(ActivityRequiresApproval(id))` and the CLI wraps that as a fatal generic error (`failed to create TVC deployment` / `failed to post manifest approval`). Operators read this as a failure and re-run; each run stamps a fresh `timestampMs`, the server fingerprint is `sha256(raw body)`, so every retry creates a **duplicate activity**.

### `turnkey_client`
- New error variant `TurnkeyClientError::ActivityPendingConsensus { activity_id, fingerprint }` returned for `ACTIVITY_STATUS_CONSENSUS_NEEDED`. It carries the activity **fingerprint**, which is what quorum members need to vote (`ApproveActivityIntent { fingerprint }`).
- `ActivityRequiresApproval(activity_id)` is kept as-is and still returned for `ACTIVITY_STATUS_AUTHENTICATORS_NEEDED`, so existing consumers matching that variant keep compiling.
- **Semver note:** adding a variant to the (non-`#[non_exhaustive]`) `TurnkeyClientError` enum is technically breaking for exhaustive matches without a wildcard arm; in practice consumers match specific variants with a catch-all. Consumers relying on `ActivityRequiresApproval` for consensus flows must migrate to the new variant.

### `tvc` CLI
- `deploy create` and `deploy approve` no longer print a generic failure for consensus-needed. They print a clear pending-consensus message: the activity was created successfully, its **activity ID** and **fingerprint**, and the command another quorum member should run (`tvc activity approve --fingerprint <fp>` — subcommand ships in the stacked follow-up PR).
- **Exit-code semantics:** the CLI exits with dedicated code **2** (`tvc::exit::CONSENSUS_NEEDED_EXIT_CODE`) — non-zero because the operation is not complete, but distinct from `1` (generic error) so scripts/CI can tell "pending consensus" apart from a real failure.

## Test coverage
- `client`: unit test asserting ConsensusNeeded maps to `ActivityPendingConsensus` with both id and fingerprint (authenticators-needed test unchanged).
- `tvc/tests/deploy_consensus.rs` (new, wiremock-based): `deploy create` and `deploy approve` against a mock API returning consensus-needed — asserts exit code 2, pending-consensus message with activity id + fingerprint + approval instruction, and that the old generic failure text is absent.

## Verification
- `cargo fmt -- --check` ✅
- `cargo clippy --all-targets --all-features -- -D warnings` ✅
- `cargo test -p tvc -p turnkey_client` ✅ (25 client + 127 tvc unit + all integration tests)
- `git diff --check` ✅

### DIFF
--- a/tvc/src/commands/consensus.rs
+++ b/tvc/src/commands/consensus.rs
@@ -1,0 +1,43 @@
+//! Helpers for activity responses that were accepted but still need quorum.
+
+use anyhow::Result;
+use turnkey_client::TurnkeyClientError;
+
+pub(crate) struct PendingConsensusActivity<'a> {
+    pub activity_id: &'a str,
+    pub fingerprint: &'a str,
+}
+
+pub(crate) fn pending_consensus_from_error(
+    error: &TurnkeyClientError,
+) -> Option<PendingConsensusActivity<'_>> {
+    match error {
+        TurnkeyClientError::ActivityPendingConsensus {
+            activity_id,
+            fingerprint,
+        } => Some(PendingConsensusActivity {
+            activity_id,
+            fingerprint,
+        }),
+        _ => None,
+    }
+}
+
+pub(crate) fn print_pending_consensus(activity: &PendingConsensusActivity<'_>) {
+    println!();
+    println!("Consensus needed. The activity was created successfully but is pending quorum.");
+    println!();
+    println!("Activity ID: {}", activity.activity_id);
+    println!("Fingerprint: {}", activity.fingerprint);
+    println!();
+    println!("Next steps:");
+    println!(
+        "  - Ask another quorum member to run `tvc activity approve --fingerprint {}`",
+        activity.fingerprint
+    );
+}
+
+pub(crate) fn pending_consensus_result(activity: &PendingConsensusActivity<'_>) -> Result<()> {
+    print_pending_consensus(activity);
+    Err(crate::exit::ExitError::consensus_needed().into())
+}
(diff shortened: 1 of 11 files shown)


tvc/src/commands/consensus.rs
```diff
@@ -0,0 +1,43 @@
...
+pub(crate) fn print_pending_consensus(activity: &PendingConsensusActivity<'_>) {
+    println!();
+    println!("Consensus needed. The activity was created successfully but is pending quorum.");
+    println!();
+    println!("Activity ID: {}", activity.activity_id);
+    println!("Fingerprint: {}", activity.fingerprint);
+    println!();
+    println!("Next steps:");
+    println!(
+        "  - Ask another quorum member to run `tvc activity approve --fingerprint {}`",
```
emostov, line 35: This isn't necessarily a quorum member, it could be another user on the consensus policy

emostov:
Closing this draft in favor of Linear tracking: [TVC-183](https://linear.app/turnkey/issue/TVC-183/tvc-cli-treats-consensus-needed-as-an-error-causing-duplicate-manifest), under the [TVC-219 CLI consensus handling](https://linear.app/turnkey/issue/TVC-219/cli-consensus-handling) meta issue.

PR #195, opened 2026-07-15
feat(tvc)!: --operator-seed takes the seed value; add --operator-seed-path

richardpringle:
Breaking: `--operator-seed` (env `TVC_OPERATOR_SEED`) is now the hex seed itself, validated at the clap layer by a new `HexSeed` type (optional `0x` prefix, redacting `Debug`, zeroized on drop). The old file-path behavior moves to `--operator-seed-path` (env `TVC_OPERATOR_SEED_PATH`). The flags are mutually exclusive; with neither, the logged-in org operator key is used as before.

Keeps CI seeds off disk without encoding a protocol into the value. Supersedes #173 — for stdin, use `xargs` / `"$(cat ...)"`.


tvc/src/operator_key.rs
```diff
@@ -1,31 +1,93 @@
...
+}
+
+impl OperatorSeedSource {
+    /// Convert the raw `--operator-seed` / `--operator-seed-path` args into a
+    /// seed source. `None` when neither flag was given (the caller falls back
+    /// to the stored operator key); an error when both were given.
+    pub fn from_args(seed: Option<HexSeed>, path: Option<PathBuf>) -> anyhow::Result<Option<Self>> {
+        match (seed, path) {
+            (Some(_), Some(_)) => {
+                bail!("--operator-seed and --operator-seed-path are mutually exclusive")
```
richardpringle:
```suggestion
                bail!("--operator-seed and --operator-seed-path are mutually exclusive, please provide only one")
```

emostov (approved):
🔥

PR #199, opened 2026-07-15
tvc-axum: signed responses with attestation/manifest headers

emostov:
Adds a `tvc-axum` crate: a tower `ResponseSigningLayer` that signs every response with the enclave ephemeral key (and optionally the quorum key) using RFC 9421/9530 headers, and attaches `x-tvc-attestation-doc` (user_data = manifest hash, public_key = ephemeral key) and `x-tvc-manifest-envelope` headers so clients can verify the full trust chain. Includes a `QosJson<T>` response type for canonical `qos_json` bodies.

Ships the client-side verifiers: `verify_quorum_signature` (quorum response signature against a known quorum key) and `verify_response_trust_chain` (qos_core attestation/manifest verification, then the ephemeral response signature with the attested key).

The trust-chain verifier lives in qos_core (tkhq/qos#765); `[patch.crates-io]` points qos_* at that branch until it publishes.

Verification: `cargo fmt -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test -p tvc-axum`, `cargo build --workspace`.


tvc-axum/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+                [(header::CONTENT_TYPE, "application/json")],
+                r#"{"error":"serialization failed"}"#,
+            )
+                .into_response(),
+        }
+    }
+}
+
+/// Request an attestation document binding `user_data` and `public_key`.
+fn request_attestation_doc(
```
emostov, line 167: Example of NSM request

PR #200, opened 2026-07-16
feat(tvc)!: reuse operator by default in app create

richardpringle:
Closes TVC-54.

`tvc app create` reuses the operator from the last `app create` by default instead of minting a new operator ID each run (all backed by the same local key). `--no-operator-reuse` (env `TVC_NO_OPERATOR_REUSE`) forces a new one.

- 1 saved operator → reuse silently
- multiple → prompt (interactive) / error (non-interactive)
- first run, explicit `existingOperatorIds`, or `manifestSetId` → untouched; manifest set only

**Breaking:** non-interactive runs with multiple saved operators now error (set `existingOperatorIds` or pass `--no-operator-reuse`). Overlaps TVC-123's interactive-prompt piece for `app create`.

### DIFF
--- a/tvc/src/commands/app/create.rs
+++ b/tvc/src/commands/app/create.rs
@@ -6,13 +6,17 @@ use crate::{
         app::{AppConfig, AppConfigValidationErrors, OperatorSetParams},
         turnkey::{self, StoredQosOperatorKey},
     },
-    output::StdCtx,
+    output::{Ctx, StdCtx},
     prompts, shell_println,
 };
-use anyhow::{Context, Result, anyhow};
+use anyhow::{Context, Result, anyhow, bail};
 use clap::Args as ClapArgs;
-use std::path::{Path, PathBuf};
 use std::time::{SystemTime, UNIX_EPOCH};
+use std::{
+    io,
+    path::{Path, PathBuf},
+};
+use tracing::debug;
 use turnkey_client::generated::{CreateTvcAppIntent, TvcOperatorParams, TvcOperatorSetParams};
 
 /// Create a new TVC application from a config file.
@@ -23,6 +27,14 @@ pub struct Args {
     /// Path to the app configuration file (JSON).
     #[arg(short = 'c', long, value_name = "PATH", env = "TVC_APP_CONFIG")]
     pub config_file: PathBuf,
+
+    /// Create a new operator instead of reusing the most recently created one.
+    ///
+    /// By default `app create` reuses the operator from your last `app create`
+    /// (the same local operator key) rather than minting a new operator ID each
+    /// time. Pass this to force creating a new operator.
+    #[arg(long, env = "TVC_NO_OPERATOR_REUSE")]
+    pub no_operator_reuse: bool,
 
     #[command(flatten)]
     overrides: Overrides,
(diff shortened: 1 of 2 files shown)


tvc/src/commands/app/create.rs
```diff
@@ -383,4 +499,93 @@ mod tests {
...
+            OperatorReuse::KeepConfig
+        );
+    }
+
+    /// First run (nothing created yet) has nothing to reuse -> create new.
+    #[test]
+    fn decide_reuse_keeps_config_without_saved_ids() {
+        let params = manifest_params_with_new_operator();
+        assert_eq!(
+            decide_operator_reuse(false, Some(&params), &[]),
```
daniilrrr, line 526: i think all of these unit tests testing the same pure func **could**  be a single table-driven test with a descriptive test name to reduce visual clutter. However I can see that some of them have differing setup so I leave that to you

tvc/src/commands/app/create.rs
```diff
@@ -24,6 +28,19 @@ pub struct Args {
...
+    /// Create a new operator instead of reusing the most recently created one.
+    ///
+    /// By default `app create` reuses the operator from your last `app create`
+    /// (the same local operator key) rather than minting a new operator ID each
+    /// time. Pass this to force creating a new operator.
+    #[arg(
+        long,
+        env = "TVC_NO_OPERATOR_REUSE",
+        action = ArgAction::SetTrue,
+        value_parser = BoolishValueParser::new()
```
daniilrrr: are these 2 strictly necessary for a bool flag? internet says `ArgAction::SetTrue` is the default for `bool`s that get triggered. Believe you if so, just looks weird and new to me

tvc/src/commands/app/create.rs
```diff
@@ -46,10 +63,106 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
...
+        &saved_ids,
+    ) {
+        OperatorReuse::KeepConfig => {}
+        OperatorReuse::Reuse(id) => apply_operator_reuse(ctx, &mut app_config, id)?,
+        OperatorReuse::MultipleCandidates(ids) => {
+            if ctx.is_non_interactive() {
+                bail!(
+                    "multiple saved operator IDs for the active org; \
+                     set manifestSetParams.existingOperatorIds in your config to reuse one, \
+                     or pass --no-operator-reuse to create a new operator"
```
daniilrrr, line 80: "pass a flag to not reuse your operator" reads weird to me. Seems like a an unnecessary negative. Could we call if `force_create_operator` instead? Will comment on the bool too

tvc/src/commands/app/create.rs
```diff
@@ -24,6 +28,19 @@ pub struct Args {
...
+    /// By default `app create` reuses the operator from your last `app create`
+    /// (the same local operator key) rather than minting a new operator ID each
+    /// time. Pass this to force creating a new operator.
+    #[arg(
+        long,
+        env = "TVC_NO_OPERATOR_REUSE",
+        action = ArgAction::SetTrue,
+        value_parser = BoolishValueParser::new()
+    )]
+    pub no_operator_reuse: bool,
```
daniilrrr, line 37: I think that where possible, it makes more sense in code (and text) to express the positive version of a thing rather than the negative. 

how about the following or similar? same concept

```suggestion
    pub force_fresh_operator: bool,
```

daniilrrr (commented):
I think it could be merged as-is however I had some suggestions

tvc/src/commands/app/create.rs
richardpringle: Good catch

tvc/src/commands/app/create.rs
richardpringle: I went back and forth on this a few times. I had `--new-operator` initially (which was one of the suggestions). The reason I went with `--no-operator-reuse` was to actually intentionally hide it in a way. The default behaviour is what most people should be using, and only those who opt in should be using the flag `--no-operator-reuse`. If we went with something like "fresh/new operator", I get the sense that it would be something that someone is more likely to try out without understanding that operators _should_ be re-used. 

I'm going to keep this as is for now

tvc/src/commands/app/create.rs
richardpringle: Table driven tests are an anti-pattern in Rust. Each of these tests are run in parallel and give a specific error message when they fail. It's easy to configure cargo to either exit the test harness on failure or run until completion. You give that up if you write table tests the way you do in Go or other languages. 

We could use use the helper crate like ['test-case'](https://docs.rs/test-case/latest/test_case/) do achieve something similar, but it's not worth spending the review time right now

tvc/src/commands/app/create.rs
daniilrrr: ok, could you add a line in the doc comment explaining (probably to an LLM user) why they might want this flag? I think it was originally added because a user was moving back and forth between multiple operators, right?

daniilrrr (approved):
approved with nit

PR #202, opened 2026-07-17
feat(tvc)!: add versioned operator registry

annieke:
This is the first change laying the foundation for hosted operator support in the TVC CLI. It introduces a config version, moves local operator paths into named operator registry entries, and will lazily migrate unversioned configs to v1 (on writes, so `login`, `profile delete`, and `app create`). Explicit operator seed override behavior is preserved.

Note: this is a breaking change, in that once a config is rewritten as v1, an older `tvc` binary will fail to load it due to moving the `operator_key_path`.


richardpringle:
@annieke, the `!` in your commit signals the breaking change, right? Just double checking

tvc/src/commands/login.rs
```diff
@@ -730,12 +750,18 @@ mod tests {
                 OrgConfig {
                     id: "org-test".to_string(),
                     api_key_path: PathBuf::from("api_key.json"),
-                    operator_key_path: PathBuf::from("operator.json"),
                     api_base_url: api_base_url.to_string(),
+                    operator_kind: crate::config::turnkey::OperatorKind::Local,
+                    operators: vec![crate::config::turnkey::OperatorRecord::local(
```
richardpringle: nit: replace qualified path with use
(`cmd + .` in zed or vscode)

tvc/src/config/turnkey/mod.rs
```diff
@@ -24,21 +25,69 @@ const CONFIG_FILE: &str = "tvc.config.toml";
 const ORGS_DIR: &str = "orgs";
 const API_KEY_FILE: &str = "api_key.json";
 const OPERATOR_KEY_FILE: &str = "operator.json";
+const CONFIG_VERSION: u64 = 1;
```
richardpringle: `u64` 😳 ... that's a lot of possible versions

tvc/src/config/turnkey/mod.rs
```diff
@@ -24,21 +25,69 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
+    #[serde(default)]
+    last_created_app_id: HashMap<String, String>,
+    #[serde(default)]
+    last_operator_ids: HashMap<String, Vec<String>>,
+    #[serde(default, flatten)]
+    extra: toml::Table,
+}
+
+#[derive(Debug, Deserialize)]
+struct OrgConfigV0 {
```
richardpringle: ~~Huge nit~~... but it makes a lot easier to read and review code. If you want to keep everything in the same file, I would add an inline `mod`. Or we can have a `config/turnkey/org.rs` (not quite sure why `turnkey` is there, but I guess that's a separate thing). 

I'm happy with either inline or separate file, but I think they need to be together.

That was you can use things like `use config::org::Config` and `use config::org::v1::Config`... could even do `config::v0::Config`. I would leave the type name collision to make it more explicit that you have to rename `config::v0::Config as ConfigV0` forcing callers to be aware that they are simultaneously using multiple types of configs why encouraging use of the latest version. 

So, a few different options here, I'm happy with whatever you like the most, but things need to be consolidated into some types of modules

tvc/src/config/turnkey/mod.rs
```diff
@@ -65,18 +114,94 @@ pub fn dashboard_base_url(api_base_url: &str) -> &'static str {
...
     pub api_base_url: String,
+    /// Operator backend selected for default command resolution.
+    #[serde(default)]
+    pub operator_kind: OperatorKind,
+    /// Durable local and hosted operator metadata.
+    #[serde(default)]
+    pub operators: Vec<OperatorRecord>,
+    /// Unrecognized organization fields from a supported config version.
+    #[serde(default, flatten)]
+    pub extra: toml::Table,
```
richardpringle, line 328: If these all default, why do we even need `OrgConfigV0`?

tvc/src/config/turnkey/mod.rs
```diff
@@ -255,3 +417,176 @@ impl Config {
...
+        }
+    }
+}
+
+impl Serialize for Config {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        ConfigV1::from(self.clone()).serialize(serializer)
```
richardpringle: The implicit type conversion (especially with an implicit clone) is a bit of an anti-pattern. It would be better to have a function that's able to pick and choose the fields to serialize and re-use that function for both `Config` and `ConfigV0`

tvc/src/config/turnkey/mod.rs
```diff
@@ -151,14 +271,51 @@ impl Config {
...
         // Ensure parent directory exists
         if let Some(parent) = path.parent() {
             tokio::fs::create_dir_all(parent).await.with_context(|| {
                 format!("failed to create config directory: {}", parent.display())
             })?;
         }
 
-        let content = toml::to_string_pretty(self).context("failed to serialize config")?;
+        let content = toml::to_string_pretty(&ConfigV1::from(self.clone()))
+            .context("failed to serialize config")?;
```
richardpringle: I'm not a fan of both the implicit conversion and clone. If you need to save the file to path force the caller to do `ConfigV1::from(config).save_to_path(path)`

annieke:
@richardpringle yep! but in that the new configs will be unreadable by older versions of tvc cli, instead of commands themselves breaking

tvc/src/config/turnkey/mod.rs
```diff
@@ -151,14 +271,51 @@ impl Config {
             "saving tvc config"
         );
 
+        self.save_to_path(&path).await
+    }
+
+    async fn load_from_path(path: &Path) -> Result<Self> {
+        if !path.exists() {
+            return Ok(Self::default());
+        }
```
richardpringle: This is exactly why the combinator `unwrap_or_default()` exists. It's better to make the caller aware if they are dealing with a something that was actually loaded from a file or if they are getting a default value. I understand you might not want to call `unwrap_or_default()` in case the error is related to serialization, but if that's the case, you can return an error enum, match on the variant inside the `or_else` combinator and decide if it's a path error to default or if it's a serialization error, then surface it. 

The TL;DR here is that this function is going too much which makes the calling code harder to change/extend

tvc/src/config/turnkey/mod.rs
```diff
@@ -151,14 +271,51 @@ impl Config {
...
+
+        let content = tokio::fs::read_to_string(path)
+            .await
+            .with_context(|| format!("failed to read config file: {}", path.display()))?;
+
+        let header: ConfigVersionHeader = toml::from_str(&content)
+            .with_context(|| format!("failed to parse config file: {}", path.display()))?;
```
richardpringle: This is cool... another reason toml is better than json

tvc/src/config/turnkey/mod.rs
```diff
@@ -255,3 +417,176 @@ impl Config {
...
+
+        config.save_to_path(&path).await.unwrap();
+        let saved = tokio::fs::read_to_string(&path).await.unwrap();
+        assert!(saved.contains("version = 1"));
+        assert!(saved.contains("operator_kind = \"local\""));
+        assert!(saved.contains("kind = \"local\""));
+        assert!(saved.contains("name = \"default\""));
+        assert!(saved.contains("key_path = \"/keys/operator.json\""));
+        assert!(!saved.contains("operator_key_path"));
+        assert!(saved.contains("future_root = \"keep-root\""));
```
richardpringle: nit, prefer the use of raw string literals

tvc/src/config/turnkey/mod.rs
```diff
@@ -255,3 +417,176 @@ impl Config {
...
+        assert!(saved.contains("future_org = 42"));
```
richardpringle: I think it would be better to do an exact match assertion here.

tvc/src/config/turnkey/mod.rs
```diff
@@ -255,3 +417,176 @@ impl Config {
...
+        let original = tokio::fs::read_to_string(&path).await.unwrap();
+        assert_eq!(original, V0_CONFIG);
+
+        let org = &config.orgs["default"];
+        assert_eq!(org.operator_kind, OperatorKind::Local);
+        assert_eq!(org.operators.len(), 1);
+        assert_eq!(org.operators[0].name, DEFAULT_OPERATOR_NAME);
+        let OperatorRecordKind::Local(local) = &org.operators[0].kind else {
+            panic!("migrated operator must be local");
+        };
```
richardpringle, line 608: Just an FYI, you can do 
```rust
#[derive(Serialize, Deserialize)]
#[cfg_attr(test, derive(Debug, PartialEq))]
struct MyType { ... }
```
so that you can use `assert_eq!`

tvc/src/config/turnkey/mod.rs
```diff
@@ -255,3 +417,176 @@ impl Config {
...
+            OperatorRecordKind::Local(_)
+        ));
+        assert!(matches!(
+            org.operators[1].kind,
+            OperatorRecordKind::Hosted(_)
+        ));
+        assert_eq!(reparsed.extra["future_root"].as_str(), Some("keep"));
+        assert_eq!(org.extra["future_org"].as_str(), Some("keep"));
+        let OperatorRecordKind::Local(local) = &org.operators[0].kind else {
+            unreachable!()
```
richardpringle: See my comment above. We shouldn't need to use `unreachable, even in tests`. If this ever gets hit (if someone is making changes to the test), the error is confusing

tvc/src/config/turnkey/qos_operator_key.rs
```diff
@@ -15,9 +15,8 @@ pub struct StoredQosOperatorKey {
 }
 
 impl StoredQosOperatorKey {
-    /// Load operator key from the path specified in org config
-    pub async fn load(org_config: &OrgConfig) -> Result<Option<Self>> {
-        let path = &org_config.operator_key_path;
+    /// Load an operator key from a registered `operator.json` path.
+    pub async fn load(path: &Path) -> Result<Option<Self>> {
```
richardpringle, line 19: 👍

tvc/src/config/turnkey/qos_operator_key.rs
```diff
@@ -40,9 +39,8 @@ impl StoredQosOperatorKey {
         Ok(Some(key))
     }
 
-    /// Save operator key to the path specified in org config
-    pub async fn save(&self, org_config: &OrgConfig) -> Result<()> {
-        let path = &org_config.operator_key_path;
+    /// Save an operator key to a registered `operator.json` path.
+    pub async fn save(&self, path: &Path) -> Result<()> {
```
richardpringle, line 43: 👍

tvc/src/local_operator_key.rs
```diff
@@ -1,23 +1,23 @@
-//! Resolve an operator's QOS key pair from CLI args or the active org config.
+//! Resolve local operator credentials from CLI args or the active org registry.
```
richardpringle, line 1: Should we rename the file/module to `local_operator_key`?

tvc/src/operator_key.rs
```diff
@@ -1,23 +1,23 @@
...
-use crate::config::turnkey::{Config, StoredQosOperatorKey};
+use crate::config::turnkey::{Config, OperatorKind, OperatorRecordKind, StoredQosOperatorKey};
 use crate::pair::{HexSeed, LocalPair};
-use anyhow::{anyhow, bail};
+use anyhow::{Context, anyhow, bail};
 use std::path::PathBuf;
 
 /// An explicit operator master seed given on the command line.
 #[derive(Debug)]
 pub enum OperatorSeedSource {
```
richardpringle: Should this be renamed to `LocalOperatorSeedSource`? 

Seems like there's rejigging of the code that should be done

tvc/src/local_operator_key.rs
```diff
@@ -33,38 +33,95 @@ impl OperatorSeedSource {
...
+            })?;
+            LocalPair::from_hex_seed(&key.private_key)
+                .with_context(|| format!("failed to load operator key: {}", path.display()))
+        }
+    }
+}
+
+/// Return the sole active local record, for workflows such as login that need
+/// its durable key path rather than a loaded key pair.
+pub fn select_local_record<'a>(
```
richardpringle: This should be a method instead of a free function

tvc/src/operator_key.rs
```diff
@@ -33,38 +33,95 @@ impl OperatorSeedSource {
...
+}
+
+/// Return the sole active local record, for workflows such as login that need
+/// its durable key path rather than a loaded key pair.
+pub fn select_local_record<'a>(
+    org_alias: &str,
+    org_config: &'a crate::config::turnkey::OrgConfig,
+) -> anyhow::Result<&'a crate::config::turnkey::LocalOperatorRecord> {
+    if org_config.operator_kind != OperatorKind::Local {
+        bail!("the active operator kind for org '{org_alias}' is hosted")
```
richardpringle: What if in the future we have more operator kinds? You should be able to do something like this:

```suggestion
        bail!("the active operator kind for org '{org_alias}' is {}", org_config.operator_kind)
```

tvc/src/operator_key.rs
```diff
@@ -33,38 +33,95 @@ impl OperatorSeedSource {
...
+    }
+
+    let candidates: Vec<_> = org_config
+        .operators
+        .iter()
+        .filter_map(|operator| match &operator.kind {
+            OperatorRecordKind::Local(local) => Some(local),
+            OperatorRecordKind::Hosted(_) => None,
```
richardpringle: Since this is explicitly for local operators, the code is more easily extended with a wild card here:

```suggestion
            _ => None,
```

tvc/src/local_operator_key.rs
```diff
@@ -33,38 +33,95 @@ impl OperatorSeedSource {
...
+        .filter_map(|operator| match &operator.kind {
+            OperatorRecordKind::Local(local) => Some(local),
+            OperatorRecordKind::Hosted(_) => None,
+        })
+        .collect();
+
+    match candidates.as_slice() {
+        [] => bail!("No local operator configured for org '{org_alias}'"),
+        [operator] => Ok(*operator),
+        _ => bail!("Multiple local operators are configured for org '{org_alias}'"),
```
richardpringle: 🤔  how does this work in interactive mode? Usually we would do some type of selection here... 

I think this is an artifact of where this function is defined. Too many decisions are being made here. It would be better if we had `org_config.get_local_records(alias)` that returned a `Vec<&LocalOperatorRecord>`. Then, depending on where/how the method is being called, we can decide whether it's valid or not to have multiple records instead of always erroring

tvc/src/local_operator_key.rs
```diff
@@ -33,38 +33,95 @@ impl OperatorSeedSource {
...
         }
+        LocalCredentialSource::Explicit(OperatorSeedSource::Path(path)) => {
+            LocalPair::from_master_seed(&path).await
+        }
+        LocalCredentialSource::RegisteredKeyFile(path) => {
+            let key = StoredQosOperatorKey::load(&path).await?.ok_or_else(|| {
+                anyhow!(
+                    "No operator key found at '{}'. Run `tvc login` first or provide \
+                     --operator-seed or --operator-seed-path.",
+                    path.display()
```
richardpringle, line 84: This is a bit of a maintenance burden. I don't think this is the right place to be mentioning any other commands. 

I'm sure this is in other areas of the codebase though, don't need to change it for this PR, but maybe add a TODO?

tvc/src/local_operator_key.rs
```diff
@@ -33,38 +33,95 @@ impl OperatorSeedSource {
...
-            let (alias, org_config) = tvc_config.active_org_config().ok_or_else(|| {
+            let config = Config::load().await?;
+            let (alias, org_config) = config.active_org_config().ok_or_else(|| {
                 anyhow!(
                     "No active organization. Run `tvc login` first or provide \
                      --operator-seed or --operator-seed-path."
                 )
             })?;
+            let local = select_local_record(alias, org_config)?;
+            LocalCredentialSource::RegisteredKeyFile(local.key_path.clone())
```
richardpringle, line 61: You shouldn't need to clone here which means that something else is odd about this code. I don't have a fix right now, so it's fine for this PR... But let me think out loud for a second. 

I think that the issue is we're doing too much inside this function. The dependency for this logic is having a loaded `Config`. If that's the case, we should load the config separately in case it's needed for other reasons. If it's not, it's easy to find an owned local record without cloning. 

I might already have a ticket for fixing the way we call `Config::load`

tvc/src/commands/login.rs
```diff
@@ -530,11 +551,13 @@ fn wait_for_dashboard_registration(ctx: &mut StdCtx) -> Result<()> {
 
 async fn find_or_generate_operator_key(
     ctx: &mut StdCtx,
+    org_alias: &str,
     org_config: &OrgConfig,
```
richardpringle, line 553: The answer might be in the code already... but why do we need the org-alias AND the `OrgConfig`. Isn't the alias only needed to find the config? Something seems off here, but I can't exactly put my finger on it, yet.

richardpringle: So, it looks like it's just used in `select_local_record`, the new function

richardpringle: Ah, I see the pattern. It's just used for generating errors. I think this is a broader problem in the code-base, but I don't think we should make the problem larger by passing down this alias. If that means logging the ID for now then fixing this later, that's a better solution than extending the maintenance burden

tvc/src/commands/login.rs
```diff
@@ -158,8 +161,20 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<()> {
...
     // placed them deliberately and they may live outside our config tree.
-    let uses_default_layout = removed.api_key_path == default_api_key_path(&alias)?
-        && removed.operator_key_path == default_operator_key_path(&alias)?;
+    let default_api_key = default_api_key_path(&alias)?;
+    let default_operator_key = default_operator_key_path(&alias)?;
+    let local_key_paths: Vec<_> = removed
+        .operators
+        .iter()
+        .filter_map(|operator| match &operator.kind {
+            OperatorRecordKind::Local(local) => Some(local.key_path.clone()),
```
richardpringle: You don't need to clone here
```suggestion
            OperatorRecordKind::Local(local) => Some(&local.key_path),
```

tvc/src/commands/login.rs
```diff
@@ -158,8 +161,20 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<()> {
...
+            OperatorRecordKind::Hosted(_) => None,
+        })
+        .collect();
+    let uses_default_layout = removed.api_key_path == default_api_key
+        && local_key_paths
+            .iter()
+            .all(|path| path == &default_operator_key);
```
richardpringle: If you don't clone above, you'll need to do this:
```suggestion
            .all(|path| *path == &default_operator_key);
```

tvc/src/commands/login.rs
```diff
@@ -184,8 +199,14 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<()> {
             "WARNING: custom key paths are configured and were NOT deleted."
         )?;
         shell_eprintln!(ctx, "Remove them manually if no longer needed:")?;
-        shell_eprintln!(ctx, "  {}", removed.api_key_path.display())?;
-        shell_eprintln!(ctx, "  {}", removed.operator_key_path.display())?;
+        let mut custom_paths = BTreeSet::new();
+        custom_paths.insert(removed.api_key_path.clone());
```
```suggestion
        custom_paths.insert(&removed.api_key_path);
```

tvc/src/commands/login.rs
```diff
@@ -184,8 +199,14 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<()> {
...
+        for path in local_key_paths {
+            custom_paths.insert(path);
+        }
```
richardpringle: Just an FYI:
```rust
let custom_paths =
    local_key_paths.into_iter().chain(Some(&removed.api_key_path)).collect::<BTreeSet<_>>();

File: tvc/src/config/turnkey/mod.rs [hunk 28]
```diff
@@ -24,21 +25,69 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
+}
+
+#[derive(Debug, Deserialize)]
+struct ConfigVersionHeader {
+    version: Option<u64>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct ConfigV1 {
+    version: u64,
```
richardpringle: We shouldn't have this... we could do something like 
```rust
struct VersionedConfig<C> {
    #[serde(flatten)]
    header: ConfigVersionHeader;
    #[serde(flatten)]
    config: C
}
```

But we shouldn't create types that have invalid states when there are workarounds.

tvc/src/config/turnkey/mod.rs
```diff
@@ -24,21 +25,69 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
+    last_operator_ids: HashMap<String, Vec<String>>,
+    #[serde(default, flatten)]
+    extra: toml::Table,
+}
+
+#[derive(Debug, Deserialize)]
+struct ConfigV0 {
+    active_org: Option<String>,
+    #[serde(default)]
+    orgs: HashMap<String, OrgConfigV0>,
```
richardpringle: Just confirming that the different is just `version` and `OrgConfigV0`, right? This should be explicitly stated in doc comments

tvc/src/config/turnkey/mod.rs
```diff
@@ -24,21 +25,69 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
+struct ConfigVersionHeader {
+    version: Option<u64>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct ConfigV1 {
+    version: u64,
+    active_org: Option<String>,
+    #[serde(default)]
+    orgs: HashMap<String, OrgConfig>,
```
richardpringle: So we have `Config`, `ConfigV0`, and `ConfigV1` but we just have `OrgConfig` and `OrgConfigV0`. Why aren't we consistent here? Or rather, this seems inconsistent to me, is there a good reason not to follow the same pattern?

tvc/src/config/turnkey/mod.rs
```diff
@@ -24,21 +25,69 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
-    #[serde(default)]
     pub orgs: HashMap<String, OrgConfig>,
     /// Map of org alias -> last created app ID (for convenience)
-    #[serde(default)]
     pub last_created_app_id: HashMap<String, String>,
     /// Map of org alias -> last manifest set operator IDs (for convenience)
-    #[serde(default)]
     pub last_operator_ids: HashMap<String, Vec<String>>,
+    /// Unrecognized top-level fields from a supported config version.
+    pub extra: toml::Table,
```
richardpringle: Why do we need this? We aren't using it anywhere except in tests.

tvc/src/config/turnkey/mod.rs
```diff
@@ -24,21 +25,69 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
+    version: u64,
+    active_org: Option<String>,
+    #[serde(default)]
+    orgs: HashMap<String, OrgConfig>,
+    #[serde(default)]
+    last_created_app_id: HashMap<String, String>,
+    #[serde(default)]
+    last_operator_ids: HashMap<String, Vec<String>>,
+    #[serde(default, flatten)]
+    extra: toml::Table,
```
richardpringle: same comment

tvc/src/config/turnkey/mod.rs
```diff
@@ -24,21 +25,69 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
+struct ConfigV0 {
+    active_org: Option<String>,
+    #[serde(default)]
+    orgs: HashMap<String, OrgConfigV0>,
+    #[serde(default)]
+    last_created_app_id: HashMap<String, String>,
+    #[serde(default)]
+    last_operator_ids: HashMap<String, Vec<String>>,
+    #[serde(default, flatten)]
+    extra: toml::Table,
```
richardpringle: same comment

tvc/src/config/turnkey/mod.rs
```diff
@@ -24,21 +25,69 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
+
+#[derive(Debug, Deserialize)]
+struct OrgConfigV0 {
+    id: String,
+    api_key_path: PathBuf,
+    operator_key_path: PathBuf,
+    #[serde(default = "default_api_base_url")]
+    api_base_url: String,
+    #[serde(default, flatten)]
+    extra: toml::Table,
```
richardpringle: same comment

tvc/src/config/turnkey/mod.rs
```diff
@@ -65,18 +114,94 @@ pub fn dashboard_base_url(api_base_url: &str) -> &'static str {
...
     /// Path to the API key file
     pub api_key_path: PathBuf,
-    /// Path to the operator key file
-    pub operator_key_path: PathBuf,
     /// API base URL for this organization
     #[serde(default = "default_api_base_url")]
     pub api_base_url: String,
+    /// Operator backend selected for default command resolution.
+    #[serde(default)]
+    pub operator_kind: OperatorKind,
```
richardpringle: So we've got `OperatorKind` and `OperatorRecordKind`. Something seems a little off to me. In theory, operators can be either kind, right? Why are we doing this again, just for the sake of backwards compatibility? I know I commented on this in your proposal, but seeing the code is a little different. 

If we have multiple operators, should we just have something like `default_operator_kind: OperatorKind`. 

I'm looking at how this is used and the only place this is actually called is in `select_operator_record`, where we bail if it's not a local kind. It seems like we don't really need this. You could just search for a local operator and bail if none exist, right?

tvc/src/config/turnkey/mod.rs
```diff
@@ -65,18 +114,94 @@ pub fn dashboard_base_url(api_base_url: &str) -> &'static str {
...
+    /// Unique human-readable name within the organization.
+    pub name: String,
+    #[serde(flatten)]
+    pub kind: OperatorRecordKind,
+}
+
+impl OperatorRecord {
+    pub fn local(name: impl Into<String>, key_path: PathBuf) -> Self {
+        Self {
+            name: name.into(),
```
richardpringle: The `name` arg is only ever `"default"`. Why don't we just remove the argument from the `local` static method until we have reason to make it configurable?

richardpringle (changes requested):
Okay, so there are a lot of comments here, but the only real thing that I think is blocking is figuring out the `Config` types in the most extensible way possible. 

I think we should do something like the `VersionConfig<C>` as I suggested in one of the comments and if the version isn't specified, assume version 0, migrate, save then continue on as normal. 

So I think the only comments that require changes right now are in `tvc/src/config/turnkey/mod.rs`. 

Happy to talk about this on a call though. It's definitely a bigger priority than any of the other changes to the cli

tvc/src/config/turnkey/mod.rs
annieke: more reasonable to use `u16` or something even smaller?

tvc/src/local_operator_key.rs
annieke: oo yes

tvc/src/config/turnkey/mod.rs
annieke: used u16 c12301e3ce57e74b22fb298a7a8f21f215fe931e

tvc/src/config/turnkey/mod.rs
annieke: added an inline `disk` mod for read and editing the DiskConfig c12301e3ce57e74b22fb298a7a8f21f215fe931e

tvc/src/config/turnkey/mod.rs
annieke: simplified this into `vo::Config` and explained we basically want to just parse out `operator_key_path` c12301e3ce57e74b22fb298a7a8f21f215fe931e

tvc/src/config/turnkey/mod.rs
annieke: refactored to saving as `disk::to_toml(self)` without cloning or implicit conversion c12301e3ce57e74b22fb298a7a8f21f215fe931e

tvc/src/config/turnkey/mod.rs
annieke: split this function up
- `Config::load` owns missing-file defaulting
- `load_from_path` only performs file io
- disk mod handles version parsing and migrating

c12301e3ce57e74b22fb298a7a8f21f215fe931e

tvc/src/local_operator_key.rs
annieke: made it a method of `OrgConfig` c12301e3ce57e74b22fb298a7a8f21f215fe931e

tvc/src/local_operator_key.rs
annieke: didn't implement this in interactive mode; for now we still assume one local record and error on multiple

tvc/src/config/turnkey/mod.rs
annieke: c12301e3ce57e74b22fb298a7a8f21f215fe931e

tvc/src/config/turnkey/mod.rs
annieke: c12301e3ce57e74b22fb298a7a8f21f215fe931e

tvc/src/config/turnkey/mod.rs
annieke: renamed to `default_operator_kind`; this will be the switch we flip when we want to default to hosted keys, and it's not used yet but going to add to it when i start adding the hosted key creation + usage commands

tvc/src/config/turnkey/mod.rs
annieke: i kept this in so extra fields the CLI doesn't expect will not be deleted, juuust in case. as i type that i'm realizing maybe that's not ideal, but since this is our first time migrating the config this could be useful. maybe we warn on this not being empty?

annieke:
@richardpringle 
- I addressed all straightforward comments / nits in c0ee64513a6e5bf820a83d2959e1ab4f1643b726 and resolved those comments
- refactored into the [wrapping config type](https://github.com/tkhq/rust-sdk/pull/202#discussion_r3616965547) in c12301e3ce57e74b22fb298a7a8f21f215fe931e
- kept toml::Table to preserve unexpected fields; see this comment: https://github.com/tkhq/rust-sdk/pull/202#discussion_r3618004989 (
- didn't rip out org_alias from login.rs or touch the debug logs since this is a greater issue and for now this might be helpful for me to debug; lmk if this is grating though

tvc/src/config/turnkey/mod.rs
richardpringle: oooooh, it's for re-writing (save)? I think you just need a doc comment explaining this. Smart, I didn't think about that

tvc/src/local_operator_key.rs
richardpringle: Yeah, this is part of a bigger problem. Here's a ticket in case you're interested:  https://linear.app/turnkey/issue/TVC-202/derive-command-references-from-clap-and-isolate-user-facing-messaging

tvc/src/local_operator_key.rs
richardpringle: Since you have to make changes... because of conflicts, would you (@annieke) mind adding `TODO(TVC-202)` here? as a comment

tvc/src/config/turnkey/mod.rs
```diff
@@ -39,6 +48,156 @@ pub struct Config {
...
+        let mut org: OrgConfig = toml::Value::Table(table)
+            .try_into()
+            .with_context(|| format!("failed to migrate v0 config for organization '{alias}'"))?;
+
+        org.default_operator_kind = OperatorKind::Local;
+        org.operators = vec![OperatorRecord::local(operator_key_path)];
+        Ok(org)
+    }
+
+    pub(super) mod v0 {
```
richardpringle, line 179: Yes! Very obvious this shouldn't be used with only `super` visibility, I like it!

tvc/src/config/turnkey/mod.rs
```diff
@@ -65,18 +224,135 @@ pub fn dashboard_base_url(api_base_url: &str) -> &'static str {
...
+    #[default]
+    Local,
+    Hosted,
+}
+
+impl std::fmt::Display for OperatorKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Local => f.write_str("local"),
+            Self::Hosted => f.write_str("hosted"),
```
richardpringle: Another FYI, you *can* (but do not need to) do this:

```suggestion
            Self::Local => "local".fmt(f),
            Self::Hosted => "hosted".fmt(f),
```

For another time, though: 

Would be nice to have [`strum`](https://docs.rs/strum/latest/strum/index.html) here... you can do fancy things like get the names of enums. An older version is already in our list of approved dependencies

tvc/src/config/turnkey/mod.rs
```diff
@@ -65,18 +224,135 @@ pub fn dashboard_base_url(api_base_url: &str) -> &'static str {
...
+                OperatorRecordKind::Local(local) => Some(local),
+                _ => None,
+            })
+            .collect();
+
+        match candidates.as_slice() {
+            [] => bail!("No local operator configured for org '{org_alias}'"),
+            [operator] => Ok(*operator),
+            _ => bail!("Multiple local operators are configured for org '{org_alias}'"),
+        }
```
richardpringle, line 356: You don't need to change this now, but can you add a todo here. I think I said something in another comment about being able to decide what to do with the "candidates" outside of this function. Right now, *how/where* this function is being called is informing the error messages and even the logic inside of it. That means we've tightly coupled this function with all of its current callsites. tight coupling makes things very hard to change, and while this is only one layer, all it takes is calling another tightly coupled function inside this one and it's turtles all the way down

tvc/src/config/turnkey/mod.rs
richardpringle, line 1: Still a few things in here that I would like to see changes (in a follow up... eventually), but this looks so much better in general, 🥇

richardpringle (dismissed):
Would be nice if you add a couple TODOs that I mentioned since you have to make changes anyway 😄 ... 

But in general, I really like the changes and the shape of the config looks good to me

richardpringle (approved):

PR #203, opened 2026-07-18
tvc: validate org ID on login, document --api-base-url, add AGENTS.md

yourbuddyconner:
## 1. Validate org ID on `tvc login` before persisting config

Previously `tvc login` called `config.save()` before running the `whoami`
verification, so an invalid or unreachable org ID left a config entry on disk
pointing at nothing. Users then hit an opaque error on the next `tvc` command
instead of a clear "that org ID is wrong" message during login.

- `execute_login` in `tvc/src/commands/login.rs` now reorders so
  `verify_credentials` (the lightweight `GetWhoami` call, already present) runs
  **before** `config.save()`. No new API surface added.
- Verification failures are wrapped via a new `org_verification_error` helper
  that produces the user-actionable message specified in the ask:
  > organization ID `<id>` could not be verified. Please check the ID and try again.
  > You can copy the org ID from the Turnkey dashboard home page.
- Added an integration test `login_rejects_unverifiable_org_before_persisting_config`
  in `tvc/tests/login.rs`. It seeds a config with a valid keypair but points
  `api_base_url` at an unroutable local port (`http://127.0.0.1:1`), runs
  `tvc login --non-interactive --org test`, and asserts both the new error
  message and that the on-disk config file is byte-for-byte unchanged after
  the failed login.

## 2. Document `--api-base-url`

The `--api-base-url` flag (and its `TVC_API_BASE_URL` env var) exists on
`tvc login` but was effectively undocumented — the clap doc-comment was a
single terse sentence and the README didn't mention the flag at all.

- Expanded the `#[arg]` doc-comment in `tvc/src/commands/login.rs` so
  `tvc login --help` prints the default (`https://api.turnkey.com`), the use
  case (non-prod Turnkey environments), and how the stored value flows through
  to subsequent commands. (Shipped in the login commit because they touch the
  same struct.)
- Added a **Configuration** section to `tvc/README.md` documenting the flag,
  its default, when to reach for it, and both interactive (`tvc login
  --api-base-url ...`) and CI (`TVC_API_BASE_URL=... tvc app list`) worked
  examples.

## 3. Add `tvc/AGENTS.md`

New agent onboarding guide at `tvc/AGENTS.md` (using the emerging `AGENTS.md`
cross-tool convention rather than a tool-specific filename). Audience is AI
coding agents driving `tvc` on behalf of a human. Structure is a runnable
playbook:

1. **What TVC is** — Nitro Enclaves + remote attestation, 2 sentences.
2. **Prerequisites** — including the org-ID-is-on-dashboard-home-page callout.
3. **First-time `tvc login`** — with the whoami verification step called out.
4. **Common workflows** — `app init/create/list/status`, `deploy
   init/create/approve/status/get-status/debug-logs`, `app set-live-deploy`,
   deletes. Every command name verified by reading `tvc/src/commands/` — no
   invented flags or subcommands.
5. **Configuration overrides** — `--non-interactive`, JSON mode,
   `--api-base-url`, per-command env vars, and precedence order.
6. **Where things live** — pointers to `tkhq/tvc-examples` and the Turnkey
   docs.
7. **Common gotchas** — org ID source, TTY-required first login, non-
   interactive login can't generate keys, debug-log double opt-in, `status`
   vs `get-status`, all-or-nothing env auth, and a note that failed logins
   now leave the config untouched (from change 1 above).

## Testing

Ran from the repo root:

```
cargo build -p tvc                        # ok
cargo test  -p tvc                        # all suites pass; 141 unit tests +
                                          # all integration suites; new
                                          # login_rejects_unverifiable_org_before_persisting_config
                                          # among them
cargo fmt   -p tvc -- --check             # clean
cargo clippy -p tvc --all-targets -- -D warnings   # clean
```

Spot-checked `tvc login --help` output to confirm the new `--api-base-url`
help text renders as expected.

## Scope discipline

- No API contract changes.
- No new CLI subcommands.
- No changes outside `tvc/`.
- No changes to existing test fixtures (new test seeds its own config).

### DIFF
--- a/tvc/AGENTS.md
+++ b/tvc/AGENTS.md
@@ -1,0 +1,200 @@
+# TVC CLI — Agent Guide
+
+This document teaches an AI coding agent (Claude Code, Cursor, similar) how to
+drive the `tvc` CLI end-to-end on behalf of a human. It is a runnable playbook,
+not a reference manual — for exhaustive flag listings run `tvc <command> --help`.
+
+## 1. What TVC is
+
+`tvc` is the CLI for **Turnkey Verifiable Cloud (TVC)** — a platform that runs
+your code inside AWS Nitro Enclaves with cryptographic remote attestation.
+`tvc` is how you (or the agent driving on your behalf) authenticate, create
+applications, deploy them, approve deployment manifests, and inspect live
+deployment state.
+
+Concepts and dashboard docs live at
+<https://docs.turnkey.com/features/verifiable-cloud/overview>.
+
+## 2. Prerequisites
+
+Before running any `tvc` command, confirm all of the following:
+
+1. The user has a **Turnkey account** and belongs to at least one organization.
+2. You know the **organization ID** — a UUID like `01234567-89ab-cdef-0123-456789abcdef`. It is on the Turnkey dashboard **home page** (click-to-copy). Do NOT read an ID off a deployment or app detail page — that's a different UUID.
+3. The user has (or can create) **Turnkey API credentials** for that org. `tvc login` will generate a keypair for you and walk the user through registering the public key on the dashboard.
+4. The `tvc` binary is installed. Preferred: `cargo install tvc`. Verify with `tvc --version`.
+
+If any of these are missing, stop and ask the user rather than guessing.
+
+## 3. First-time setup: `tvc login`
+
+Run `tvc login` interactively (do NOT pass `--non-interactive` on first setup —
+the flow needs to prompt for the org ID and pause for dashboard registration).
+
+```bash
+tvc login
+```
+
+The flow:
+
+1. If no orgs are configured yet, `tvc` prints the dashboard URL and prompts:
+   - **Organization ID** — paste the UUID from the dashboard home page.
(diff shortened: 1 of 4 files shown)


tvc/AGENTS.md
```diff
@@ -0,0 +1,200 @@
...
+tvc deploy debug-logs --deploy-id <DEPLOYMENT_UUID>              # one-shot
+tvc deploy debug-logs --deploy-id <DEPLOYMENT_UUID> --poll       # follow
+```
+
+### Set the live deployment
+
+```bash
+tvc app set-live-deploy --app-id <APP_UUID> --deploy-id <DEPLOYMENT_UUID>
+```
+
```
am-hernandez, line 157:
```suggestion
### Set the live deployment

```bash
tvc app set-live-deploy --deploy-id <DEPLOYMENT_UUID>
```

---

`set-live-deploy` does not take an argument of `--app-id`, and if we pass one we will see:

```
error: unexpected argument '--app-id' found

Usage: tvc app set-live-deploy [OPTIONS] --deploy-id <DEPLOY_ID>
```

So we should remove `--app-id <APP_UUID>` here.

PR #204, opened 2026-07-20
feat(tvc)!: add hosted quorum key creation and rename local key commands

annieke:
This PR renames existing commands to be explicitly `local`, and adds a new `tvc keys create-quorum-key` command for creating hosted quorum keys:
```bash
tvc keys create-quorum-key \
  --threshold <N> \
  --operator-encrypt-keys <HEX,HEX,...>
```

Once https://github.com/tkhq/rust-sdk/pull/202 (operator registry tracking hosted operators) is merged, this command can be refactored to also take a mutually exclusive `--operators` arg with `id`s instead of explicit keys, with keys populated by the registry.


tvc/src/commands/keys/create_quorum_key.rs
```diff
@@ -0,0 +1,340 @@
...
+#[derive(Debug, ClapArgs)]
+#[command(about, long_about = None)]
+pub struct Args {
+    /// Number of operator shares required to reconstruct the quorum key.
+    #[arg(long, env = "TVC_QUORUM_KEY_THRESHOLD")]
+    pub threshold: u8,
+
+    /// Comma-separated, uncompressed P-256 operator encryption public keys.
+    #[arg(long, value_name = "HEX,HEX,...", env = "TVC_OPERATOR_ENCRYPT_KEYS")]
+    pub operator_encrypt_keys: String,
```
Avi-D-coder, line 29: non blocking nit: if we introduced a `struct PubKey` we could give it a `FromStr` impl.
Then we could use https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_delimiter


```rust
#[arg(
    long,
    value_name = "HEX",
    value_delimiter = ',',
    env = "TVC_OPERATOR_ENCRYPT_KEYS"
)]
pub operator_encrypt_keys: Vec<PubKey>,
```


Here pub key could be `String` or a crypto type that validates it if we already have on in the package (not sure).

Avi-D-coder (approved):

tvc/src/commands/keys/create_quorum_key.rs
annieke: oo this is a great suggestion! i will add it in my follow up 🫡

PR #205, opened 2026-07-22
feat(tvc): add hosted operator creation and deploy approval

annieke:
This PR adds the `tvc operator create` command for creating and locally registering Turnkey-hosted operators, and extends `tvc deploy approve` to support hosted-operator approvals. It does not yet change any existing behavior (eg flip an org's `default_operator_kind` or set any defaults), and requires explicitly providing a hosted operator's UUID.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4296,6 +4296,9 @@ name = "uuid"
 version = "1.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ced87ca4be083373936a67f8de945faa23b6b42384bd5b64434850802c6dccd0"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "vcpkg"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ qos_p256 = { version = "0.12.1", default-features = false }
 base64 = { version = "0.22.0", default-features = false, features = ["std"] }
 bs58 = { version = "0.5.0", features = ["std", "check"], default-features = false }
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
-uuid = { version = "1.13.1", default-features = false }
+uuid = { version = "1.13.1", default-features = false, features = ["serde", "std"] }
 serde = { version = "1.0.219", default-features = false, features = ["std", "derive"] }
 serde_json = { version = "1.0.140", default-features = false, features = ["std"] }
 serde_bytes = { version = "0.11", default-features = false }
--- a/tvc/src/commands/deploy/approve.rs
+++ b/tvc/src/commands/deploy/approve.rs
@@ -1,32 +1,33 @@
 //! Approve deploy command - cryptographically approve a QOS manifest.
 
-use crate::config::turnkey::Config;
-use crate::local_operator_key::{LocalOperatorSeedSource, resolve_local_operator};
-use crate::outcome::Outcome;
-use crate::output::StdCtx;
-use crate::pair::HexSeed;
-use crate::prompts;
-use crate::prompts::{bail_required_in_non_interactive, stdin_can_prompt};
-use crate::util::{read_file_to_string, write_file};
-use crate::{shell_print, shell_println};
+use crate::{
+    client::build_client,
+    config::turnkey::Config,
+    local_operator_key::LocalOperatorSeedSource,
+    operator::{OperatorCtx, ResolvedOperator, resolve_operator},
+    outcome::Outcome,
+    output::StdCtx,
+    pair::HexSeed,
+    prompts::{self, bail_required_in_non_interactive, stdin_can_prompt},
+    shell_print, shell_println,
+    util::{read_file_to_string, write_file},
+};
 use anyhow::{Context, anyhow, bail};
 use clap::{ArgGroup, Args as ClapArgs};
-use qos_core::protocol::services::boot::Approval;
 use qos_core::protocol::services::boot::{
-    ManifestSet, Namespace, NitroConfig, QuorumMember, ShareSet, VersionedManifest,
+    Approval, ManifestSet, Namespace, NitroConfig, QuorumMember, ShareSet, VersionedManifest,
 };
 use serde::{Serialize, Serializer};
 use std::collections::HashSet;
-use std::fmt::Write;
-use std::fmt::{self, Display, Formatter};
-use std::path::Path;
-use std::path::PathBuf;
+use std::fmt::{self, Display, Formatter, Write};
+use std::path::{Path, PathBuf};
 use std::time::{SystemTime, UNIX_EPOCH};
 use tracing::debug;
 use turnkey_client::generated::external::data::v1::TvcDeployment;
 use turnkey_client::generated::{
     CreateTvcManifestApprovalsIntent, GetTvcDeploymentRequest, TvcManifestApproval,
 };
+use uuid::Uuid;
 
 const QUORUM_REACHED_MESSAGE: &str =
     "Manifest approval quorum reached. Your deployment will be available soon.";
@@ -60,10 +61,11 @@ pub struct Args {
     #[arg(long, env = "TVC_MANIFEST_ID")]
     pub manifest_id: Option<String>,
 
-    /// Turnkey operator ID (UUID) for the approving operator.
-    /// Required when posting approval to the API.
+    /// Turnkey operator UUID for the approving operator. When posting without
+    /// this flag, TVC selects from the last manifest's saved operator IDs. A
+    /// configured hosted operator is selected by its UUID.
     #[arg(long, env = "TVC_OPERATOR_ID")]
-    pub operator_id: Option<String>,
+    pub operator_id: Option<Uuid>,
 
     /// Hex-encoded 32-byte master seed for the operator key.
     /// If no seed flag is provided, uses the operator key from the logged-in org config.
@@ -96,14 +98,15 @@ pub struct Args {
     #[arg(short = 'o', long, value_name = "PATH", env = "TVC_APPROVAL_OUT")]
     pub approval_out: Option<PathBuf>,
 
-    /// Don't post approval to the API.
+    /// Generate a local-operator approval without posting it to the API. This
+    /// supports offline signing and cannot be used with a hosted operator.
     #[arg(long, env = "TVC_SKIP_POST")]
     pub skip_post: bool,
 }
 
 struct PostApprovalPlan<'a> {
     manifest_id: &'a str,
-    operator_id: &'a str,
+    operator_id: &'a Uuid,
     deploy_id: Option<&'a str>,
 }
 
@@ -238,15 +241,16 @@ impl Display for ApprovalDryRun {
 
 struct PostTarget {
     manifest_id: String,
-    operator_id: String,
     deploy_id: Option<String>,
 }
 
 struct ResolvedApproveInputs {
     manifest: VersionedManifest,
     operator_seed_source: Option<LocalOperatorSeedSource>,
+    operator_id: Option<Uuid>,
     approval_out: Option<PathBuf>,
     dry_run: bool,
+    skip_post: bool,
     post_target: Option<PostTarget>,
 }
 
@@ -285,11 +289,11 @@ async fn build_inputs_interactive(
         interactive_approve(ctx, &manifest)?;
     }
 
+    let mut operator_id = args.operator_id;
     let post_target = if !args.dry_run && !args.skip_post {
         let manifest_id = resolve_manifest_id(&args, fetched_manifest_id.as_deref())?;
-        let operator_id = match &args.operator_id {
-            Some(id) => id.clone(),
-            None => {
+        if operator_id.is_none() {
+            operator_id = Some({
                 let saved_ids = load_saved_operator_ids().await?;
                 match &*saved_ids {
                     [] => bail!(
@@ -297,21 +301,18 @@ async fn build_inputs_interactive(
                          No saved operator IDs found. \
                          Use --skip-post to only generate the approval locally."
                     ),
-                    [id] => id.clone(),
-                    _ if stdin_can_prompt() => prompts::select(
-                        "Select approving operator",
-                        saved_ids.iter().map(String::as_str).collect::<Vec<_>>(),
-                    )?
-                    .to_string(),
+                    [id] => *id,
+                    _ if stdin_can_prompt() => {
+                        prompts::select("Select approving operator", saved_ids)?
+                    }
                     _ => bail!(
                         "--operator-id is required to post approval to API when multiple saved operator IDs are available"
                     ),
                 }
-            }
-        };
+            });
+        }
         Some(PostTarget {
             manifest_id,
-            operator_id,
             deploy_id: args.deploy_id.clone(),
         })
     } else {
@@ -321,8 +322,10 @@ async fn build_inputs_interactive(
     Ok(ResolvedApproveInputs {
         manifest,
         operator_seed_source,
+        operator_id,
         approval_out: args.approval_out,
         dry_run: args.dry_run,
+        skip_post: args.skip_post,
         post_target,
     })
 }
@@ -338,11 +341,11 @@ async fn build_inputs_non_interactive(
 
     let (manifest, fetched_manifest_id) = load_manifest(ctx, &args).await?;
 
+    let mut operator_id = args.operator_id;
     let post_target = if !args.dry_run && !args.skip_post {
         let manifest_id = resolve_manifest_id(&args, fetched_manifest_id.as_deref())?;
-        let operator_id = match &args.operator_id {
-            Some(id) => id.clone(),
-            None => {
+        if operator_id.is_none() {
+            operator_id = Some({
                 let saved_ids = load_saved_operator_ids().await?;
                 match &*saved_ids {
                     [] => bail!(
@@ -350,16 +353,15 @@ async fn build_inputs_non_interactive(
                          No saved operator IDs found. \
                          Use --skip-post to only generate the approval locally."
                     ),
-                    [id] => id.clone(),
+                    [id] => *id,
                     _ => bail!(
                         "--operator-id is required to post approval to API when multiple saved operator IDs are available"
                     ),
                 }
-            }
-        };
+            });
+        }
         Some(PostTarget {
             manifest_id,
-            operator_id,
             deploy_id: args.deploy_id.clone(),
         })
     } else {
@@ -369,8 +371,10 @@ async fn build_inputs_non_interactive(
     Ok(ResolvedApproveInputs {
         manifest,
         operator_seed_source,
+        operator_id,
         approval_out: args.approval_out,
         dry_run: args.dry_run,
+        skip_post: args.skip_post,
         post_target,
     })
 }
@@ -383,8 +387,20 @@ async fn run_with_resolved_inputs(
         return Ok(ApproveOutcome::DryRun(ApprovalDryRun {}));
     }
 
+    let operator = resolve_operator(inputs.operator_seed_source, inputs.operator_id).await?;
+    if inputs.skip_post && operator.is_hosted() {
+        bail!("--skip-post is only supported for local operators");
+    }
+    let auth = if operator.is_hosted() {
+        Some(build_client().await?)
+    } else {
+        None
+    };
     let approval = sign_and_write_approval(
-        inputs.operator_seed_source,
+        &operator,
+        &OperatorCtx {
+            auth: auth.as_ref(),
+        },
         inputs.approval_out.as_deref(),
         &inputs.manifest,
     )
@@ -404,9 +420,12 @@ async fn run_with_resolved_inputs(
 
     match inputs.post_target {
         Some(target) => {
+            let operator_id = operator
+                .id()
+                .context("resolved operator ID required to post approval")?;
             let plan = PostApprovalPlan {
                 manifest_id: target.manifest_id.as_str(),
-                operator_id: target.operator_id.as_str(),
+                operator_id: &operator_id,
                 deploy_id: target.deploy_id.as_deref(),
             };
             let posted = post_approval_to_api(ctx, plan, &approval).await?;
@@ -415,7 +434,7 @@ async fn run_with_resolved_inputs(
                 approval: inline_approval,
                 written_to,
                 manifest_id: target.manifest_id,
-                operator_id: target.operator_id,
+                operator_id: operator_id.to_string(),
                 approval_ids: posted.approval_ids,
                 quorum_reached: posted.quorum_reached,
             }))
@@ -445,14 +464,12 @@ async fn load_manifest(
 /// that file. Reporting the approval (inline payload or file path) is the
 /// terminal outcome's job.
 async fn sign_and_write_approval(
-    operator_seed_source: Option<LocalOperatorSeedSource>,
+    operator: &ResolvedOperator,
+    operator_ctx: &OperatorCtx<'_>,
     approval_out: Option<&Path>,
     manifest: &VersionedManifest,
 ) -> anyhow::Result<Approval> {
-    let pair: Box<dyn crate::pair::Pair> =
-        Box::new(resolve_local_operator(operator_seed_source).await?);
-
-    let approval = generate_approval(pair, manifest).await?;
+    let approval = operator.approve_manifest(operator_ctx, manifest).await?;
 
     if let Some(path) = approval_out {
         let json = serde_json::to_string_pretty(&approval)?;
@@ -474,9 +491,16 @@ fn resolve_manifest_id(args: &Args, fetched_manifest_id: Option<&str>) -> anyhow
         })
 }
 
-async fn load_saved_operator_ids() -> anyhow::Result<Vec<String>> {
+async fn load_saved_operator_ids() -> anyhow::Result<Vec<Uuid>> {
     let config = Config::load().await?;
-    Ok(config.get_last_operator_ids().unwrap_or_default())
+    config
+        .get_last_operator_ids()
+        .unwrap_or_default()
+        .into_iter()
+        .map(|id| {
+            Uuid::parse_str(&id).with_context(|| format!("saved operator ID '{id}' is not a UUID"))
+        })
+        .collect()
 }
 
 async fn post_approval_to_api(
@@ -490,7 +514,7 @@ async fn post_approval_to_api(
     let auth = crate::client::build_client().await?;
 
     let tvc_approval = TvcManifestApproval {
-        operator_id: plan.operator_id.into(),
+        operator_id: plan.operator_id.to_string(),
         signature: hex::encode(&approval.signature),
     };
 
@@ -566,29 +590,6 @@ fn manifest_approval_quorum_reached(deployment: &TvcDeployment) -> bool {
     }
 
     operator_ids.len() >= manifest_set.threshold as usize
-}
-
-async fn generate_approval(
-    pair: Box<dyn crate::pair::Pair>,
-    manifest: &VersionedManifest,
-) -> anyhow::Result<Approval> {
-    let public_key = pair.public_key();
-    let member = manifest
-        .manifest_set()
-        .members
-        .iter()
-        .find(|m| m.pub_key == public_key)
-        .cloned()
-        .ok_or_else(|| {
-            anyhow!(
-                "operator ({}) not part of manifest set",
-                hex::encode(&public_key)
-            )
-        })?;
-
-    let signature = pair.sign(manifest.manifest_hash().to_vec()).await?;
-
-    Ok(Approval { signature, member })
 }
 
 /// Walk the user through each section of the manifest for approval.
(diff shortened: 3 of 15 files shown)


tvc/src/commands/deploy/approve.rs
```diff
@@ -1,7 +1,9 @@
 //! Approve deploy command - cryptographically approve a QOS manifest.
 
+use crate::client::build_client;
 use crate::config::turnkey::Config;
-use crate::local_operator_key::{LocalOperatorSeedSource, resolve_local_operator};
+use crate::local_operator_key::LocalOperatorSeedSource;
+use crate::operator::{OperatorCtx, ResolvedOperator, resolve_operator};
```
richardpringle: No need to change for this PR, but I would love it if we made a best effort to keep imports "merged". Eventually, `rustfmt` should do this, but it's taking them a while to land this in stable: [issue](https://github.com/rust-lang/rustfmt/issues/4991)

tvc/src/commands/deploy/approve.rs
```diff
@@ -58,8 +60,9 @@ pub struct Args {
     #[arg(long, env = "TVC_MANIFEST_ID")]
     pub manifest_id: Option<String>,
 
-    /// Turnkey operator ID (UUID) for the approving operator.
-    /// Required when posting approval to the API.
+    /// Turnkey operator ID (UUID) for the approving operator. A configured
+    /// hosted operator is selected by this ID. Required when posting approval
```
richardpringle: Is this always a hosted operator now? What happens if this isn't specified?

tvc/src/commands/deploy/approve.rs
```diff
@@ -94,7 +97,8 @@ pub struct Args {
     #[arg(short = 'o', long, value_name = "PATH", env = "TVC_APPROVAL_OUT")]
     pub approval_out: Option<PathBuf>,
 
-    /// Don't post approval to the API.
+    /// Don't post the manifest approval to the API. Hosted signing still
+    /// submits a raw-payload signing activity.
```
richardpringle: Not a blocker, but should we really say this? If I'm a user, I might ask what a "raw-payload signing activity" is and what's in the payload. Should we still be posting? Why does this option exist... 

I feel like this is one of those cases where I want the help text to say "if you don't know what this does already, you probably shouldn't be using it" then maybe point to the docs. 

No need to change anything, I'm just curious

tvc/src/commands/deploy/approve.rs
```diff
@@ -153,11 +158,13 @@ async fn build_inputs_interactive(
         interactive_approve(ctx, &manifest)?;
     }
 
+    let operator_id = args.operator_id.clone();
```
richardpringle: Can do this in a follow up, but if you're adding the `operator_id` to `ResolvedApproveInput`, it no longer needs to be in `PostTarget`  because the only place `PostTarget` is used, there will always be the same operator-id. 

This is why I generally like to treat `clone` as a code smell. Sometimes it genuinely makes the code easier to read/write instead of adding a bunch of lifetime generics everywhere, but for a place like this, cloning `operator_id` means that you can treat `ResolvedApproveInputs` and `PostTarget` as out of sync and can therefore represent invalid state. 

"But Richard, the code is so simple, how could anyone ever mess this up?" you might ask... and I would say that's fair, but unless you can also write tests to prevent it, the one time someone makes a code change in a hurry and it isn't caught in tests, it's going to cause a production bug with a root-cause that's really hard to find. 

Anyway, can fix in a follow up, as getting this feature out the door is more important

tvc/src/commands/deploy/approve.rs
```diff
@@ -206,11 +214,13 @@ async fn build_inputs_non_interactive(
 
     let (manifest, fetched_manifest_id) = load_manifest(ctx, &args).await?;
 
+    let operator_id = args.operator_id.clone();
```
richardpringle: Same comment here

tvc/src/commands/deploy/approve.rs
```diff
@@ -252,18 +263,31 @@ async fn run_with_resolved_inputs(
         return Ok(());
     }
 
+    let operator =
+        resolve_operator(inputs.operator_seed_source, inputs.operator_id.as_deref()).await?;
+    let auth = if operator.is_hosted() {
+        Some(build_client().await?)
+    } else {
+        None
+    };
```
richardpringle, line 398: 👍

tvc/src/commands/deploy/approve.rs
```diff
@@ -252,18 +263,31 @@ async fn run_with_resolved_inputs(
...
     let approval = sign_and_output(
         ctx,
-        inputs.operator_seed_source,
+        &operator,
+        &OperatorCtx {
+            auth: auth.as_ref(),
+        },
```
richardpringle, line 403: Again, not for this PR

We should just make an operator trait and use dynamic dispatch here. The different operator types would have separate implementations for signing

tvc/src/commands/deploy/approve.rs
```diff
@@ -252,18 +263,31 @@ async fn run_with_resolved_inputs(
...
         inputs.approval_out.as_deref(),
         &inputs.manifest,
     )
     .await?;
 
     if let Some(target) = inputs.post_target {
+        let operator_id = operator
+            .operator_id()
```
richardpringle: nit (can do any nit in a follow up): 

```suggestion
        let operator_id = operator.id()
```

tvc/src/commands/operator/mod.rs
```diff
@@ -0,0 +1,3 @@
+//! Hosted operator commands.
```
richardpringle, line 1: Just a note, we should be using 2018 edition filenames `opertor.rs` and `operator/create.rs` instead of the `mod.rs` pattern. 

I can fix this all in one shot, so no need to change it here, but in the future, would be better to follow this pattern for all new modules

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
+//! Create a Turnkey-hosted TVC operator.
+
+use crate::client::build_client;
+use crate::config::turnkey::{Config, OperatorRecord, OperatorRecordKind};
+use crate::operator::{HostedOperatorSpec, create_hosted_operator, ensure_authenticated_org};
+use crate::output::{Message, StdCtx};
```
richardpringle: nit: please merge imports

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
...
+const DEFAULT_OPERATOR_NAME: &str = "tvc-operator";
+const DEFAULT_WALLET_NAME: &str = "tvc-wallet";
+
+/// Create one hosted TVC operator and save it to the active organization.
+#[derive(Debug, ClapArgs)]
+#[command(about, long_about = None)]
+#[command(group(
+    ArgGroup::new("wallet")
+        .args(["wallet_name", "wallet_id"])
+        .multiple(false)
```
richardpringle, line 28: 🥇

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
...
+    #[arg(long, env = "TVC_OPERATOR_NAME", default_value = DEFAULT_OPERATOR_NAME)]
+    pub name: String,
+
+    /// Name for a newly created wallet. Defaults to tvc-wallet.
+    #[arg(long, env = "TVC_OPERATOR_WALLET_NAME")]
+    pub wallet_name: Option<String>,
+
+    /// Existing wallet UUID in which to create the operator accounts.
+    #[arg(long, env = "TVC_OPERATOR_WALLET_ID")]
+    pub wallet_id: Option<String>,
```
richardpringle: This should be a UUID, shouldn't it?

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
...
+        .multiple(false)
+))]
+pub struct Args {
+    /// Human-readable operator name.
+    #[arg(long, env = "TVC_OPERATOR_NAME", default_value = DEFAULT_OPERATOR_NAME)]
+    pub name: String,
+
+    /// Name for a newly created wallet. Defaults to tvc-wallet.
+    #[arg(long, env = "TVC_OPERATOR_WALLET_NAME")]
+    pub wallet_name: Option<String>,
```
richardpringle: Why not do this:
```suggestion
    #[arg(long, env = "TVC_OPERATOR_WALLET_NAME", default_value = DEFAULT_WALLET_NAME)]
    pub wallet_name: String,
```
It simplifies the code a lot. You ignore the name if `args.wallet_id.is_some()` and rely on clap to make sure that you never have args constructed with invalid inputs.

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
...
+
+/// Create one hosted TVC operator and save it to the active organization.
+#[derive(Debug, ClapArgs)]
+#[command(about, long_about = None)]
+#[command(group(
+    ArgGroup::new("wallet")
+        .args(["wallet_name", "wallet_id"])
+        .multiple(false)
+))]
+pub struct Args {
```
richardpringle, line 30: I would remove `pub` visibility from all of the fields. You don't want the `Args` struct to be built without the `clap` validation

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
...
+
+pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<()> {
+    let mut config = Config::load().await?;
+    let (alias, configured_org_id) = config
+        .active_org_config()
+        .map(|(alias, org)| (alias.clone(), org.id.clone()))
+        .context("No active organization. Run `tvc login` first.")?;
+
+    let auth = build_client().await?;
+    ensure_authenticated_org(&auth.org_id, &configured_org_id)?;
```
richardpringle:
```suggestion
    let (alias, configured_org_id) = config
        .active_org_config()
        .map(|(alias, org)| (alias.clone(), org.id.as_str()))
        .context("No active organization. Run `tvc login` first.")?;

    let auth = build_client().await?;
    ensure_authenticated_org(&auth.org_id, configured_org_id)?;
```

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    ctx.shell().emit(&output)
+}
+
+fn hosted_operator_spec(args: Args) -> HostedOperatorSpec {
+    let wallet_name = match (&args.wallet_name, &args.wallet_id) {
+        (None, None) => Some(DEFAULT_WALLET_NAME.to_string()),
+        _ => args.wallet_name,
+    };
```
richardpringle: This pattern create more and more nested code. Instead of following the generic pattern:
```rust
let thing = create_thing(inputs)?;
```
It's much better to flatten out the callstack
```rust
let validated_inputs = validate_inputs(inputs)?;
let thing = create_thing(validated_inputs);
```
It decouples validation so that it can be tested separately. You can also use encapsulation here to minimize the number of possible code-paths. 

If `validate_inputs` returns `ValidatedInputs` whose fields aren't public, and if `create_thing` only takes `ValidatedInputs`, then we simplify our tests, make the code easier to read, and... we've made the code a lot more flexible. 

If we have two paths to create `thing` from a cli, for example, maybe one of them allows you to try again (interactive mode), the validation error that we want to return is different.

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
...
+
+fn add_operator_record(config: &mut Config, alias: &str, record: OperatorRecord) -> Result<()> {
+    config
+        .orgs
+        .get_mut(alias)
+        .with_context(|| format!("active organization '{alias}' disappeared from config"))?
+        .operators
+        .push(record);
+    Ok(())
+}
```
richardpringle: This is the type of code that's worth repeating instead of wrapping in a function. `alias` is going to be something that's validated early in the code path. Once that's validated, we should be able to pass around `&mut OrgConfig` where we need it and write the error context inline.

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
...
+
+fn output_from_record(record: &OperatorRecord) -> Result<OperatorCreated> {
+    let OperatorRecordKind::Hosted(hosted) = &record.kind else {
+        return Err(anyhow!("hosted operator creation returned a local record"));
+    };
```
richardpringle, line 150: Same comment about validating your inputs then passing them in. Generally speaking, the flatter the callstack, the easier the code is to change.

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
...
+    let composite_public_key = format!("{}{}", hosted.encrypt_public_key, hosted.sign_public_key);
+
+    Ok(OperatorCreated {
+        name: record.name.clone(),
+        operator_id: hosted.operator_id.clone(),
+        encrypt_public_key: hosted.encrypt_public_key.clone(),
+        sign_public_key: hosted.sign_public_key.clone(),
```
richardpringle: In general, cloning inside helper functions like this is an anti-pattern. You should be able to pass in the arguments you need as owned values. 

There are a lot of small functions that are really only called once or twice that force this cloning pattern. It's usually better to just inline the logic

tvc/src/commands/operator/create.rs
```diff
@@ -0,0 +1,254 @@
...
+        assert_eq!(
+            spec,
+            HostedOperatorSpec {
+                name: "tvc-operator".to_string(),
+                wallet_name: Some("tvc-wallet".to_string()),
+                wallet_id: None,
+                path: None,
+            }
+        );
+    }
```
richardpringle: If you use the `clap` defaults, you shouldn't have to test this.

tvc/src/operator.rs
```diff
@@ -0,0 +1,660 @@
...
+    /// and signing role components.
+    pub path: Option<String>,
+}
+
+#[derive(Debug)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+struct ValidatedHostedOperatorSpec {
+    name: String,
+    wallet_name: Option<String>,
+    wallet_id: Option<String>,
```
richardpringle:
```rust
enum NameOrId {
    Name(String), 
    Id(Uuid),
}
```
would help here

tvc/src/operator.rs
```diff
@@ -0,0 +1,660 @@
...
+            None => DEFAULT_HOSTED_OPERATOR_BASE_PATH.to_string(),
+        };
+        let (wallet_name, wallet_id) = match (spec.wallet_name, spec.wallet_id) {
+            (Some(name), None) => (Some(require_nonempty(name, "wallet name")?), None),
+            (None, Some(id)) => (None, Some(canonical_uuid(&id, "existing wallet ID")?)),
+            (None, None) => bail!("wallet name or existing wallet ID is required"),
+            (Some(_), Some(_)) => {
+                bail!("wallet name and existing wallet ID are mutually exclusive")
+            }
+        };
```
richardpringle: This simplifies to two cases if you force clap to validate the inputs and add the default wallet name. Then you just check for `(_, Some(wallet_id))` and `(name, None)`. 

That removes two of the error paths here (and the two test cases that you should have for these paths). I'm sure you can use clap to remove more, if not all

tvc/src/operator.rs
```diff
@@ -0,0 +1,660 @@
...
+pub async fn create_hosted_operator(
+    auth: &AuthenticatedClient,
+    spec: HostedOperatorSpec,
+) -> Result<OperatorRecord> {
+    let ValidatedHostedOperatorSpec {
+        name,
+        wallet_name,
+        wallet_id,
+        path,
+    } = spec.try_into()?;
```
richardpringle: Don't do this. Validate outside, then pass in the validated type

tvc/src/commands/deploy/approve.rs
daniilrrr: AI comment: 

```
[P2] Pass the auto-selected operator ID into signer resolution — /Users/daniil/tkhq/code/rust-sdk/tvc/src/commands/deploy/
    approve.rs:161-161
    When --operator-id is omitted, a saved ID may be selected for post_target, but ResolvedApproveInputs.operator_id remains None. In a
    mixed registry where that saved ID belongs to a hosted operator while the default kind is local, resolve_operator signs with the local
    key and then posts under the hosted ID, causing membership failure or an invalid/mislabeled approval. Use the resolved post-target ID
    when resolving the signer, or reject hosted saved IDs unless explicitly selected.
```

tvc/src/operator.rs
```diff
@@ -0,0 +1,660 @@
...
+        "authenticated organization ({authenticated_org_id}) does not match configured organization ({configured_org_id})"
+    );
+    Ok(())
+}
+
+pub(crate) async fn resolve_operator(
+    explicit: Option<LocalOperatorSeedSource>,
+    operator_id: Option<&str>,
+) -> Result<ResolvedOperator> {
+    let config = Config::load().await?;
```
daniilrrr: `config` isn't strictly necessary for explicit-seed. you can rework the order here 

AI comments below: 

```
[P2] Resolve explicit seeds before loading the registry — /Users/daniil/tkhq/code/rust-sdk/tvc/src/operator.rs:320-320
    An approval using --operator-seed or --operator-seed-path now unconditionally loads Config before handling the explicit credential.
    Consequently, formerly standalone flows such as deploy approve --operator-seed ... --skip-post fail when HOME is unset or the unrelated
    config file is malformed, even though no registry data is needed. Avoid loading config for explicit local signing unless an operator ID
    actually requires hosted-record disambiguation.
```

Something like 

```rust
pub(crate) async fn resolve_operator(
      explicit: Option<LocalOperatorSeedSource>,
      operator_id: Option<&str>,
  ) -> Result<ResolvedOperator> {
      if let Some(explicit) = explicit {
          if let Some(id) = operator_id {
              let config = Config::load().await?;
              ensure!(
                  find_hosted_operator(&config, id)?.is_none(),
                  "explicit local operator seed cannot be used with a hosted operator ID"
              );
          }

          return Ok(ResolvedOperator {
              name: None,
              operator_id: operator_id.map(str::to_string),
              organization_id: None,
              kind: ResolvedOperatorKind::Local(
                  resolve_local_operator(Some(explicit)).await?,
              ),
          });
      }

      let config = Config::load().await?;
      let hosted = match operator_id {
          Some(id) => find_hosted_operator(&config, id)?,
          None => None,
      };

      if let Some((organization_id, name, record)) = hosted {
          return Ok(ResolvedOperator {
              name: Some(name),
              operator_id: Some(record.operator_id.clone()),
              organization_id: Some(organization_id),
              kind: ResolvedOperatorKind::Hosted(record),
          });
      }

      // Existing configured-local resolution continues here...
```

tvc/src/commands/deploy/approve.rs
daniilrrr: so, immediately after `let post_target = ...` is constructed, do 

```rust
 let operator_id = operator_id.or_else(|| {
        post_target
            .as_ref()
            .map(|target| target.operator_id.clone())
    });
```

in both
`build_inputs_non_interactive()` and `build_inputs_interactive()`

tvc/src/operator.rs
daniilrrr: this is valid Rust, however the `ValidatedHostedOperatorSpec` is immediately consumed so it doesn't provide a ton of value imo as a new type. The only difference between the Validated and regular versions is that `path` becomes a `String` not an `Option<String>`. 

If you instead set the defaulting upstream in `create_hosted_operator()`, you could remove the new `ValidatedHostedOperatorSpec` type and instead do your validation in a `Validate impl` func which I've seen more often in other Rust code, and I think it reads a bit cleaner. 

```rust
fn hosted_operator_spec(args: Args) -> HostedOperatorSpec {
      let wallet_name = match (&args.wallet_name, &args.wallet_id) {
          (None, None) => Some(DEFAULT_WALLET_NAME.to_string()),
          _ => args.wallet_name,
      };

      let path = args
          .account_path
          .unwrap_or_else(|| DEFAULT_HOSTED_OPERATOR_BASE_PATH.to_string());

      HostedOperatorSpec {
          name: args.name,
          wallet_name,
          wallet_id: args.wallet_id,
          path,
      }
  }
```

then, your type can have `path: String`

```rust
pub struct HostedOperatorSpec {
      pub name: String,
      pub wallet_name: Option<String>,
      pub wallet_id: Option<String>,
      pub path: String,
  }
```

then,

```rust
impl HostedOperatorSpec {
      fn validate(self) -> Result<Self> {
          let Self {
              name,
              wallet_name,
              wallet_id,
              path,
          } = self;

          let name = require_nonempty(name, "operator name")?;
          let path = require_nonempty(path, "hosted operator account path")?;

          let (wallet_name, wallet_id) = match (wallet_name, wallet_id) {
              (Some(name), None) => (
                  Some(require_nonempty(name, "wallet name")?),
                  None,
              ),
              (None, Some(id)) => (
                  None,
                  Some(canonical_uuid(&id, "existing wallet ID")?),
              ),
              (None, None) => {
                  bail!("wallet name or existing wallet ID is required")
              }
              (Some(_), Some(_)) => {
                  bail!("wallet name and existing wallet ID are mutually exclusive")
              }
          };

          Ok(Self {
              name,
              wallet_name,
              wallet_id,
              path,
          })
      }
  }
```
and 

```rust
pub async fn create_hosted_operator(
      auth: &AuthenticatedClient,
      spec: HostedOperatorSpec,
  ) -> Result<OperatorRecord> {
      let HostedOperatorSpec {
          name,
          wallet_name,
          wallet_id,
          path,
      } = spec.validate()?;
...
...
```

makes unit testing specs slightly easier too by just calling `.validate()`

daniilrrr (commented):
some misc comments

daniilrrr:
based on the kind of mixed-config errors that AI review is flagging, I think we need to have some integration tests (would that be Vivosuite?) for TVC CLI asap to prevent introducing tricky bugs like the mixed registry issue that AI [flagged](https://github.com/tkhq/rust-sdk/pull/205#discussion_r3633151420). We should have some common user scenarios under test (multi-operator, multi-org, etc.) under test to ensure we don't break them with new features

tvc/src/operator.rs
```diff
@@ -0,0 +1,660 @@
...
+    })
+}
+
+fn require_nonempty(value: String, field: &str) -> Result<String> {
+    ensure!(!value.trim().is_empty(), "{field} must not be empty");
+    Ok(value)
+}
+
+fn canonical_uuid(value: &str, field: &str) -> Result<String> {
+    Uuid::parse_str(value)
```
richardpringle: Yeah, you can just set this type on the `clap::Args` field and you'll get this validation by default

richardpringle (changes requested):
Can you just fix up the type that implements `clap::Args`? It just prevents us from having to deal with a bunch of validation/validation error messages. 

Feel free to push back on this, the shape is fine, the code just needs cleaning up to make everything a little more maintainable

tvc/src/commands/operator/create.rs
daniilrrr: have https://linear.app/turnkey/issue/SYS-103/refactor-rust-sdk-cliinternal-ids-to-uuid-where-applicable to address this everywhere

tvc/src/operator.rs
daniilrrr: you could combine `create_hosted_operator()` and `validate()` into one func, and separate your inputs (`Args`) from your immutable `HostedOperatorSpec()` output, as Richard suggested [below](https://github.com/tkhq/rust-sdk/pull/205#discussion_r3632841515)

tvc/src/commands/deploy/approve.rs
annieke: for the local key case, local signing without posting is available so a signature can be generated offline (on an airgapped machine) and then the approval uploaded separately. honestly for the API path i should just remove this option instead of returning the raw signing output

tvc/src/commands/deploy/approve.rs
annieke: nice catch!

tvc/src/operator.rs
annieke: moved validation to be handled by clap via richard's suggestions

tvc/src/commands/deploy/approve.rs
annieke: specifying the uuid of the local operator still works; if it's unspecified it defaults to the org's local operator. 

however, this defaulting behavior now makes less sense and i'm somewhat inclined to remove defaulting / implicit operators since that pattern breaks with mixed dashboard usage.

tvc/src/commands/deploy/approve.rs
annieke: removed the ability to skip post for hosted signing

tvc/src/commands/deploy/approve.rs
annieke: fixed this via @daniilrrr 's suggestion above!

tvc/src/commands/deploy/approve.rs
annieke: https://linear.app/turnkey/issue/TVC-216/cli-make-operator-trait-and-use-dynamic-dispatch-for-signing

annieke: @richardpringle if you have more context for this feel free to add to the issue!

tvc/src/commands/operator/create.rs
annieke: converted this to a UUID

tvc/src/commands/operator/create.rs
annieke: clap is so powerful!

tvc/src/commands/deploy/approve.rs
richardpringle: 🤔 ticket?

tvc/src/operator.rs
annieke: configured the clap ArgGroup to not allow doubled wallet and name

tvc/src/commands/deploy/approve.rs
richardpringle: We should just have a `--offline` mode instead of saying things like "skip posting"

tvc/src/commands/operator/create.rs
annieke: fields are all private

tvc/src/commands/deploy/approve.rs
annieke: https://linear.app/turnkey/issue/TVC-217/cli-remove-uuid-defaulting-behavior-for-operators

tvc/src/commands/operator/create.rs
richardpringle: `clap` is the real reason to write clis in Rust

tvc/src/commands/deploy/approve.rs
```diff
@@ -1,32 +1,33 @@
...
+    config::turnkey::Config,
+    local_operator_key::LocalOperatorSeedSource,
+    operator::{OperatorCtx, ResolvedOperator, resolve_operator},
+    outcome::Outcome,
+    output::StdCtx,
+    pair::HexSeed,
+    prompts::{self, bail_required_in_non_interactive, stdin_can_prompt},
+    shell_print, shell_println,
+    util::{read_file_to_string, write_file},
+};
```
richardpringle, line 14: yaaaaas

tvc/src/commands/deploy/approve.rs
```diff
@@ -60,10 +61,11 @@ pub struct Args {
...
     pub manifest_id: Option<String>,
 
-    /// Turnkey operator ID (UUID) for the approving operator.
-    /// Required when posting approval to the API.
+    /// Turnkey operator UUID for the approving operator. When posting without
+    /// this flag, TVC selects from the last manifest's saved operator IDs. A
+    /// configured hosted operator is selected by its UUID.
     #[arg(long, env = "TVC_OPERATOR_ID")]
-    pub operator_id: Option<String>,
+    pub operator_id: Option<Uuid>,
```
richardpringle, line 68: 🚀

Cargo.lock
```diff
@@ -4296,6 +4296,9 @@ name = "uuid"
 version = "1.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ced87ca4be083373936a67f8de945faa23b6b42384bd5b64434850802c6dccd0"
+dependencies = [
+ "serde",
+]
```
richardpringle, line 4301: 👍

Cargo.toml
```diff
@@ -27,7 +27,7 @@ qos_p256 = { version = "0.12.1", default-features = false }
 base64 = { version = "0.22.0", default-features = false, features = ["std"] }
 bs58 = { version = "0.5.0", features = ["std", "check"], default-features = false }
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
-uuid = { version = "1.13.1", default-features = false }
+uuid = { version = "1.13.1", default-features = false, features = ["serde", "std"] }
```
richardpringle, line 30: 👍

richardpringle (approved):

PR #206, opened 2026-07-22
feat(tvc): build tvc binaries on release, optionally publish

daniilrrr:
## Summary

Build and publish `tvc` release binaries with a repository-owned GitHub Actions workflow whose build, packaging, and upload commands are visible directly in the workflow.

The workflow:

- Builds the four supported targets in parallel with a static matrix:
  - `x86_64-unknown-linux-musl`
  - `aarch64-unknown-linux-musl`
  - `aarch64-apple-darwin`
  - `x86_64-apple-darwin`
- Uses `cross` for Linux targets and `cargo` for macOS targets.
- Packages each binary as `tvc-<target>.tar.gz`, with `tvc` at the archive root.
- Generates a matching `tvc-<target>.tar.gz.sha256` checksum.
- Runs independently from `release-plz`; it does not parse or depend on release-plz action outputs.

## Validation and publishing

For every PR update targeting `main`, the workflow builds, packages, and checksums all four targets. It uploads each archive and checksum as a GitHub Actions artifact retained for three days. Manual runs perform the same non-publishing validation.

Release publishing is triggered separately after a successful `Release-plz` workflow on a `release-plz-*` branch. Once all four builds succeed, the publish job:

1. Reads the `tvc` version from `cargo metadata`.
2. Downloads the exact artifacts produced by the matrix jobs.
3. Verifies that the matching `tvc-v<version>` GitHub Release exists.
4. Uploads the archives and checksums to that release.

The release upload job is gated to successful `workflow_run` events whose triggering workflow is named `Release-plz`; PR and manual runs cannot publish release assets.

## Validation status

The validation for this PR is the current CI build: all four target jobs must compile, package, checksum, and upload their workflow artifacts successfully. The GitHub Release publishing path will be validated by the next real release.

macOS binaries are currently unsigned and unnotarized; code signing and Apple notarization remain follow-up work.

## Tracking

https://linear.app/turnkey/issue/TVC-224/build-tvc-cli-binaries-on-release


daniilrrr:
## Rationale: target & tooling choices

**Why these targets:**

| Target | Why |
|---|---|
| `x86_64-unknown-linux-musl` (static, `+crt-static`) | Matches the QOS build convention (`qos` `src/images/common/Containerfile` builds everything musl + crt-static). Fully static = zero glibc/runtime deps, so the binary drops straight into scratch/stagex containers — which directly serves the follow-up plan of having mono consume a published tvc binary instead of rebuilding it from source ([Slack thread](https://turnkeycrypto.slack.com/archives/C07TWDK573N/p1784749909595239)). |
| `aarch64-apple-darwin` | Apple Silicon — covers every Mac shipped since 2020, which is the overwhelming majority of dev laptops. |

**Deliberately excluded (for now):**
- `aarch64-unknown-linux-musl` — deferred pending team input (cc @r-n-o); no known ARM-Linux tvc users today. It's a one-line matrix addition if wanted.
- `x86_64-apple-darwin` — Intel Macs are legacy; add on demand.
- Windows — no known users; add on demand.
- Binary signing — follow-up work, not v1.

**Why `taiki-e/upload-rust-binary-action` (not cargo-dist):**
- It's the tool release-plz itself uses in its own `cd.yml` — the most "native" pairing with our existing release-plz setup.
- It only builds + attaches assets to the release release-plz already created; cargo-dist wants to own the whole release process (generated workflows, its own tag conventions), which conflicts with release-plz.
- cargo-dist's maintenance situation is shaky since axo.dev wound down (2025).

**Why a same-workflow job instead of `on: release`:**
Releases created with the default `GITHUB_TOKEN` (which the release-plz job uses) do not trigger downstream `on: release` workflows — GitHub suppresses cascading runs. Adding the job to the same workflow with `needs: release-plz-release` avoids introducing a PAT/App token, and inherits the existing `release` environment approval gate. The job only fires when release-plz actually created a `tvc-v*` release (checked via the action's `releases_created`/`releases` outputs).

daniilrrr:
## Testing approach: dry-run in this PR

To validate the binary build matrix **before** merging, this PR temporarily includes:

1. A widened trigger (`types: [ closed, opened, synchronize, reopened ]`) so the workflow runs on pushes to this PR
2. A branch-whitelisted `test-tvc-binaries` job that runs the exact same matrix (x86_64-linux-musl via cross, aarch64-apple-darwin) through `taiki-e/upload-rust-binary-action` with **`dry-run: true`** — it builds and archives the binaries but [uploads nothing](https://github.com/taiki-e/upload-rust-binary-action/blob/main/action.yml#L110). No token is used, no tags or releases are created.
3. A temporary guard on `release-plz-pr` so it doesn't run on PR sync events

**No releases are created by this PR.** The real `upload-tvc-binaries` job only runs after a release-plz version-bump PR (labeled `release`) merges and the `release` environment is approved — same gating as crate publishing today.

All temporary bits (trigger widening, test job, temp guard) will be removed before merge, leaving only the production `upload-tvc-binaries` job.

daniilrrr:
## Pre-merge testing: dry-run only

For this PR we can test the binary build matrix before merging by temporarily adding a branch-whitelisted PR test job that invokes `taiki-e/upload-rust-binary-action` with:

```yaml
dry-run: true
dry-run-intended: true
```

That path still builds `tvc` for the configured targets and creates the archives, but **does not upload release assets**. In dry-run mode the action also intentionally drops the token, so this test path should not be able to mutate GitHub Releases.

This means PR/synchronize runs are only validating:

- the Linux musl/static build (`x86_64-unknown-linux-musl`)
- the Apple Silicon build (`aarch64-apple-darwin`)
- archive creation via `upload-rust-binary-action`

It does **not** validate the final upload against a real `tvc-vX.Y.Z` release. That part only runs after `release-plz release` creates a real release and exposes the tag through its outputs.

The real release path remains gated behind the existing release job conditions: merged PR + `release` label + release environment approval. PR update runs should not create releases.

If a GitHub Release or release asset were accidentally created, we can delete the GitHub Release/assets and tag. The important caveat is crates.io: if release-plz publishes a crate version, that cannot be deleted from crates.io, only yanked. So for this dry-run testing, keep the release-plz publishing path unchanged/gated and only exercise the dry-run build job before merge.

daniilrrr:
## How this PR is being tested (now live)

The branch currently carries three **temporary** changes (all marked `TEMPORARY` in the workflow) so we can validate the release pipeline before merging:

1. **Widened trigger** — `pull_request` types include `opened/synchronize/reopened` so the workflow runs on pushes to this PR (production only needs `closed`).
2. **`TEST tvc binary (<target>)` jobs** — branch-whitelisted to `daniil/tvc-release-binaries`. They run the *exact same build path* as the production `upload-tvc-binaries` job — same action pins, same matrix (`x86_64-unknown-linux-musl` static via `cross`, `aarch64-apple-darwin` on macos-14), same `--locked` build and tar packaging — but with [`dry-run: true`](https://github.com/taiki-e/upload-rust-binary-action/blob/main/action.yml#L110): binaries are built and archived, **nothing is uploaded**, no token is used, no tags/releases are created.
3. **Temp guard on `release-plz-pr`** — restricted to `closed` events so the widened trigger doesn't run `release-plz release-pr` on every push.

**What green means:** both targets compile under the release configuration and produce valid `tvc-<target>.tar.gz` archives — i.e. the only untested part left is the release-plz output plumbing + actual asset upload, which can only run on a real release (worst case there: the crate release still succeeds, and we fix the upload job for the next release).

**What cannot happen from this PR:** no releases, no tags, no crates.io publishes. The production `upload-tvc-binaries` job requires a merged `release`-labeled PR *and* `release` environment approval, same as crate publishing today.

Before merge: all three temporary changes get reverted (grep `TEMPORARY`), leaving only the production `upload-tvc-binaries` job.

.github/workflows/release-plz.yml
```diff
@@ -7,7 +7,8 @@ permissions:
 
 on:
   pull_request:
-    types: [ closed ]
+    # TEMPORARY: testing dry-run builds in PR #206 — revert to [ closed ] before merge
+    types: [ closed, opened, synchronize, reopened ]
```
richardpringle: Just tagging this

.github/workflows/release-plz.yml
```diff
@@ -40,18 +44,130 @@ jobs:
...
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-musl
+            os: ubuntu-latest
+            build-tool: cross
+            rustflags: -C target-feature=+crt-static
+          - target: aarch64-apple-darwin
+            os: macos-14
+            build-tool: cargo
+            rustflags: ""
```
richardpringle: We should probably add `x86_64-apple-darwin`, `aarch64-unknown-linux-musl`, and maybe `x86_64-unknown-linux-gnu` though if the musl build just works on gnu platforms, we might be able to skip the latter

.github/workflows/release-plz.yml
```diff
@@ -40,18 +44,130 @@ jobs:
...
+          dry-run-intended: true
+
   # Create a PR with the new versions and changelog, preparing the next release.
   release-plz-pr:
     name: Release-plz PR
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    if: ${{ github.repository_owner == 'tkhq' }}
+    # TEMPORARY guard while trigger is widened — restore to repository_owner check only before merge
+    if: ${{ github.repository_owner == 'tkhq' && github.event.action == 'closed' }}
```
richardpringle: Just flagging this with a comment as well

richardpringle (commented):
LGTM! 🏆 

Had one comment to add other targets - not blocking though. 

I'm just marking this as "request changes" for your two comments for testing. I can approve once your happy with the tests! 

Copilot's comments actually seemed useful too, but shouldn't be blocking

richardpringle:
💩  I didn't leave the review above properly

.github/workflows/release-plz.yml
richardpringle: 2nd attempt to flag

.github/workflows/release-plz.yml
richardpringle: 2nd attempt to flag

richardpringle (dismissed):
Just for the "TEMPORARY" lines.

.github/workflows/release-plz.yml
daniilrrr: added `x86_64-apple-darwin` and `aarch64-unknown-linux-musl`

daniilrrr:
Addressed both of @Copilot's comments in the latest push:

**Brittle substring `if:` check** — moved tvc-release detection out of a raw JSON substring match. The `release-plz-release` job now parses the release list once and exposes a `tvc_tag` output; `upload-tvc-binaries` gates on `needs.release-plz-release.outputs.tvc_tag != ''` instead of `contains(..., '"package_name":"tvc"')`. No more sensitivity to JSON whitespace/key ordering.

**Multiline `jq` output** — the tag extraction now uses `... | map(select(.package_name == "tvc")) | first | .tag // empty`. `first` guarantees a single scalar even if multiple tvc entries ever appear, and `// empty` avoids writing a literal `null` to `GITHUB_OUTPUT`. The per-matrix extraction step was removed entirely, so the tag is parsed exactly once as the single source of truth.

Also added the requested targets (per @richardpringle): `aarch64-unknown-linux-musl` and `x86_64-apple-darwin`, bringing the matrix to 4 (both Linux-musl archs + both macOS archs). Skipping `x86_64-unknown-linux-gnu` since the static musl builds run fine on gnu systems. All 4 are covered by the dry-run test matrix.

daniilrrr:
the test jobs succeeded for the 4 platforms we want 
image

going to remove the widened guards and then merge

.github/workflows/release-plz.yml
```diff
@@ -40,11 +44,83 @@ jobs:
...
+        uses: taiki-e/upload-rust-binary-action@f0d45ae91ee7b8ee928de7a9d04d893a08bcbec6 # v1.30.2
+        with:
+          bin: tvc
+          package: tvc
+          target: ${{ matrix.target }}
+          build-tool: ${{ matrix.build-tool }}
+          locked: true
+          tar: all
+          ref: refs/tags/${{ needs.release-plz-release.outputs.tvc_tag }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```
r-n-o: These new actions are looking a bit sketchy for a security-sensitive repo, and I'd like to retain a bit more direct control over how we produce and upload binaries, because we want these signed eventually (not even super long-term, I think we should basically try to get that in place in the next few weeks?)

Having magical wrapper actions will make our lives harder I'm afraid. How about a simple set of jobs running alongside release-plz? That's the other part of my feedback here: if this is not done by release-plz, it shouldn't live in `release-plz.yml`. Let's create a new `tvc-binaries-release.yml` to separate responsibilities.

In my mind this looks like a simple matrix job, running `cargo build` on macos-latest or ubuntu-latest, then tar / sha256sum, then using `gh release upload` or equivalent to attach the built artifacts.

Another advantage with this separate job: you can produce the binaries for every PR, which means you get to test the correctness a lot more easily (maybe for PRs you simply upload to GH artifacts (the upload to GH releases can't happen for every PR!))

daniilrrr:
Builds succeed, however Mac builds require some additional code signing and Apple notarization steps to run unencumbered on Macs

image

From LLM:
___
Yes. The macOS binary is valid, but for normal external distribution it needs both:

  1. Code signing with a Developer ID Application certificate.
  2. Apple notarization so Gatekeeper can verify it was scanned by Apple.

  Apple requires Developer ID signing, hardened runtime, and a secure timestamp for distributed command-line tools. The notarized ticket for a standalone executable is
  available through Apple’s online Gatekeeper service, but cannot be stapled directly to the binary. Apple’s notarization requirements
  (https://developer.apple.com/documentation/security/notarizing-macos-software-before-distribution), custom notarization workflow
  (https://developer.apple.com/documentation/security/customizing-the-notarization-workflow).

  Our macOS jobs should therefore do:

  cargo build
    → codesign tvc
    → verify signature
    → place signed tvc in a temporary ZIP
    → submit ZIP with notarytool and wait
    → package the signed tvc in the existing tar.gz
    → sha256
    → publish

  The ZIP is only for Apple’s notarization service; we can retain the existing .tar.gz release format. Linux binaries are unaffected.

  To implement this in CI, we’ll need access to the organization’s:

  - Developer ID Application certificate and private key.
  - Apple notarization credentials, preferably an App Store Connect API key.
  - Apple team/issuer identifiers.

  For this test binary, you can verify the checksum and then use System Settings → Privacy & Security → Open Anyway if you trust this particular build. That is only a local
  test workaround, not acceptable release behavior. Apple’s Gatekeeper instructions (https://support.apple.com/en-euro/guide/mac-help/mchleab3a043/mac).

daniilrrr:
More LLM:
___
If Turnkey already has an Apple Developer account, this is a moderate CI change—probably several hours plus an iteration or two with Apple’s notary service. If it does not,
  the administrative setup will take longer than the code.

  The company setup is:

  - Enroll Turnkey as an organization in the Apple Developer Program.
  - Have an authorized employee act as Account Holder.
  - Provide the legal entity information and D‑U‑N‑S number.
  - Pay the current $99/year membership fee.
  - Create a Developer ID Application certificate.
  - Create notarization credentials.

  Apple documents these requirements in its organization enrollment guide (https://developer.apple.com/programs/enroll/) and Developer ID certificate guide
  (https://developer.apple.com/help/account/certificates/create-developer-id-certificates/).

  For GitHub Actions, we would then store the following in the protected release environment:

  - Developer ID certificate exported as a password-protected .p12.
  - Certificate password.
  - App Store Connect/notarization credentials.
  - Apple team, issuer, and key identifiers.

  The release flow becomes:

  build
    → smoke-test compatible binaries
    → import certificate into temporary keychain
    → codesign with hardened runtime and timestamp
    → verify signature
    → submit temporary ZIP to Apple with notarytool
    → wait for acceptance
    → create the existing tar.gz
    → checksum
    → publish**

daniilrrr:
Some more blogs on this 

https://dennisbabkin.com/blog/?t=how-to-get-certificate-code-sign-notarize-macos-binaries-outside-apple-app-store (older, from 2023)
https://tuist.dev/blog/2024/12/31/signing-macos-clis

.github/workflows/release-plz.yml
```diff
@@ -40,11 +42,39 @@ jobs:
...
+          RELEASES: ${{ steps.release-plz.outputs.releases }}
+        run: |
+          set -euo pipefail
+          # Robustly extract the single tvc tag: normalize to an array, select the
+          # tvc package, take the first match, and emit empty (not "null") if absent.
+          tag=$(jq -r 'if type == "array" then . else [.] end
+                       | map(select(.package_name == "tvc"))
+                       | first
+                       | .tag // empty' <<< "${RELEASES}")
+          echo "tag=${tag}" >> "${GITHUB_OUTPUT}"
```
r-n-o, line 64: My brain hurts looking at this 😵‍💫 

Default to "empty" if the tag can't be found doesn't make sense to me:
* tvc is released alongside other packages
* if we have a release without tvc in it I think something is seriously off and we should error instead of continuing with "tag" set to "empty".

More importantly: don't think this is required at all, see my other comment below! 👀

.github/workflows/release-plz.yml
```diff
@@ -40,11 +42,39 @@ jobs:
...
+    needs: release-plz-release
+    if: >
+      github.repository_owner == 'tkhq' &&
+      needs.release-plz-release.outputs.tvc_tag != ''
+    permissions:
+      contents: write
+    uses: ./.github/workflows/tvc-binaries-release.yml
+    with:
+      publish_release_assets: true
+      release_tag: ${{ needs.release-plz-release.outputs.tvc_tag }}
```
r-n-o, line 77: I don't think this should be in release-plz. Let's break this workflow logic up in a separate one, which triggers after release plz is done. Not as a separate step.

The logic to extract the right tag from release-plz output also feels unecessary: release-plz PRs bump Cargo.toml. So we can use normal rust to do this?

You can get version info with:
```
cargo metadata --no-deps --format-version 1 |  jq -r '.packages[] | select(.name == "tvc") | .version'
```

No need to hook into release plz's output.

Would look something like this:
```yaml
name: Publish TVC binaries

on:
  pull_request:
    types: [ opened, synchronize, reopened]
  workflow_run:
    workflows: ["Release-plz release"]
    types: [completed]
    branches: [main]

jobs:
  publish:
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v6
        with:
          ref: ${{ github.event.workflow_run.head_sha }}

      - name: Read tvc version
        id: version
        run: |
          version=$(cargo metadata \
            --no-deps \
            --format-version 1 |
            jq -r '.packages[] | select(.name == "tvc") | .version')

          echo "version=$version" >> "$GITHUB_OUTPUT"

      - name: Build and publish TVC binaries
        env:
          VERSION: ${{ steps.version.outputs.version }}
        run: |
          echo "Publishing tvc $VERSION"
          # ...
```

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,247 @@
...
+        description: Optional comma-separated targets; empty builds all supported targets
+        required: false
+        type: string
+        default: ""
+
+concurrency:
+  # Give each release tag and manual run its own group, while all ordinary post-merge validations share the "main" group.
+  group: tvc-binaries-${{ inputs.release_tag || (github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'release') && 'main') || github.run_id }}
+  # A newer ordinary merge supersedes an older main-branch validation. Never cancel release builds, manual runs, or release-labeled merge handling.
+  cancel-in-progress: ${{ github.event_name == 'pull_request' && inputs.release_tag == '' && !contains(github.event.pull_request.labels.*.name, 'release') }}
```
r-n-o, line 41: This seems very complex and hard to reason about. What happens if you just remove it?

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,247 @@
+name: TVC binaries
+
+on:
+  pull_request:
+    # TEMPORARY: opened/synchronize/reopened let PR #206 validate all four
+    # archives before merge. Remove them after the downloadable artifacts have
+    # been inspected, leaving only `closed` for post-merge validation.
+    types: [ opened, synchronize, reopened, closed ]
```
r-n-o, line 8: don't think closed or reopened is needed. We just need PR builds and release builds I think? I think the default `on: pull_request:` (without specifying "types") is good. We don't need to build on closed pull requests!

> By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened

(https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request)

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,247 @@
...
+#    types: [ closed ]
+    branches: [ main ]
+  workflow_dispatch:
+    inputs:
+      targets:
+        description: Optional comma-separated targets; empty builds all supported targets
+        required: false
+        type: string
+        default: ""
```
r-n-o, line 17: What's this for? To mock releases without a true release I imagine?

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,247 @@
...
+        github.event_name != 'pull_request' ||
+        (
+          github.event_name == 'pull_request' &&
+          github.event.action != 'closed'
+        ) ||
+        (
+          github.event.pull_request.merged == true &&
+          !contains(github.event.pull_request.labels.*.name, 'release')
+        )
+      )
```
r-n-o, line 62: I don't really understand why this PR is a special case. I'd hope we can still produce binaries on PR builds to test?

(testing the release behavior will be a bit more challenging though)

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,247 @@
...
+              fi
+              if [[ -n "${selected[${target}]+present}" ]]; then
+                echo "Duplicate target: ${target}" >&2
+                exit 1
+              fi
+              selected["${target}"]=true
+              targets+=("${target}")
+            done
+          fi
+
```
r-n-o, line 125: I'm very confused. We're dynamically generating targets based on a static list it sounds like? A bunch of stuff can go away and get simplified with a simple github matrix strategy. Something like this:

```yaml
jobs:
  build:
    name: Build ${{ matrix.target }}
    strategy:
      fail-fast: false
      matrix:
        include:
          - target: x86_64-unknown-linux-musl
            os: ubuntu-latest
            build_tool: cross
            rustflags: -C target-feature=+crt-static

          - target: aarch64-unknown-linux-musl
            os: ubuntu-latest
            build_tool: cross
            rustflags: -C target-feature=+crt-static

          - target: aarch64-apple-darwin
            os: macos-14
            build_tool: cargo
            rustflags: ""

          - target: x86_64-apple-darwin
            os: macos-14
            build_tool: cargo
            rustflags: ""

    runs-on: ${{ matrix.os }}

    steps:
      - uses: actions/checkout@v6

      - uses: dtolnay/rust-toolchain@stable
        with:
          targets: ${{ matrix.target }}

      - name: Install cross
        # ...

      - name: Build
        env:
          RUSTFLAGS: ${{ matrix.rustflags }}
        run: |
          ${{ matrix.build_tool }} build \
            --release \
            --package tvc \
            --target "${{ matrix.target }}"
```

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,247 @@
...
+        if: needs.prepare.outputs.publish_release_assets == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RELEASE_TAG: ${{ inputs.release_tag }}
+        run: |
+          set -euo pipefail
+          gh release upload "${RELEASE_TAG}" \
+            "tvc-${TARGET}.tar.gz" \
+            "tvc-${TARGET}.tar.gz.sha256" \
+            --clobber
```
r-n-o, line 247: This is the step you'd only do if the workflow is triggered from release (vs. PR). The rest can stay the same! 🎉

.github/workflows/tvc-binaries-release.yml
daniilrrr: removed

.github/workflows/tvc-binaries-release.yml
daniilrrr: going with the static matrix approach 👍

.github/workflows/tvc-binaries-release.yml
daniilrrr: yep, if you wanted a one-off platform build for testing. Or a manual run with workflow dispatch

.github/workflows/tvc-binaries-release.yml
daniilrrr: done

.github/workflows/release-plz.yml
daniilrrr: moved to `tvc-binaries-release.yml`

.github/workflows/tvc-binaries-release.yml
daniilrrr: you're right. I've removed the special casing and now the binaries will be built on each push to a PR and retained for `retention-days: 3`. This is distinct from retention for Github Releases which are retained indefinitely

daniilrrr: we can see the 4 build steps happening on this PR 👍

.github/workflows/release-plz.yml
daniilrrr: cool, now getting the version with 

```
version=$(cargo metadata --no-deps --format-version 1 |
    jq -r '.packages[] | select(.name == "tvc") | .version')
```
  then
```
  tvc-v${version}
```

.github/workflows/tvc-binaries-release.yml
daniilrrr: sounds good. I think to test this part, we would need to merge this PR in then trigger a new "test" release to see that the `release-plz` workflow triggers this step correctly. I can't think of a way to test this without merging the PR in, can you?

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,168 @@
...
+          # workflow_run's github.sha points at the default branch, so release
+          # builds must explicitly use the commit that release-plz ran against.
+          ref: ${{ github.event_name == 'workflow_run' && github.event.workflow_run.head_sha || github.sha }}
+          persist-credentials: false
+
+      - name: Install Rust target
+        run: |
+          set -euo pipefail
+          rustup show active-toolchain || rustup toolchain install
+          rustup target add "${TARGET}"
```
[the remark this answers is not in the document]
daniilrrr: (AGENT) `rustup toolchain install` accepts an omitted toolchain argument and installs or updates the active toolchain. This repository defines that active toolchain in `rust-toolchain.toml` with `channel = "1.95"`, so rustup can resolve it here without duplicating the version in the workflow. The current four-target matrix also passed this setup step.

emostov (dismissed):
Excited to get this merged and we can always iterate on top

.github/workflows/tvc-binaries-release.yml
r-n-o: yep! I think that's right. Short of cloning the repo and trying it out there, I think we'll have to try this out on the next real release and fix as needed. Shouldn't be a big deal. Worst case it doesn't work and we need to patch something. It's not like people rely on binaries being available today :p

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,168 @@
...
+
+on:
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
+  workflow_run:
+    workflows: [ "Release-plz" ]
+    types: [ completed ]
+    # Release-plz runs on the head branch of the merged release PR.
+    branches: [ "release-plz-*" ]
```
r-n-o, line 11: I'm not sure that's correct. Release plz does run release-plz branches but there's 2 parts to the workflow: update the release PR, and actually release.

I'm worried that this will run every time a release plz PR is updated, but I guess we can give it a shot?

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,168 @@
...
+
+jobs:
+  build:
+    name: Build tvc binary (${{ matrix.target }})
+    if: >
+      github.repository_owner == 'tkhq' &&
+      (
+        github.event_name != 'workflow_run' ||
+        github.event.workflow_run.conclusion == 'success'
+      )
```
r-n-o, line 29: I don't think that's quite what we want: this would prevent PRs from building TVC binaries (when a pull request is opened). I'd remove this `if` guard (maybe keep the repo owner guard though? I'm not sure what the consequences would be if this ran on forks so maybe better to leave that part in)

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,168 @@
...
+      RUSTFLAGS: ${{ matrix.rustflags }}
+      TARGET: ${{ matrix.target }}
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          # workflow_run's github.sha points at the default branch, so release
+          # builds must explicitly use the commit that release-plz ran against.
+          ref: ${{ github.event_name == 'workflow_run' && github.event.workflow_run.head_sha || github.sha }}
+          persist-credentials: false
```
r-n-o, line 63: what's this "persist-credentials: false" thing doing? Might be worth a comment

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,168 @@
...
+        run: |
+          set -euo pipefail
+          archive="tvc-${TARGET}.tar.gz"
+          binary_dir="target/${TARGET}/release"
+
+          tar -C "${binary_dir}" -czf "${archive}" tvc
+          if command -v sha256sum >/dev/null 2>&1; then
+            sha256sum "${archive}" > "${archive}.sha256"
+          else
+            shasum -a 256 "${archive}" > "${archive}.sha256"
```
r-n-o, line 114: That's also worth a comment because it's surprising: is it actually the case that not all platforms have `sha256sum` available? If so, worth naming which one is the "problem"

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,168 @@
...
+          "${BUILD_TOOL}" build \
+            --release \
+            --bin tvc \
+            --locked \
+            --package tvc \
+            --target "${TARGET}"
+
+      - name: Package and checksum tvc
+        env:
+          COPYFILE_DISABLE: "1"
```
r-n-o, line 103: this requires a comment. What is it doing?

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,168 @@
...
+          fi
+          cat "${archive}.sha256"
+
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: tvc-${{ matrix.target }}
+          path: |
+            tvc-${{ matrix.target }}.tar.gz
+            tvc-${{ matrix.target }}.tar.gz.sha256
```
r-n-o, line 124: wondering how this will work: don't we want commit hashes in this? Otherwise the files will likely be overridden?

.github/workflows/tvc-binaries-release.yml
```diff
@@ -0,0 +1,168 @@
...
+            exit 1
+          fi
+          echo "tag=tvc-v${version}" >> "${GITHUB_OUTPUT}"
+
+      - name: Download build artifacts
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
+        with:
+          pattern: tvc-*
+          path: artifacts
+          merge-multiple: true
```
r-n-o, line 184: not familiar with this. What does "merge-multiple" do?

.github/workflows/tvc-binaries-release.yml
daniilrrr: it's good practice to not leave the access token on the runner if unnecessary
https://release-plz.dev/docs/github/persist-credentials

will add a comment

.github/workflows/tvc-binaries-release.yml
daniilrrr: this is a flag to prevent `tar` on macOS from adding unnecessary metadata like `._tvc`. Not strictly necessary 

https://superuser.com/questions/259703/get-mac-tar-to-stop-putting-filenames-in-tar-archives

I checked by downloading the artifact from here https://github.com/tkhq/rust-sdk/actions/runs/32908928851/job/97998914384?pr=206, unzipping (is that the verb for tars lol?) the `.tar.gz` file, and only a `tvc` executable emerged

I'll add a comment unless you want me to remove it

.github/workflows/tvc-binaries-release.yml
daniilrrr: according to LLM, macOS is the one that provides `shasum -a` instead. This conditional is to avoid having to download CoreUtils https://unix.stackexchange.com/a/744422

I will add a comment

.github/workflows/tvc-binaries-release.yml
daniilrrr: according to LLM, this will be fine and the `upload-artifact` action now mandates immutable artifacts, and each workflow will create unique artifacts despite the same name. 

https://github.com/actions/upload-artifact/blob/main/README.md#not-uploading-to-the-same-artifact

once I push a new commit to this PR and trigger new artifacts under the same name, we will see what they look like 

We _**COULD**_ add commit hashes purely for artifact naming and easier matching between a given release commit <> resulting artifact. But it is not necessary for this Github action to work correctly.

.github/workflows/tvc-binaries-release.yml
daniilrrr: it puts them all in the same `/artifacts` dir instead of separate subdirs per target

Last paragraph here https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact

.github/workflows/tvc-binaries-release.yml
daniilrrr: `release-plz` workflow runs [only](https://github.com/tkhq/rust-sdk/blob/dbda0b402a48a87ee37ac3fe18240b0bcb216d5d/.github/workflows/release-plz.yml#L8-L11) when the PR is closed so this shouldn't be an issue

```
on:
  pull_request:
    types: [ closed ]
    branches: [ main ]
```

.github/workflows/tvc-binaries-release.yml
daniilrrr: that isn't true because this PR as-written triggered 4 successful CI binary build jobs - [example](https://github.com/tkhq/rust-sdk/actions/runs/32908928851/job/97998914384?pr=206)

at the top of this file the 3 valid triggers are `pull_request`, `workflow_dispatch` (manual run from Github UI or CLI), or `workflow_run` of `release-plz`. And this `if` is saying 

"Build unconditionally for PR/manual events, but for workflow_run, build only if Release-plz succeeded."

however, this is def confusing to read so i will make it more explicit and add a comment

.github/workflows/tvc-binaries-release.yml
daniilrrr: Commit 1 (current one) of this PR: 
- job: https://github.com/tkhq/rust-sdk/actions/runs/32908928851/job/97998914384?pr=206 
> Artifact tvc-aarch64-apple-darwin has been successfully uploaded! Final size is 7442762 bytes. Artifact ID is 9585902472
Artifact download URL: https://github.com/tkhq/rust-sdk/actions/runs/32908928851/artifacts/9585902472

unzips as `tvc-aarch64-apple-darwin.tar.gz`

daniilrrr: Commit 2: 
- job: https://github.com/tkhq/rust-sdk/actions/runs/32999806197/job/98278346901?pr=206
> Artifact tvc-aarch64-apple-darwin.zip successfully finalized. Artifact ID 9618192707
Artifact tvc-aarch64-apple-darwin has been successfully uploaded! Final size is 7442763 bytes. Artifact ID is 9618192707
Artifact download URL: https://github.com/tkhq/rust-sdk/actions/runs/32999806197/artifacts/9618192707

unzips with the same name `tvc-aarch64-apple-darwin.tar.gz`

.github/workflows/tvc-binaries-release.yml
r-n-o: yeah I think that's maybe correct, but there's also a chance that it triggers because the release-plz workflow triggers on PR closing (sure), but then the release-plz workflow updates the release PR, which has release-plz-* as a branch. So I wouldn't be surprised if this counted as a "workflow_run" even on a "release-plz-*" branch.

I'm down to merge as-is and adjust though.

.github/workflows/tvc-binaries-release.yml
r-n-o: Interesting! Thanks for the link! TIL `github.event.workflow_run.conclusion == 'success'` can be true for PR-triggered event? I guess that makes sense 👍

.github/workflows/tvc-binaries-release.yml
r-n-o: I just tried on my mac:
```
$ which sha256sum
/sbin/sha256sum
```

So this might only be valid for old versions or something?

r-n-o: If you can, try to standardize on `sha256sum`, I believe it's available everywhere. But if that's not the case then...yeah, comment it is :D

.github/workflows/tvc-binaries-release.yml
r-n-o: oh. The apple-double files! Yeah no that's a good thing to prevent for sure. An extra comment is perfect 👍

.github/workflows/tvc-binaries-release.yml
r-n-o: noice, thank you for looking at this 👏

.github/workflows/tvc-binaries-release.yml
daniilrrr: i will push up a version without the conditional and if it works then 👍

r-n-o (dismissed):

.github/workflows/tvc-binaries-release.yml
daniilrrr: addressed this edge case with an explicit `Verify release-plz-release job succeeded` step

.github/workflows/tvc-binaries-release.yml
daniilrrr: well dang, bot was right. jobs failed 

https://github.com/tkhq/rust-sdk/actions/runs/33002953228/job/98289212560?pr=206
`/Users/runner/work/_temp/4f50c2e0-16bf-4dc7-89c2-702f64ab0342.sh: line 6: sha256sum: command not found
` 

going to revert back

r-n-o: (to our bot overlords) sorry, I should know better. I shall not challenge Your Infinite Wisdom!

r-n-o (approved):

r-n-o (approved):

PR #208, opened 2026-07-23
feat(tvc): cryptographically validate manifest approvals

richardpringle:
- `deploy approve`: validates posted approvals against the manifest set and warns about any QOS would reject at boot (bad signature, non-member, duplicate).
- `deploy approve`: verifies its own freshly signed approval against the manifest hash before writing or posting.
- `deploy approve`: short-circuits when the operator already approved (matched by operator ID or member key) instead of re-posting.
- Quorum counts distinct members with cryptographically valid approvals (was a plain approval count).
- `deploy status`: validated-approvals section in human and JSON output — per-approval verdicts, valid count, threshold, quorum reached.
- Deterministic verdicts: approvals sort by `createdAt` (IDs are UUIDv4, unsortable); earliest valid approval per member wins, later ones flagged duplicate.
- Manifest sets validate once at parse (`ValidatedManifest`): every member key must parse; threshold 0 requires no approvals (enclave semantics).
- Boundary types: `--manifest-id` and operator/approval IDs parse as UUIDs; the approvals module uses typed errors, no `anyhow`.
- `tracing` instrumentation on approval posting and deployment fetch.

Closes ENG-4274.


tvc/src/commands/deploy/approve.rs
```diff
@@ -144,34 +147,57 @@ impl Display for ApproveOutcome {
...
 }
 
-#[derive(Default, Serialize)]
+impl Display for ApprovalOrPath {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::WrittenTo(path) => write!(f, "Approval written to: {path}"),
+            Self::Approval(approval) => f.write_str(
+                &serde_json::to_string_pretty(approval)
+                    .expect("serializing manifest approval should not fail"),
```
richardpringle: We should only have to write expect statements in very exceptional cases. 
```suggestion
                    .map_err(|_| fmt::Error)?,
```

tvc/src/commands/deploy/approve.rs
```diff
@@ -406,56 +445,94 @@ async fn run_with_resolved_inputs(
...
+    operator_id: &Uuid,
+    member_pub_key: &[u8],
+) -> Option<&'a TvcOperatorApproval> {
+    approvals.iter().find(|approval| {
+        approval.operator.as_ref().is_some_and(|operator| {
+            operator.id.parse::<Uuid>().ok().as_ref() == Some(operator_id)
+                || hex::decode(&operator.public_key).is_ok_and(|key| key == member_pub_key)
+        })
+    })
+}
```
richardpringle: This shouldn't in a separate function like this, it's easier to maintain inline

tvc/src/commands/deploy/approve.rs
```diff
@@ -406,56 +445,94 @@ async fn run_with_resolved_inputs(
...
+
+        ctx.shell().human().warn(format!(
+            "existing approval from {} is {}; QOS will reject it at enclave boot",
+            approval.operator_label(),
+            approval.verdict
+        ))?;
+    }
+
+    Ok(())
+}
```
richardpringle: Another function that should just be written inline

tvc/src/commands/deploy/approve.rs
```diff
@@ -252,22 +281,29 @@ struct ResolvedApproveInputs {
...
+    posted_approvals: Option<Vec<OperatorApproval>>,
+}
+impl Run for Args {
+    type Outcome = ApproveOutcome;
+
+    async fn run(mut self, ctx: &mut StdCtx) -> anyhow::Result<Self::Outcome> {
+        let operator_seed_source = LocalOperatorSeedSource::from_args(
+            self.operator_seed.take(),
+            self.operator_seed_path.take(),
+        )?;
```
richardpringle, line 377: We shouldn't be mutating `self` here. `build_inputs(_non)_interactive` should take a different type that doesn't have those fields. The fix should be in this shape:

```suggestion
    async fn run(mut self, ctx: &mut StdCtx) -> anyhow::Result<Self::Outcome> {
        let (operator_seed_source, args) = LocalOperatorSeedSource::from_args(self)?;
```
Then `args` can be passed into the branches of the if-statement below

tvc/src/commands/deploy/approve.rs
```diff
@@ -384,7 +422,26 @@ async fn run_with_resolved_inputs(
...
 ) -> anyhow::Result<ApproveOutcome> {
     if inputs.dry_run {
-        return Ok(ApproveOutcome::DryRun(ApprovalDryRun {}));
+        return Ok(ApproveOutcome::DryRun(ApprovalDryRun));
+    }
+
+    let validation = match inputs.posted_approvals {
+        Some(approvals) => Some(inputs.manifest.validate_approvals(approvals)?),
+        None => None,
+    };
```
richardpringle: Use `map` and `transpose` here instead of a match

tvc/src/commands/deploy/approve.rs
```diff
@@ -384,7 +422,26 @@ async fn run_with_resolved_inputs(
...
+            if validated.verdict == ApprovalVerdict::Valid {
+                continue;
+            }
+
+            ctx.shell().human().warn(format!(
+                "existing approval from {} is {}; QOS will reject it at enclave boot",
+                validated.operator_label(),
+                validated.verdict
+            ))?;
+        }
```
richardpringle: Call `validation.approvals.iter().filter(...).try_for_each(...)` instead of a for loop with `continue` and stuffing that `?` operator in there. 

Should we even fail on warning here? I don't think I would. Let's get rid of that too, maybe we an put errors in debug.

richardpringle: Further, why shouldn't we want to warn as a part of the json logs?

tvc/src/commands/deploy/approve.rs
```diff
@@ -406,56 +463,64 @@ async fn run_with_resolved_inputs(
...
         Some(target) => {
             let operator_id = operator
                 .id()
                 .context("resolved operator ID required to post approval")?;
+
+            // An approval already posted by this operator, matched by
+            // operator ID or by the manifest set member public key.
+            let existing = validation.as_ref().and_then(|validation| {
+                validation.approvals.iter().find(|validated| {
+                    validated.approval.operator_id.parse::<Uuid>().ok().as_ref()
```
richardpringle: Why are we parsing the uuid here? It should already be parsed. I'm guessing that's because this value can be an empty string? This should be represented with an enum instead then or an `Option<Uuid>`

tvc/src/commands/deploy/approve.rs
```diff
@@ -406,56 +463,64 @@ async fn run_with_resolved_inputs(
...
+                        == Some(&operator_id)
+                        || validated.approval.public_key.to_bytes() == approval.member.pub_key
+                })
+            });
+
+            if let Some(existing) = existing {
+                return Ok(ApproveOutcome::AlreadyPosted(ApprovalAlreadyPosted {
+                    operator_id: operator_id.to_string(),
+                    approval_id: existing.approval.id.clone(),
```
richardpringle: We should not need to clone on a return..

tvc/src/commands/deploy/approve.rs
```diff
@@ -406,56 +463,64 @@ async fn run_with_resolved_inputs(
...
+                }));
+            }
+
             let plan = PostApprovalPlan {
                 manifest_id: target.manifest_id.as_str(),
                 operator_id: &operator_id,
                 deploy_id: target.deploy_id.as_deref(),
             };
-            let posted = post_approval_to_api(ctx, plan, &approval).await?;
+            let posted = post_approval_to_api(ctx, plan, &approval, &inputs.manifest).await?;
```
richardpringle, line 535: side thought, has this been instrumented?

tvc/src/commands/deploy/approve.rs
```diff
@@ -406,56 +463,64 @@ async fn run_with_resolved_inputs(
...
 }
 
 async fn load_manifest(
     ctx: &mut StdCtx,
     args: &Args,
-) -> anyhow::Result<(VersionedManifest, Option<String>)> {
+) -> anyhow::Result<(
+    VersionedManifest,
+    Option<String>,
+    Option<Vec<OperatorApproval>>,
```
richardpringle: It appears that when the operator-id returns there's also ALWAYS approvals. This should be `Option<(String, Vec<OperatorApprovals>)>`

tvc/src/commands/deploy/status.rs
```diff
@@ -59,18 +61,34 @@ pub async fn run(_ctx: &mut StdCtx, args: Args) -> anyhow::Result<Outcome> {
         organization_id: _,
         manifest_set: _,
         share_set: _,
-        manifest_approvals: _,
+        manifest_approvals,
     } = deployment;
 
-    let manifest =
+    let tvc_manifest =
```
richardpringle: why the prefix?

tvc/src/commands/deploy/status.rs
```diff
@@ -59,18 +61,34 @@ pub async fn run(_ctx: &mut StdCtx, args: Args) -> anyhow::Result<Outcome> {
...
+    let manifest_approvals = match VersionedManifest::try_from_slice_compat(&tvc_manifest.manifest)
+    {
+        Ok(manifest) => Some(manifest.validate_approvals(manifest_approvals)?),
+        Err(error) => {
+            ctx.shell().human().warn(format!(
+                "failed to parse manifest; cannot validate approvals: {error}"
+            ))?;
+            None
+        }
+    };
```
richardpringle: One step at a time. And we should probably not fail on not being able to warn

```suggestion
    let manifest = VersionedManifest::try_from_slice_compat(&maniest.manifest).map_err(|err| {
            let write_err = ctx.shell().human().warn(format!(
                "failed to parse manifest; cannot validate approvals: {error}"
            ));
            
            debug!(/* do something with the write_err here */);
            
            err
    }).ok();
    
    let manifest_approvlas = manifest
        .map(|manifest| manifest.validate_approvals(manifest_approvals))
        .transpose()?;
```

tvc/src/cli.rs
```diff
@@ -159,8 +159,12 @@ fn handle_parse_error(error: clap::Error) -> ExitCode {
                     .to_string();
                 let error_message = ErrorMessage::usage_error(message);
                 let mut stdout = std::io::stdout();
+
+                // it shouldn't be possible for serialization to fail here.
+                let msg = serde_json::to_string(&error_message).unwrap_or_else(|e| e.to_string());
```
richardpringle, line 164: `error_message` should have a `to_json` method that guarantees no failure path and is well tested

tvc/src/operator.rs
```diff
@@ -361,7 +361,27 @@ impl ResolvedOperator {
...
+        // Membership is already proven — `member` came out of the manifest
+        // set — so verifying the fresh signature is the only remaining check.
+        qos_p256::P256Public::from_bytes(&approval.member.pub_key)
+            .map_err(|error| {
+                anyhow!(
+                    "manifest set member key for {} is not a valid P256 key: {error:?}",
+                    approval.member.alias
+                )
+            })?
+            .verify(&manifest.manifest_hash(), &approval.signature)
```
richardpringle: Why is this not using the trait extension?

tvc/src/output.rs
```diff
@@ -392,22 +371,19 @@ mod tests {
     }
 
     #[derive(Serialize)]
+    #[serde(tag = "reason", rename = "test_message")]
```
richardpringle: this is pointless

Cargo.lock
```diff
@@ -3652,6 +3652,27 @@ version = "0.11.1"
...
+name = "strum_macros"
+version = "0.27.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7695ce3845ea4b33927c055a39dc438a45b059f7c1b3d91d38d10355fb8cbca7"
+dependencies = [
+ "heck",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.100",
+]
```
richardpringle, line 3674: Same version we use in mono

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+use std::{collections::HashSet, fmt};
+use turnkey_client::generated::external::data::v1::{TvcOperator, TvcOperatorApproval};
+
+/// A posted manifest approval, parsed at the API boundary: the operator is
+/// present and its public key is a valid P256 key. Existence of this type is
+/// proof of well-formedness; it says nothing about signature validity or
+/// manifest set membership — that is [`ManifestExt::validate_approvals`].
+pub(crate) struct OperatorApproval {
+    pub id: String,
+    pub operator_id: String,
```
richardpringle: Uuids?

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+    pub operator_name: String,
+    pub public_key: P256Public,
+    pub signature: Vec<u8>,
```
richardpringle: Does this type live long enough to warrant owning the signature?

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+
+impl ValidatedApproval {
+    /// "name (id)" for human-facing approval lines.
+    pub(crate) fn operator_label(&self) -> String {
+        format!(
+            "{} ({})",
+            self.approval.operator_name, self.approval.operator_id
+        )
+    }
+}
```
richardpringle: This should be a `Display` implementation

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+
+impl Serialize for ValidatedApproval {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let mut state = serializer.serialize_struct("ValidatedApproval", 3)?;
+        state.serialize_field("operatorId", &self.approval.operator_id)?;
+        state.serialize_field("operatorName", &self.approval.operator_name)?;
+        state.serialize_field("verdict", &self.verdict)?;
+        state.end()
+    }
+}
```
richardpringle: This is really dumb... Why not just have a `Serialize` implementation on `OperatorApproval`, then use `serde(flatten)` here. Saves a lot of code surface for mistakes... (and would have saved output tokens)

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+
+pub(crate) struct ApprovalValidation {
+    pub approvals: Vec<ValidatedApproval>,
+    pub threshold: u32,
```
richardpringle: This should just be a `usize` or possibly a `NonZeroUsize`

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+    /// Number of distinct manifest set members with a valid approval.
+    pub(crate) fn valid_count(&self) -> usize {
+        self.approvals
+            .iter()
+            .filter(|validated| validated.verdict == ApprovalVerdict::Valid)
+            .count()
+    }
+
+    pub(crate) fn quorum_reached(&self) -> bool {
+        self.threshold > 0 && self.valid_count() >= self.threshold as usize
```
richardpringle: Is quorum never reached if the threshold is 0? That doesn't make sense to me, seems like a threshold of 0 just wouldn't require approvals. We should check the backend to see what the rules are. If 0 is invalid, we should leverage the type system here.

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+}
+
+impl Serialize for ApprovalValidation {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let mut state = serializer.serialize_struct("ApprovalValidation", 4)?;
+        state.serialize_field("approvals", &self.approvals)?;
+        state.serialize_field("validCount", &self.valid_count())?;
+        state.serialize_field("threshold", &self.threshold)?;
+        state.serialize_field("quorumReached", &self.quorum_reached())?;
+        state.end()
```
richardpringle:
```suggestion
    #[derive(Serialize)]
    #[serde(rename_all = "camelCase")]
    struct SerializeApprovalValidation<'a> {
        #[serde(flatten)]
        validation: &'a ApprovalValidation,
        valid_count: usize,
        quorum_reached: bool,
    }
    
    SerializeApprovalValidation {
        count: self.valid_count(),
        quorum_reached: self.quorum_reached(),
        validation: self,
    }.serialize(serializer)
```

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+            .iter()
+            .map(|member| {
+                P256Public::from_bytes(&member.pub_key)
+                    .map(|_| member.pub_key.as_slice())
+                    .map_err(|error| {
+                        anyhow!(
+                            "manifest set contains an invalid public key for member {}: {error:?}",
+                            member.alias
+                        )
+                    })
```
richardpringle, line 259: Instead of doing this here, can we make a type that wraps the manifest and validate the members every time? Part of the "parse, don't validate" mindset

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+                        .is_err() =>
+                    {
+                        ApprovalVerdict::InvalidSignature
+                    }
+                    _ if counted.contains(&key_bytes) => ApprovalVerdict::Duplicate,
+                    _ => {
+                        counted.insert(key_bytes);
+                        ApprovalVerdict::Valid
+                    }
+                };
```
richardpringle: This would be a lot simpler if approvals were just deduped with a map of verdicts `HashMap<&[u8], Verdict>` and a `HashSet<Approval>` of duplicates. Everything should probably be collected into a `BTreeMap` sort the results by timestamp so that they're always displayed in order. We shouldn't be marking the first approval with a specific pubkey as invalid and a later one as valid

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+        pair: &P256Pair,
+        manifest: &VersionedManifest,
+    ) -> OperatorApproval {
+        OperatorApproval {
+            id: format!("approval-{operator_id}"),
+            operator_id: operator_id.to_string(),
+            operator_name: format!("operator-{operator_id}"),
+            public_key: pair.public_key(),
+            signature: pair.sign(&manifest.manifest_hash()).unwrap(),
+        }
```
richardpringle, line 460: I guess we don't have a timestamp here... we should probably have a stamp?

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+    }
+
+    fn api_approval(
+        operator_id: &str,
+        public_key: String,
+        signature: Vec<u8>,
+    ) -> TvcOperatorApproval {
+        TvcOperatorApproval {
+            id: format!("approval-{operator_id}"),
```
richardpringle: What type of UUIDs are these? Are they sortable?

richardpringle (commented):
😞

tvc/src/approvals.rs
```diff
@@ -0,0 +1,505 @@
...
+impl fmt::Display for ApprovalVerdict {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(match self {
+            Self::Valid => "valid",
+            Self::InvalidSignature => "invalid signature",
+            Self::NotInManifestSet => "not in manifest set",
+            Self::Duplicate => "duplicate",
+        })
+    }
+}
```
richardpringle: use `displaydoc@0.2.5` for this

tvc/src/commands/deploy/approve.rs
```diff
@@ -106,10 +109,62 @@ pub struct Args {
     pub skip_post: bool,
 }
 
+impl Run for Args {
```
richardpringle, line 112: If this makes the review too difficult, let me know. I started to find this file a little hard to navigate, there was a lot of clutter before the actual `Run` implementation, so I moved it. I can move it back if that makes the review easier and open a separate PR cleaning up the file making it easier to navigate

tvc/src/commands/deploy/approve.rs
```diff
@@ -191,21 +265,17 @@ pub struct ApprovalPosted {
...
+impl fmt::Display for ApprovalPosted {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        f.write_str(&approval_payload_human_message(
-            &self.approval,
-            &self.written_to,
-        ))?;
         write!(
             f,
-            r#"
+            r#"{}
```
richardpringle, line 272: Same output, logic moved to `impl Display for ApprovalOrPath`

tvc/src/commands/deploy/approve.rs
```diff
@@ -214,208 +284,212 @@ Operator ID: {}"#,
...
+/// {0}
+pub struct ApprovalGenerated(ApprovalOrPath);
 
-#[derive(Default, Serialize)]
+#[derive(Display, Default, Serialize)]
 #[serde(rename_all = "camelCase")]
+/// Operator {operator_id} has already approved this manifest (approval ID: {approval_id}). Nothing to post.
 pub struct ApprovalAlreadyPosted {
     operator_id: String,
     approval_id: String,
```
richardpringle, line 305: Seems these should both be UUIDs, I'm going to leave it for now though, the diff is big enough already

tvc/src/commands/deploy/approve.rs
```diff
@@ -214,210 +284,210 @@ Operator ID: {}"#,
...
 
+    let validation = inputs.manifest.validate_approvals(inputs.posted_approvals);
+
+    validation
+        .approvals
+        .iter()
+        .filter(|validated| validated.verdict != ApprovalVerdict::Valid)
+        .for_each(|validated| {
+            let message = format!(
+                "existing approval from {} is {}; QOS will reject it at enclave boot",
```
richardpringle, line 482: too much information?

tvc/src/approvals.rs
```diff
@@ -0,0 +1,687 @@
+//! Cryptographic validation of manifest approvals against a manifest set.
+//!
+//! API approvals are parsed once at the fetch boundary into
+//! [`OperatorApproval`] — construction fails on a missing operator or a
+//! malformed public key, so validation only ever sees well-formed approvals.
+//! [`ValidatedManifest`] then classifies every approval instead of failing
+//! fast, so callers can report problems before they surface at enclave boot:
+//! an approval counts toward the threshold iff it is signed over the manifest
```
richardpringle:
```suggestion
//! an approval counts toward the threshold if it is signed over the manifest
```

tvc/src/approvals.rs
```diff
@@ -0,0 +1,710 @@
...
+        approvals.sort_by_key(|approval| {
+            let stamp = approval.created_at.as_ref().and_then(|created_at| {
+                Some((
+                    created_at.seconds.parse::<i64>().ok()?,
+                    created_at.nanos.parse::<u32>().ok()?,
+                ))
+            });
+
+            (stamp.is_none(), stamp, approval.id)
+        });
```
richardpringle, line 299: I don't know that this is the right place to do this. We should probably rely on the sort order of the API

tvc/src/commands/deploy/approve.rs
```diff
@@ -135,54 +192,71 @@ impl From<ApproveOutcome> for Outcome {
...
+/// The approval payload as an outcome reports it: inline when it went to
+/// stdout, the file path when `--approval-out` wrote it.
+#[derive(Display, Serialize)]
+#[serde(rename_all = "camelCase")]
+enum ApprovalOrPath {
+    /// Approval written to: {0}
+    WrittenTo(PathBuf),
+    /// {0}
+    Approval(JsonPretty<Approval>),
 }
```
annieke, line 210: not a request for change and more of a convention question: is this type of enum common? in my mind, an approval and path are so completely separate concepts, that this name confused me during my first scan. 

maybe `ApprovalOutput` as an enum name would make this less confusing

tvc/src/commands/deploy/approve.rs
annieke: i found this pretty easy to read

tvc/src/commands/deploy/approve.rs
annieke: Hm i see there's also `ApproveOutcome`, but seems separate enough so up to you!

tvc/src/commands/deploy/approve.rs
```diff
@@ -214,210 +284,210 @@ Operator ID: {}"#,
...
-                    _ => bail!(
+        match saved_ids.len() {
+                0 => bail!(
+                    "--operator-id is required to post approval to API. \
+             No saved operator IDs found. \
+             Use --skip-post to only generate the approval locally."
+                ),
+                1 => saved_ids[0],
+                _ => select(saved_ids)?.ok_or_else(|| {
+                    anyhow!(
```
annieke, line 451: not within scope of this PR but i feel tempted to fully remove this ID defaulting behavior, now that the config is equipped handle more than one operator per org. wonder what you think after interacting with this logic

tvc/src/commands/deploy/approve.rs
```diff
@@ -214,210 +284,210 @@ Operator ID: {}"#,
...
     ctx: &mut StdCtx,
-    inputs: ResolvedApproveInputs,
+    inputs: ResolvedApproveInputs<'_>,
 ) -> anyhow::Result<ApproveOutcome> {
     if inputs.dry_run {
-        return Ok(ApproveOutcome::DryRun(ApprovalDryRun {}));
+        return Ok(ApproveOutcome::DryRun(ApprovalDryRun));
     }
 
+    let validation = inputs.manifest.poop(inputs.posted_approvals);
```
annieke: 💩 lol

tvc/src/commands/deploy/approve.rs
```diff
@@ -214,210 +284,210 @@ Operator ID: {}"#,
...
+
+    validation
+        .approvals
+        .iter()
+        .filter(|validated| validated.verdict != ApprovalVerdict::Valid)
+        .for_each(|validated| {
+            let message = format!(
+                "existing approval from {} is {}; QOS will reject it at enclave boot",
```
annieke, line 482: i'm not against the original error, but the following might be more specific w/ expected behavior
```suggestion
                "existing approval from {} is {}; enclave will reject this approval and fail to start",
```

tvc/src/commands/deploy/approve.rs
```diff
@@ -437,56 +507,57 @@ async fn run_with_resolved_inputs(
...
             let operator_id = operator
                 .id()
                 .context("resolved operator ID required to post approval")?;
+
+            // An approval already posted by this operator, matched by
+            // operator ID or by the manifest set member public key.
+            let existing = validation.approvals.iter().find(|validated| {
+                validated.approval.operator_id == operator_id
+                    || validated.approval.public_key.to_bytes() == approval.member.pub_key
+            });
```
annieke, line 521: not within scope for this PR: do we have server-side checks for not persisting repeated approvals? i can imagine the edge case of this passing if we post another approval a split second earlier but it wasn't fetched at the beginning of the command

tvc/src/commands/deploy/approve.rs
```diff
@@ -437,56 +507,57 @@ async fn run_with_resolved_inputs(
...
+    args: &ArgsWithResolvedOperatorSeedSource,
+) -> anyhow::Result<(VersionedManifest, Option<(Uuid, Vec<OperatorApproval>)>)> {
     match (&args.manifest, &args.deploy_id) {
         (Some(path), _) => Ok((read_manifest_from_path(path).await?, None)),
         (_, Some(deploy_id)) => {
-            let (manifest, manifest_id) =
+            let (manifest, manifest_id, approvals) =
                 fetch_manifest_from_deploy(ctx, &deploy_id.to_string()).await?;
-            Ok((manifest, Some(manifest_id)))
+            Ok((manifest, Some((manifest_id, approvals))))
```
annieke, line 560: 🔥

tvc/src/approvals.rs
```diff
@@ -0,0 +1,705 @@
...
+    #[error("operator ID '{id}' is not a UUID")]
+    OperatorIdNotUuid { id: String, source: uuid::Error },
+    #[error("operator {operator_id} public key is not valid hex")]
+    PublicKeyNotHex {
+        operator_id: Uuid,
+        source: hex::FromHexError,
+    },
+    #[error("operator {operator_id} public key is not a valid P256 key: {error:?}")]
+    PublicKeyNotP256 { operator_id: Uuid, error: P256Error },
+}
```
annieke, line 71: this error handling reads nicely! loading into 🧠

tvc/src/approvals.rs
```diff
@@ -0,0 +1,705 @@
...
+    pub threshold: usize,
+}
+
+/// Quorum facts derived from a validation, computed once at packaging.
+#[derive(serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct ApprovalValidationMeta {
+    pub valid_count: usize,
+    pub quorum_reached: bool,
+}
```
annieke, line 169: what's the rationale for separating these? is it `Meta` can only be computed after submission and `ApprovalValidation` is before?

tvc/src/approvals.rs
```diff
@@ -0,0 +1,705 @@
...
+    fn deref(&self) -> &Self::Target {
+        self.manifest
+    }
+}
+
+impl ValidatedManifest<'_> {
+    /// Classify every posted approval against this manifest's set and hash.
+    /// Every approval gets a verdict instead of failing fast, so callers can
+    /// report problems before they surface at enclave boot.
+    pub(crate) fn poop(&self, mut approvals: Vec<OperatorApproval>) -> ApprovalValidation {
```
annieke: is 💩 an inside-joke type name?

annieke (dismissed):
no blocking issues found! one note that's not in scope of this PR:
- we generate a local share approval in re_encrypt_local_share https://github.com/tkhq/rust-sdk/blob/22acc9c359b309df55b11549050d1c36800276f8/tvc/src/commands/keys/re_encrypt_local_share.rs#L154-L163
- that approval is created during post_share
https://github.com/tkhq/rust-sdk/blob/22acc9c359b309df55b11549050d1c36800276f8/tvc/src/commands/deploy/post_share.rs#L72-L80

could be nice to have this parallel check in that flow, but it's less dire since the main hosted flow is in the server

tvc/src/commands/deploy/approve.rs
richardpringle: Yes this type of enum is common, but you're absolutely correct `ApprovalOutput` is the better name.

tvc/src/commands/deploy/approve.rs
richardpringle: I'll create a ticket. It's definitely fragile code

tvc/src/commands/deploy/approve.rs
richardpringle: Good question, I can create a ticket to investigate

tvc/src/approvals.rs
richardpringle: I can only take credit for pushing the usage of `thiserror` and enums

tvc/src/approvals.rs
richardpringle: Yeah. If you have ideas about better names, I'm all ears. It's not really "meta" per se... but kind of? It's actually only used to display to the end-user. So the idea is that the useful type is `ApprovalValidation`, then you call `with_meta()` to transform it for serialization/end-user display

Avi-D-coder (approved):
Confirmed No new dep only

tvc/src/commands/deploy/approve.rs
```diff
@@ -106,10 +109,62 @@ pub struct Args {
     pub skip_post: bool,
 }
 
+impl Run for Args {
+    type Outcome = ApproveOutcome;
+
+    async fn run(self, ctx: &mut StdCtx) -> anyhow::Result<Self::Outcome> {
+        let args = ArgsWithResolvedOperatorSeedSource::try_from(self)?;
```
daniilrrr, line 116: we could be a little clever here with `clap` ArgGroups and custom value parser funcs. Not sure how worth it that would be 

LLM:

```rust
 use clap::{ArgGroup, Args as ClapArgs};
  use std::{ffi::OsStr, path::PathBuf};

  fn parse_operator_seed(value: &str) -> Result<LocalOperatorSeedSource, String> {
      value
          .parse::<HexSeed>()
          .map(LocalOperatorSeedSource::Value)
          .map_err(|error| error.to_string())
  }

  fn parse_operator_seed_path(value: &OsStr) -> Result<LocalOperatorSeedSource, String> {
      if value.is_empty() {
          return Err("operator seed path cannot be empty".to_owned());
      }

      Ok(LocalOperatorSeedSource::Path(PathBuf::from(value)))
  }

  #[derive(Debug, ClapArgs)]
  #[command(group(
      ArgGroup::new("operator-seed-source")
          .args(["operator_seed", "operator_seed_path"])
          .multiple(false)
  ))]
  pub struct Args {
      // Other fields...

      /// Hex-encoded 32-byte master seed for the operator key.
      #[arg(
          long,
          value_name = "HEX_SEED",
          env = "TVC_OPERATOR_SEED",
          value_parser = parse_operator_seed,
      )]
      operator_seed: Option<LocalOperatorSeedSource>,

      /// Path containing the hex-encoded operator seed.
      #[arg(
          long,
          value_name = "PATH",
          env = "TVC_OPERATOR_SEED_PATH",
          value_parser = parse_operator_seed_path,
      )]
      operator_seed_path: Option<LocalOperatorSeedSource>,
  }
```

tvc/src/commands/deploy/approve.rs
```diff
@@ -214,210 +284,210 @@ Operator ID: {}"#,
...
-    // required but the caller has no TTY to answer them.
-    if do_prompt_user && !stdin_can_prompt() {
-        bail_required_in_non_interactive("--dangerous-skip-interactive")?;
+/// Resolve where to post and which operator approves. No post target is
+/// built for `--dry-run` or `--skip-post`; `select` chooses among multiple
+/// saved operator IDs when `--operator-id` was not given.
+async fn build_post_target(
+    args: BuildPostTargetArgs,
+    fetched_manifest_id: Option<Uuid>,
+    select: impl FnOnce(Vec<Uuid>) -> anyhow::Result<Option<Uuid>>,
```
daniilrrr, line 418: I don't see what this abstraction gains us, and the `build_post_target()` is only used in one place. Maybe this was meant to be tested with a simpler select? 

As-is, I think this should be refactored out to make the logic clearer. 

LLM output: 

```rust
let (post_target, operator_id) = build_post_target(
      ctx,
      BuildPostTargetArgs::from(&args),
      fetched.as_ref().map(|(id, _)| *id),
  )
  .await?;
```

  Then:

```rust
  async fn build_post_target(
      ctx: &StdCtx,
      args: BuildPostTargetArgs,
      fetched_manifest_id: Option<Uuid>,
  ) -> anyhow::Result<(Option<PostTarget>, Option<Uuid>)> {
      if args.dry_run || args.skip_post {
          return Ok((None, args.operator_id));
      }

      let manifest_id = fetched_manifest_id
          .or(args.manifest_id)
          .ok_or(ApproveInputError::MissingManifestId)?;

      let operator_id = if let Some(operator_id) = args.operator_id {
          operator_id
      } else {
          let config = Config::load().await?;

          let saved_ids = config
              .get_last_operator_ids()
              .unwrap_or_default()
              .into_iter()
              .map(|id| {
                  Uuid::parse_str(&id)
                      .with_context(|| format!("saved operator ID '{id}' is not a UUID"))
              })
              .collect::<anyhow::Result<Vec<_>>>()?;

          match saved_ids.as_slice() {
              [] => bail!(
                  "--operator-id is required to post approval to API. \
                   No saved operator IDs found. \
                   Use --skip-post to only generate the approval locally."
              ),
              [operator_id] => *operator_id,
              _ if ctx.is_non_interactive() || !stdin_can_prompt() => {
                  bail!(
                      "--operator-id is required to post approval to API when \
                       multiple saved operator IDs are available"
                  )
              }
              _ => prompts::select("Select approving operator", saved_ids)?,
          }
      };

      Ok((
          Some(PostTarget {
              manifest_id,
              deploy_id: args.deploy_id,
          }),
          Some(operator_id),
      ))
  }
```

tvc/src/commands/deploy/approve.rs
```diff
@@ -511,34 +582,12 @@ async fn sign_and_write_approval(
...
-        .get_last_operator_ids()
-        .unwrap_or_default()
-        .into_iter()
-        .map(|id| {
-            Uuid::parse_str(&id).with_context(|| format!("saved operator ID '{id}' is not a UUID"))
-        })
-        .collect()
-}
-
+#[instrument(skip_all, fields(manifest_id = %plan.manifest_id, operator_id = %plan.operator_id, deploy_id = ?plan.deploy_id))]
```
daniilrrr, line 585: cool!

tvc/src/commands/deploy/status.rs
```diff
@@ -150,6 +185,18 @@ Pivot Container:
...
+            write!(
+                f,
+                "\n\nManifest Approvals: {}/{} valid",
+                meta.valid_count, validation.threshold
+            )?;
+            for validated in &validation.approvals {
+                write!(f, "\n  {}: {}", validated.approval, validated.verdict)?;
+            }
+            write!(f, "\nQuorum reached: {}", yes_no(meta.quorum_reached))?;
+        }
```
daniilrrr, line 198: i think there's a missed opportunity here for an `else {}` block with vital information for the user. if the `manifest_approvals` are `None` then something is wrong and the cli won't work, right? 


```suggestion
        if let Some(ApprovalValidationWithMeta { validation, meta }) = &self.manifest_approvals {
            write!(
                f,
                "\n\nManifest Approvals: {}/{} valid",
                meta.valid_count, validation.threshold
            )?;
            for validated in &validation.approvals {
                write!(f, "\n  {}: {}", validated.approval, validated.verdict)?;
            }
            write!(f, "\nQuorum reached: {}", yes_no(meta.quorum_reached))?;
        } else {
            write!(
                f,
                "\n\nManifest Approvals: unable to validate \
                 (deployment manifest could not be parsed)\n\
                 Quorum reached: unknown"
            )?;
        }
```

tvc/src/commands/deploy/approve.rs
richardpringle: Good catch. Things came full circle. I refactor where I needed it (it saved a bunch of duplicate logic), then refactored again and it looks like I don't need it anymore

richardpringle:
> no blocking issues found! one note that's not in scope of this PR:
> 
> * we generate a local share approval in re_encrypt_local_share https://github.com/tkhq/rust-sdk/blob/22acc9c359b309df55b11549050d1c36800276f8/tvc/src/commands/keys/re_encrypt_local_share.rs#L154-L163
> * that approval is created during post_share
>   https://github.com/tkhq/rust-sdk/blob/22acc9c359b309df55b11549050d1c36800276f8/tvc/src/commands/deploy/post_share.rs#L72-L80
> 
> could be nice to have this parallel check in that flow, but it's less dire since the main hosted flow is in the server

@annieke https://linear.app/turnkey/issue/TVC-257

tvc/src/commands/deploy/approve.rs
richardpringle: https://linear.app/turnkey/issue/TVC-258/

tvc/src/commands/deploy/approve.rs
richardpringle: https://linear.app/turnkey/issue/TVC-259

tvc/src/commands/deploy/approve.rs
richardpringle: I think it's low priority, but definitely worth it because it removes the maintenance burden of message saying they're mutually exclusive. I have a ticket here, but might get to this as a part of the error message audit:
https://linear.app/turnkey/issue/TVC-260

PR #213, opened 2026-07-24
chore(tvc): type UUID-backed CLI IDs as uuid::Uuid

daniilrrr:
## SYS-103: Type UUID-backed `tvc` CLI IDs as `uuid::Uuid`

Refactors Rust-sdk's `tvc` CLI/internal ID arguments that are semantically UUIDs from `String` to `uuid::Uuid`. Parsing/validation happens at the Clap input boundary (via `uuid::Uuid`'s `FromStr` impl — clap parses it natively), and conversion back to `String` happens only at the generated-API-client boundary (`.to_string()` where the request/intent struct is built).

Invalid UUID values now produce a clear CLI error (`clap` `ValueValidation`, exit code 2) **before** any API request is made. Existing API request behavior is preserved — the generated client request/intent types stay `String`.

Also satisfies item **F10 (client-side UUID validation)** of the TVC-225 audit.

### Fields converted `String` → `uuid::Uuid`

| Command | Field |
|---|---|
| `deploy get-status` | `deploy_id` |
| `deploy status` | `deploy_id` |
| `deploy delete` | `deploy_id` |
| `deploy restore` | `deploy_id` |
| `deploy debug-logs` | `deploy_id` |
| `deploy provisioning-details` | `deploy_id` |
| `deploy approve` | `deploy_id` (`Option<Uuid>`) |
| `deploy post-share` | `share_operator_id` |
| `app delete` | `app_id` |
| `app set-live-deploy` | `deploy_id` |
| `app status` | `app_id` |

Each is converted back with `.to_string()` at the point the generated request/intent (`GetTvcDeploymentRequest`, `DeleteTvcDeploymentIntent`, `GetAppStatusRequest`, `PostTvcQuorumKeyShareIntent`, etc.) or the internal request-plan struct is built. The generated client types (`client/src/generated/...`) are untouched and remain `String`.

### UUID-looking fields intentionally left as `String` (and why)

- **`deploy create --app-id`** (`Overrides.app_id`): this is a config-file **override** merged into `DeployConfig.app_id`, which is deserialized from JSON and already UUID-validated in the config layer (`config/deploy.rs`). Converting it would ripple into the JSON-backed config structs — out of scope for a surgical ID-typing change.
- **`deploy approve --manifest-id`**: `resolve_manifest_id` interleaves the user-supplied `--manifest-id` with a manifest ID **fetched from the API** (a `String` from the response). Typing the arg as `Uuid` would pull response-side data into UUID typing — out of scope and behavior-risky.
- **`org_id` / `organization_id`**: resolved from env/config (e.g. `TVC_ORG_ID`) or persisted config, not a direct CLI-parsed request ID; kept `String`.
- **`app list --name`**: a substring match, not an ID.
- **Org config `alias`**: a user-chosen name, not a UUID.
- **Response / output / persisted-artifact fields** (e.g. `AppDeleted.activity_id`, `DeploymentStatusReport.*`, `QuorumKeySharePosted.provisioning_share_id`, `ProvisionBundle.deployment_id`, attestation `module_id`, `quorum_key_id`/`share_ids` outputs): these carry values received from the API or written to disk, not values validated at the CLI input boundary.

### Dependency

`uuid` was already a direct dependency of the `tvc` crate (`uuid = { workspace = true }`, workspace pin `1.13.1`), so no `Cargo.toml` change was required.

### Tests

Kept minimal per direction: a single smoke test in `deploy/get_status.rs` covering one valid-UUID parse and one invalid-UUID parse (`clap` `ValueValidation`). Existing repo integration tests that passed non-UUID placeholder IDs (e.g. `app_test`, `deploy-123`, `deploy_test`, `some-deploy-id`) were updated to valid UUIDs so they still exercise their original auth/API/conflict paths past the now-strict parse step.

### Verification

`cargo fmt -p tvc`, `cargo build -p tvc`, `cargo test -p tvc`, and `cargo clippy -p tvc -- -D warnings` all pass. Diff is `tvc/`-only.

### DIFF
--- a/tvc/src/commands/app/delete.rs
+++ b/tvc/src/commands/app/delete.rs
@@ -9,6 +9,7 @@ use serde::Serialize;
 use std::fmt::{self, Display, Formatter};
 use std::time::{SystemTime, UNIX_EPOCH};
 use turnkey_client::generated::{ActivityStatus, DeleteTvcAppAndDeploymentsIntent};
+use uuid::Uuid;
 
 /// Delete a TVC application and all of its deployments.
 #[derive(Debug, ClapArgs)]
@@ -16,7 +17,7 @@ use turnkey_client::generated::{ActivityStatus, DeleteTvcAppAndDeploymentsIntent
 pub struct Args {
     /// ID of the app to delete.
     #[arg(long, value_name = "APP_ID", env = "TVC_APP_ID")]
-    pub app_id: String,
+    pub app_id: Uuid,
 }
 
 /// Run the app delete command.
@@ -24,7 +25,7 @@ pub async fn run(_ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
     let auth = build_client().await?;
 
     let intent = DeleteTvcAppAndDeploymentsIntent {
-        app_id: args.app_id.clone(),
+        app_id: args.app_id.to_string(),
     };
 
     let timestamp_ms = SystemTime::now()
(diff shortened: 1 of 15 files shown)


annieke (approved):
nice!!

PR #218, opened 2026-07-29
feat(tvc): stamp X-TVC-CLIENT-VERSION and add version subcommand

richardpringle:
The backend gates TVC requests on a minimum tvc release version (enforce-when-present). The CLI now stamps `X-TVC-CLIENT-VERSION: <CARGO_PKG_VERSION>` on every API request, injected as a reqwest default header in the single client constructor — no changes to the published `turnkey_client` crate.

Also adds `tvc version`: bare version in human mode, `{"reason":"version","version":"0.13.0"}` in JSON mode.

Closes ENG-4082.

### Manual Test

I tested this in dev. I made sure that versions without the explicit header don't fail. When the version is explicitly set to an old version, this is what the error message looks like: 

```
error: failed to create TVC deployment: HTTP response was not successful: 400 ({"code":3,"message":"tvc 0.12.0 is older than the minimum version this backend supports (0.12.2); upgrade tvc to the latest release","details":[],"turnkeyErrorCode":""})
```

I would argue that this it's easy to discern that an upgrade is needed and should be pretty easy to figure out how to perform the upgrade. We should however find a way to handle this error better a display a better message. The problem with doing that now is there could be someone using the client code that wouldn't expect this?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4220,6 +4220,7 @@ dependencies = [
  "turnkey_client",
  "turnkey_enclave_encrypt",
  "uuid",
+ "wiremock",
  "zeroize",
 ]
 
--- a/tvc/Cargo.toml
+++ b/tvc/Cargo.toml
@@ -47,3 +47,4 @@ predicates = { workspace = true }
 qos_nsm = { workspace = true, features = ["mock"] }
 rexpect = { workspace = true }
 tempfile = { workspace = true }
+wiremock = { workspace = true }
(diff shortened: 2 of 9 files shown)


Cargo.lock
```diff
@@ -4220,6 +4220,7 @@ dependencies = [
  "turnkey_client",
  "turnkey_enclave_encrypt",
  "uuid",
+ "wiremock",
```
richardpringle, line 4223: already a dependency

Avi-D-coder (approved):

emostov (approved):

PR #223, opened 2026-07-29
refactor(tvc): serde-tagged outcome vocabulary and Run trait pilot

richardpringle:
## Serde-tagged `Outcome` + `Run` trait pilot

Follow-up to #188's outcome model. Two changes:

**The `reason` tag now comes from the type.** `Outcome` is internally tagged (`#[serde(tag = "reason", rename_all = "snake_case")]`), so the variant name *is* the wire reason: the vocabulary can't drift from the type, and two shapes can't share a reason without rustc rejecting the duplicate variant name. The hand-written `reason()` match and the `Message` trait are gone — `Shell::emit` takes `Serialize + Display`, and human rendering is each payload's `Display`.

**Commands migrate to a `Run` trait.** [`Run`](tvc/src/commands/mod.rs) consumes parsed `Args` and returns a command-local outcome type; `Into<Outcome>` is the whole contract. `deploy approve` is the pilot; everything else stays on the legacy free `run` functions and migrates incrementally.

Notes:

- Wire contract unchanged: the reason vocabulary is byte-identical to `main`; variant renames (`Login`→`LoggedIn`, `DeployStatus`→`DeploymentStatus`, …) track the reason strings that already existed. One addition: `manifest_approval_already_posted`, reserved for the duplicate-approval detection coming in ENG-4274 (payload type only; nothing produces it yet).
- This lands the compile-time reason uniqueness TVC-207 asks for — by construction rather than by proc-macro. The collision check against the 3 non-terminal reasons is still a runtime test, so the ticket stays open for that remainder.
- Toolchain 1.94 → 1.95; `strum` (dev) for the outcome vocabulary tests.

Part of TVC-207.

### DIFF
--- a/codegen/src/transform.rs
+++ b/codegen/src/transform.rs
@@ -404,15 +404,13 @@ fn extract_field_enum_type(field: &syn::Field) -> Option<EnumField> {
 
         for meta in metas {
             match meta {
-                syn::Meta::NameValue(nv) => {
-                    if nv.path.is_ident("enumeration") {
-                        if let syn::Expr::Lit(syn::ExprLit {
-                            lit: syn::Lit::Str(lit_str),
-                            ..
-                        }) = &nv.value
-                        {
-                            name = Some(lit_str.value());
-                        }
+                syn::Meta::NameValue(nv) if nv.path.is_ident("enumeration") => {
+                    if let syn::Expr::Lit(syn::ExprLit {
+                        lit: syn::Lit::Str(lit_str),
+                        ..
+                    }) = &nv.value
+                    {
+                        name = Some(lit_str.value());
                     }
                 }
                 syn::Meta::Path(p) => {
--- a/tvc/Cargo.toml
+++ b/tvc/Cargo.toml
@@ -46,5 +46,6 @@ hpke = { workspace = true }
 predicates = { workspace = true }
 qos_nsm = { workspace = true, features = ["mock"] }
 rexpect = { workspace = true }
+strum = { workspace = true }
 tempfile = { workspace = true }
 wiremock = { workspace = true }
--- a/tvc/src/cli.rs
+++ b/tvc/src/cli.rs
@@ -1,9 +1,9 @@
 //! CLI parsing and dispatch.
 
-use crate::commands;
+use crate::commands::{self, Run};
 use crate::errors::strip_ansi;
 use crate::outcome::Outcome;
-use crate::output::{ColorChoice, Ctx, ErrorMessage, Message, MessageFormat, Shell, StdCtx};
+use crate::output::{ColorChoice, Ctx, ErrorMessage, MessageFormat, Shell, StdCtx};
 use clap::{ArgAction, Parser, Subcommand, builder::BoolishValueParser, error::ErrorKind};
 use std::ffi::OsString;
 use std::io::Write;
@@ -159,8 +159,12 @@ fn handle_parse_error(error: clap::Error) -> ExitCode {
                     .to_string();
                 let error_message = ErrorMessage::usage_error(message);
                 let mut stdout = std::io::stdout();
+
+                // it shouldn't be possible for serialization to fail here.
+                let msg = serde_json::to_string(&error_message).unwrap_or_else(|e| e.to_string());
+
                 // Best-effort: if writing fails we still exit with the usage code.
-                let _ = writeln!(stdout, "{}", error_message.to_json_string());
+                let _ = writeln!(stdout, "{msg}");
                 std::process::exit(USAGE_ERROR_EXIT_CODE)
             } else {
                 error.exit()
@@ -193,7 +197,7 @@ impl Commands {
     async fn run(self, ctx: &mut StdCtx) -> anyhow::Result<Outcome> {
         match self {
             Commands::Deploy { command } => match command {
-                DeployCommands::Approve(args) => commands::deploy::approve::run(ctx, args).await,
+                DeployCommands::Approve(args) => args.run(ctx).await.map(Into::into),
                 DeployCommands::GetStatus(args) => {
                     commands::deploy::get_status::run(ctx, args).await
                 }
--- a/tvc/src/commands/deploy/approve.rs
+++ b/tvc/src/commands/deploy/approve.rs
@@ -2,6 +2,7 @@
 
 use crate::{
     client::build_client,
+    commands::Run,
     config::turnkey::Config,
     errors::MissingResource,
     local_operator_key::LocalOperatorSeedSource,
@@ -129,6 +130,16 @@ pub enum ApproveOutcome {
     DryRun(ApprovalDryRun),
 }
 
+impl From<ApproveOutcome> for Outcome {
+    fn from(outcome: ApproveOutcome) -> Self {
+        match outcome {
+            ApproveOutcome::Posted(msg) => Outcome::ManifestApprovalPosted(msg),
+            ApproveOutcome::NotPosted(msg) => Outcome::ManifestApprovalGenerated(msg),
+            ApproveOutcome::DryRun(msg) => Outcome::ManifestApprovalDryRun(msg),
+        }
+    }
+}
+
 impl Serialize for ApproveOutcome {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         match self {
@@ -231,6 +242,23 @@ impl Display for ApprovalGenerated {
 }
 
 #[derive(Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ApprovalAlreadyPosted {
+    operator_id: String,
+    approval_id: String,
+}
+
+impl Display for ApprovalAlreadyPosted {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "Operator {} has already approved this manifest (approval ID: {}). Nothing to post.",
+            self.operator_id, self.approval_id
+        )
+    }
+}
+
+#[derive(Default, Serialize)]
 pub struct ApprovalDryRun {}
 
 impl Display for ApprovalDryRun {
@@ -254,20 +282,23 @@ struct ResolvedApproveInputs {
     post_target: Option<PostTarget>,
 }
 
-pub async fn run(ctx: &mut StdCtx, mut args: Args) -> anyhow::Result<Outcome> {
-    let operator_seed_source = LocalOperatorSeedSource::from_args(
-        args.operator_seed.take(),
-        args.operator_seed_path.take(),
-    )?;
+impl Run for Args {
+    type Outcome = ApproveOutcome;
 
-    let inputs = if ctx.is_non_interactive() {
-        build_inputs_non_interactive(ctx, args, operator_seed_source).await?
-    } else {
-        build_inputs_interactive(ctx, args, operator_seed_source).await?
-    };
+    async fn run(mut self, ctx: &mut StdCtx) -> anyhow::Result<Self::Outcome> {
+        let operator_seed_source = LocalOperatorSeedSource::from_args(
+            self.operator_seed.take(),
+            self.operator_seed_path.take(),
+        )?;
 
-    let outcome = run_with_resolved_inputs(ctx, inputs).await?;
-    Ok(Outcome::DeployApprove(outcome))
+        let inputs = if ctx.is_non_interactive() {
+            build_inputs_non_interactive(ctx, self, operator_seed_source).await?
+        } else {
+            build_inputs_interactive(ctx, self, operator_seed_source).await?
+        };
+
+        run_with_resolved_inputs(ctx, inputs).await
+    }
 }
 
 async fn build_inputs_interactive(
@@ -756,7 +787,6 @@ async fn fetch_manifest_from_deploy(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::output::Message;
     use turnkey_client::generated::external::data::v1::{
         TvcOperator, TvcOperatorApproval, TvcOperatorSet,
     };
@@ -929,10 +959,8 @@ mod tests {
 
     #[test]
     fn approval_posted_serializes_expected_json() {
-        let value: serde_json::Value = serde_json::from_str(
-            &Outcome::DeployApprove(ApproveOutcome::Posted(posted_to_file())).to_json_string(),
-        )
-        .unwrap();
+        let value =
+            serde_json::to_value(Outcome::from(ApproveOutcome::Posted(posted_to_file()))).unwrap();
 
         assert_eq!(
             value,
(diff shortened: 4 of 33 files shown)


codegen/src/transform.rs
```diff
@@ -404,8 +404,8 @@ fn extract_field_enum_type(field: &syn::Field) -> Option<EnumField> {
 
         for meta in metas {
             match meta {
-                syn::Meta::NameValue(nv) => {
-                    if nv.path.is_ident("enumeration") {
+                syn::Meta::NameValue(nv)
+                    if nv.path.is_ident("enumeration") => {
```
richardpringle: From new clippy

Avi-D-coder (dismissed):
checked that dep is already in mono

emostov (dismissed):

codegen/src/transform.rs
```diff
@@ -404,15 +404,13 @@ fn extract_field_enum_type(field: &syn::Field) -> Option<EnumField> {
...
-                        {
-                            name = Some(lit_str.value());
-                        }
+                syn::Meta::NameValue(nv) if nv.path.is_ident("enumeration") => {
+                    if let syn::Expr::Lit(syn::ExprLit {
+                        lit: syn::Lit::Str(lit_str),
+                        ..
+                    }) = &nv.value
+                    {
+                        name = Some(lit_str.value());
```
richardpringle, line 413: new for Rust 1.95

tvc/src/commands/deploy/approve.rs
```diff
@@ -129,6 +130,16 @@ pub enum ApproveOutcome {
...
 
+impl From<ApproveOutcome> for Outcome {
+    fn from(outcome: ApproveOutcome) -> Self {
+        match outcome {
+            ApproveOutcome::Posted(msg) => Outcome::ManifestApprovalPosted(msg),
+            ApproveOutcome::NotPosted(msg) => Outcome::ManifestApprovalGenerated(msg),
+            ApproveOutcome::DryRun(msg) => Outcome::ManifestApprovalDryRun(msg),
+        }
+    }
+}
```
richardpringle, line 141: This is the new pattern enforced by the `Run` trait

tvc/src/commands/operator/create.rs
```diff
@@ -67,7 +67,6 @@ pub struct Args {
 pub struct OperatorCreated {
     name: String,
     operator_id: Uuid,
-    wallet_id: Uuid,
```
richardpringle, line 70: whoops

tvc/src/cli.rs
```diff
@@ -159,8 +159,12 @@ fn handle_parse_error(error: clap::Error) -> ExitCode {
                     .to_string();
                 let error_message = ErrorMessage::usage_error(message);
                 let mut stdout = std::io::stdout();
+
+                // it shouldn't be possible for serialization to fail here.
+                let msg = serde_json::to_string(&error_message).unwrap_or_else(|e| e.to_string());
+
                 // Best-effort: if writing fails we still exit with the usage code.
-                let _ = writeln!(stdout, "{}", error_message.to_json_string());
```
richardpringle, line 163: This will panic if there is a serialization error

tvc/src/outcome.rs
```diff
@@ -1,208 +1,138 @@
...
-            Outcome::AppInit($msg) => $body,
-            Outcome::AppSetLiveDeploy($msg) => $body,
-            Outcome::AppDelete($msg) => $body,
-            Outcome::KeysCreateQuorumKey($msg) => $body,
-            Outcome::KeysGenerateQuorumKey($msg) => $body,
-            Outcome::KeysInitQuorumKey($msg) => $body,
-            Outcome::KeysReEncryptShare($msg) => $body,
-        }
-    };
-}
```
richardpringle, line 80: Was only being used in one place after moving the `reason` logic

tvc/src/outcome.rs
```diff
@@ -1,208 +1,138 @@
...
+use std::fmt::{self, Display, Formatter};
 
-/// One wide terminal outcome per command (the wide-event model).
+/// One wide terminal outcome per command invocation (the wide-event model).
 ///
 /// Streaming messages (today: only `deploy debug-logs`'s per-line
 /// `debug_log_line`) are emitted inline by their command and are not part of
 /// this enum; the command still returns its terminal variant.
+#[derive(Serialize)]
+#[serde(tag = "reason", rename_all = "snake_case")]
```
richardpringle, line 30: Using `serde`'s derive functionality is a lot easier to scale. We have compile-time enforced consistency with `rename_all` and guaranteed uniqueness. Furthermore, this mechanism could be valuable if we ever want to implement `Deserialize` to do something automated with `Outcome`

tvc/src/outcome.rs
```diff
@@ -1,208 +1,138 @@
...
-            Outcome::KeysGenerateQuorumKey(
-                keys::generate_local_quorum_key::QuorumKeyGenerated::default(),
-            ),
-            Outcome::KeysInitQuorumKey(
-                keys::init_local_quorum_key::QuorumKeyConfigCreated::default(),
-            ),
-            Outcome::KeysReEncryptShare(
-                keys::re_encrypt_local_share::ReEncryptedShareGenerated::default(),
-            ),
-        ]
```
richardpringle, line 180: Replaced with [`strum::IntoEnumIterator`](https://docs.rs/strum/latest/strum/trait.IntoEnumIterator.html), exact same behaviour, less error prone when adding new variants

tvc/src/output.rs
```diff
@@ -28,27 +28,6 @@ pub enum ColorChoice {
...
-
-    fn human_message(&self) -> String;
-
-    fn to_json_string(&self) -> String {
-        #[derive(Serialize)]
-        struct WithReason<'a, S: Serialize + ?Sized> {
-            reason: &'a str,
-            #[serde(flatten)]
-            msg: &'a S,
-        }
```
richardpringle, line 42: exact same behaviour as tagged enum

tvc/src/output.rs
```diff
@@ -28,27 +28,6 @@ pub enum ColorChoice {
     Never,
 }
 
-pub trait Message: Serialize {
-    fn reason(&self) -> &'static str;
```
richardpringle, line 32: This wasn't being used anywhere other than serialization. If we want to use it elsewhere, we mimic the `serde_derive` behaviour

tvc/src/output.rs
```diff
@@ -94,22 +73,27 @@ impl<W, W2> Shell<W, W2> {
...
+                let text = message.to_string();
+
                 if text.is_empty() {
                     return Ok(());
                 }
+
                 self.human().line(text)
             }
             MessageFormat::Json => {
-                writeln!(self.stdout, "{}", message.to_json_string())?;
```
richardpringle, line 112: Used to panic if serialization fails

emostov (dismissed):

tvc/Cargo.toml
```diff
@@ -46,4 +46,5 @@ hpke = { workspace = true }
 predicates = { workspace = true }
 qos_nsm = { workspace = true, features = ["mock"] }
 rexpect = { workspace = true }
+strum = { version = "0.27.2", default-features = false, features = ["derive", "std", "strum_macros"] }
```
daniilrrr: this is the only dep defined within `tvc` and not at project root. is this intentional? could we define at root to be aligned with the rest of the repo?

tvc/src/commands/deploy/debug_logs.rs
```diff
@@ -229,12 +230,8 @@ pub(crate) struct DebugLogLine {
...
-impl Message for DebugLogLine {
-    fn reason(&self) -> &'static str {
-        "debug_log_line"
-    }
-
-    fn human_message(&self) -> String {
+impl Display for DebugLogLine {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         // Rebuild the API line and defer to `format_log_line` so human output
         // stays byte-identical to the pre-outcome rendering.
```
daniilrrr: nit - I love Rust's inline links and use them where possible

```suggestion
    /// Rebuild the API line and defer to [`format_log_line`] so human output
    /// stays byte-identical to the pre-outcome rendering.
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
```

tvc/src/errors.rs
```diff
@@ -103,7 +104,7 @@ pub fn classify(error: &anyhow::Error) -> Classification {
 ///
 /// Every source link is preserved, even when a parent error's `Display`
 /// already embeds the source text.
-pub fn render_error_chain(error: &anyhow::Error) -> String {
+pub(crate) fn render_error_chain(error: &anyhow::Error) -> String {
```
richardpringle, line 107: @daniilrrr, I added the visibility scope because `MAX_ERROR_MESSAGE_CHARS` is private. Not sure if that was the right thing to do or to make the constant pub.

tvc/src/outcome.rs
daniilrrr: super clever! and now we don't need proc-macros to enforce uniqueness or non-collision. rad

tvc/src/lib.rs
```diff
@@ -11,7 +11,7 @@ pub mod cli;
 pub mod client;
 pub mod commands;
 pub mod config;
-pub mod errors;
+pub(crate) mod errors;
```
richardpringle, line 14: I did this because of some docs errors (mixing private and public types in docs). 

Minimizing the public API is generally the safer thing to do in case someone decided to just start using this  crate as a library somewhere else (internally or externally).

tvc/src/outcome.rs
```diff
@@ -1,212 +1,140 @@
...
     const NON_TERMINAL_REASONS: [&str; 3] =
         ["debug_log_line", "command_error", "missing_required_input"];
 
-    fn all_reasons() -> Vec<&'static str> {
-        let mut reasons: Vec<&str> = all_terminal_shapes().iter().map(Message::reason).collect();
-        reasons.extend(NON_TERMINAL_REASONS);
-        reasons
-    }
-
     // TODO - proc macro here
```
daniilrrr, line 110: delete me

tvc/src/errors.rs
```diff
@@ -50,6 +50,7 @@ pub enum ErrorCode {
     MissingRequiredInput,
     /// Bad flags/args — a clap parse failure
     UsageError,
+    #[allow(dead_code)]
```
richardpringle, line 53: The only reason this wasn't caught as dead code before is because it was public. It seems that this is likely a mistake?

tvc/src/commands/deploy/approve.rs
daniilrrr: this rocks, excited for the rest of the commands to do this in follow-up PR

tvc/src/outcome.rs
richardpringle: see https://linear.app/turnkey/issue/TVC-242/

tvc/src/commands/mod.rs
```diff
@@ -13,3 +16,14 @@ pub mod keys;
...
+
+/// A command: consumes its parsed `Args`, returns its typed terminal outcome.
+///
+/// The associated `Outcome` is a command-local type; `Into<outcome::Outcome>`
+/// is the whole contract, since serialization (including the `reason` tag) and
+/// human rendering both live on the top-level [`outcome::Outcome`].
+pub trait Run {
+    type Outcome: Into<outcome::Outcome>;
+
+    fn run(self, ctx: &mut StdCtx) -> impl Future<Output = anyhow::Result<Self::Outcome>> + Send;
```
daniilrrr, line 28: (shouldn't do this in this PR, or anytime soon, just wanted to write down the thought)

seeing the trait written out so clearly here and the corresponding interactive + non-interactive impl in `approve.rs` makes me think there is some opportunity to abstract `RunInput` construction out to make it clear that both an interactive and non-interactive path is/must be supported, and also be well-tested and arrive at the same outputs. Basically we could have an optional trait for mode-aware commands that has a `Run` `impl`, and a `Command` must implement at least this existing one or the mode aware one

I think we are many commands away from wanting this, but it would be a nice way to tame code sprawl and prevent drift, perhaps surface errors a smidge faster, etc. 

brainstormed the following with LLM:
___
```
model the lifecycle explicitly:

  Args
    ├─ interactive resolution ───┐
    └─ non-interactive resolution ┤
                                  ▼
                           ResolvedInputs
                                  ▼
                              execute
                                  ▼
                               Outcome
```
```rust
I would keep the general Run trait and introduce an opt-in trait:

pub trait Run {
    type Outcome: Into<outcome::Outcome>;

    fn run(
        self,
        ctx: &mut StdCtx,
    ) -> impl Future<Output = anyhow::Result<Self::Outcome>> + Send;
}

pub trait ResolvedCommand: Sized + Send {
    type Inputs: Send;
    type Outcome: Into<outcome::Outcome>;

    fn resolve_interactive(
        self,
        ctx: &mut StdCtx,
    ) -> impl Future<Output = anyhow::Result<Self::Inputs>> + Send;

    fn resolve_non_interactive(
        self,
        ctx: &mut StdCtx,
    ) -> impl Future<Output = anyhow::Result<Self::Inputs>> + Send;

    fn execute(
        ctx: &mut StdCtx,
        inputs: Self::Inputs,
    ) -> impl Future<Output = anyhow::Result<Self::Outcome>> + Send;
}

Then provide a blanket Run implementation for ResolvedCommand:

impl<T> Run for T
where
    T: ResolvedCommand,
{
    type Outcome = <T as ResolvedCommand>::Outcome;

    fn run(
        self,
        ctx: &mut StdCtx,
    ) -> impl Future<Output = anyhow::Result<Self::Outcome>> + Send {
        async move {
            let inputs = if ctx.is_non_interactive() {
                self.resolve_non_interactive(ctx).await?
            } else {
                self.resolve_interactive(ctx).await?
            };

            T::execute(ctx, inputs).await
        }
    }
}

  Simple commands implement Run directly. Commands with mode-specific input acquisition implement ResolvedCommand and get Run automatically.
```
```

One additional wrinkle: approve currently performs shared seed-source preparation before branching. The new structure should avoid duplicating that work. Either keep a
small preparation step inside the resolver implementation or eventually model a three-stage pipeline:

Args → PreparedArgs → ResolvedInputs → Outcome

I would not add that third abstraction until multiple migrated commands demonstrate the same need.
```

daniilrrr (approved):
great change, many lines removed, elegant use of Serde
🥳

tvc/src/errors.rs
daniilrrr: thats fine, dont think the constant should be pub

tvc/src/errors.rs
daniilrrr: yeah I agree, all of the error code I introduced should have been `pub(crate)` not `pub`. then we would have caught the dead code easier 

looks like `InvalidInput` was never-used slop that snuck past the compiler. you can remove `InvalidInput` here in your PR or leave it for a quick follow up, up to you

emostov (approved):

tvc/src/errors.rs
richardpringle: Leaving for a follow up. All good

tvc/src/commands/mod.rs
richardpringle: Makes sense to me. I would go a step further and split `StdCtx` into `InteractiveCtx` and `StdCtx` forcing all prompts to go through `Ctx`. 

Never ending rabbit hole though...

PR #224, opened 2026-07-30
fix(tvc): one profile per organization, id-keyed key dirs, operator key backup

richardpringle:
Combines TVC-159, TVC-52, TVC-53, and TVC-55 into one PR.

## TVC-159 — deterministic alias matching

`tvc login --org <org-id>` resolved through `HashMap` iteration order, so an organization registered under multiple aliases logged into an arbitrary profile per process; `tvc profile delete` shared the bug.

- org IDs are `Uuid` end to end (config, payloads, `--org` queries parse into ID-vs-alias at the clap boundary); `Config::orgs` is an `IndexMap` (+ toml `preserve_order`), so config-file order is preserved and saves don't churn the file
- `default_alias` markers are repaired at config load: every duplicated org ID gets exactly one — the first marked profile in file order, else the first profile — persisted lazily on the next save, so resolution can always follow the marker
- non-interactive: org-ID queries resolve to the default with a warning; an explicitly named secondary alias fails with instructions; `tvc profile set-default-alias --org <alias>` switches the marker
- interactive login consolidates pre-existing duplicates: pick the profile to keep per org, one confirmation, losers get full profile-delete cleanup (config saved after each deletion), active profile repaired onto the keeper
- login refuses a new profile for an already-configured org ID, and refuses reusing an existing alias (previously a silent config overwrite)
- `tvc profile delete` never guesses among duplicates: org-ID queries prompt interactively and fail fast non-interactively, marker ignored

## TVC-52 — `tvc keys backup-operator-key`

Copies the org's operator key file byte-for-byte after validating it parses (unknown fields survive — no re-serialize). Interactive runs prompt for the destination and confirm overwrites; non-interactive requires `--output`/`--force`. Output includes the private-key warning, storage guidance, and the manual restore path (restore command deferred). Org resolution shares login's read-only rules via `resolve_org_query`. Adopts the `Run` trait from #223.

## TVC-53 — onboarding backup nudge

When login generates a fresh operator key interactively, it explains the key exists only on this machine and offers a backup (default Yes) via the shared core; backup errors degrade to a warning so login never dies. Declining prints the pointer to the standalone command; re-logins with an existing key get a single tip line. A one-shot mock whoami server carries PTY login past credential verification, so the accept/decline/tip paths are all covered end-to-end.

## TVC-55 — org-id keyed key directories

New profiles store keys under `orgs/<org-id>/` (UUID directories autocomplete; aliases can contain spaces). The alias is untouched everywhere else — config key, prompts, `--org`. No disk-schema bump: paths are data, so v1 reads both layouts, `profile delete` recognizes and cleans up either (and refuses to delete a directory another profile still points into), and interactive login migrates legacy `orgs/<alias>/` directories right after consolidation — rename, rewrite paths, save per profile, failures degrade to warnings. Non-interactive runs never touch the layout.

Closes TVC-159. Closes TVC-52. Closes TVC-53. Closes TVC-55.


tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,258 @@
...
+
+/// Back up a local operator key by copying its key file to a chosen
+/// destination.
+#[derive(Debug, ClapArgs)]
+#[command(about, long_about = None)]
+pub struct Args {
+    /// Organization alias or ID whose operator key to back up.
+    /// Defaults to the active organization.
+    #[arg(long, env = "TVC_ORG", value_name = "ORG")]
+    org: Option<String>,
```
richardpringle: Should be a UUID

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        let config = Config::load().await?;
+
+        let alias = match &self.org {
+            Some(query) => resolve_org_query(ctx, &config, query)?,
+            None => match config.active_org_config() {
+                Some((alias, _)) => alias.clone(),
+                None => bail!("No active organization. Run `tvc login` first."),
+            },
+        };
```
richardpringle: You're violating rules that I explicitly gave you including empty lines surrounding code blocks. 

```suggestion
        let alias = self.org
            .as_ref()
            .map(|query| resolve_org_query(ctx, &config, query))
            .or_else(|| {
                config
                    .active_org_config()
                    .map(|(alias, _)| alias.clone().ok_or_else(|| anyhow!("No active organization. Run `tvc login` first."))
            })
            .transpose()?;
```

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,258 @@
...
+        let org_config = config
+            .orgs
+            .get(&alias)
+            .expect("alias was resolved from this config");
```
richardpringle: 😞 

DO NOT DO THIS EVER

Find the org-config on the same path that you look for the alias and return them in a tuple.

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,258 @@
...
+        let source = &org_config.select_local_record(&alias)?.key_path;
```
richardpringle: Is the `org_config` not aware of its own alias? Something seems wrong with the data model here .

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        let destination = match self.output {
+            Some(path) => path,
+            None => PathBuf::from(prompts::text(
+                "Backup file path",
+                Some(&format!("operator-{alias}-backup.json")),
+            )?),
+        };
```
richardpringle: Quit nesting everything and use `.map(Ok).unwrap_or_else(/* prompt */)?`

richardpringle: And you should be able to use `.into` instead of `PathBuf::from`

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+}
+
+/// Copy the operator key at `source` to `destination` byte-for-byte.
+///
+/// The source is parsed to validate it and capture its public key, but the
+/// original bytes are written verbatim so any unknown fields survive the
+/// copy.
+pub(crate) async fn back_up_key(
+    alias: &str,
```
richardpringle: If you need an owned type here, pass in an owned type instead of cloning in the function

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,258 @@
...
+    let bytes = match tokio::fs::read(source).await {
+        Ok(bytes) => bytes,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => bail!(
+            "No operator key found at {} for org '{alias}'. Run `tvc login` first.",
+            source.display()
+        ),
+        Err(e) => {
+            return Err(e)
+                .with_context(|| format!("failed to read operator key: {}", source.display()));
+        }
```
richardpringle: Both these arms exit. You should be using `let bytes = fs::read(source).await.or_else(/* ... */)?;` .

Why do you insist making code more complex than it needs to be. USE COMBINATORS

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,258 @@
...
+    let key: StoredQosOperatorKey = serde_json::from_slice(&bytes).with_context(|| {
+        format!(
+            "operator key at {} is not a valid operator key file",
+            source.display()
+        )
+    })?;
+
+    if let Some(parent) = destination.parent()
+        && !parent.as_os_str().is_empty()
+    {
```
richardpringle: It's not possible to have an empty string here. Why are you doing this?

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,258 @@
...
+    tokio::fs::write(destination, &bytes)
+        .await
+        .with_context(|| format!("failed to write backup: {}", destination.display()))?;
+
+    Ok(OperatorKeyBackedUp {
+        alias: alias.to_string(),
+        public_key: key.public_key,
+        source_path: source.display().to_string(),
+        backup_path: destination.display().to_string(),
+    })
```
richardpringle, line 151: This is simply bad Rust. You've borrowed `alias`, `source_path`, and `backup_path` to construct `public_key`. Why wouldn't you return the key and create the `OperatorKeyBackedUp` outcome outside this function where you can likely use owned values. This is bad code, be better

tvc/src/commands/keys/mod.rs
```diff
@@ -1,5 +1,6 @@
 //! Key management commands.
 
+pub mod backup_operator_key;
```
richardpringle: Does this need to be pub?

tvc/src/commands/login.rs
```diff
@@ -72,17 +90,104 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
...
+/// Nothing is mutated before the confirmation, so declining leaves the config
+/// and disk untouched.
+async fn consolidate_duplicate_profiles(
+    ctx: &mut StdCtx,
+    config: &mut Config,
+    duplicates: Vec<(String, Vec<String>)>,
+) -> Result<()> {
+    if duplicates.is_empty() {
+        return Ok(());
+    }
```
richardpringle: don't call this function if the vec is empty. Do the check outside

tvc/src/commands/login.rs
```diff
@@ -72,17 +90,104 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
...
+
+    shell_eprintln!(
+        ctx,
+        "Multiple profiles are configured for the same organization. tvc keeps \
```
```suggestion
        "Multiple profiles are configured for the same organization. `tvc` keeps \
```
But you should be able to get this name at compile time from an `env!`

tvc/src/commands/login.rs
```diff
@@ -72,17 +90,104 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
...
+
+        if let Some(active) = &config.active_org
+            && aliases.contains(active)
+            && *active != keeper
+        {
+            active_repair = Some(keeper.clone());
+        }
+
+        losers.extend(aliases.iter().filter(|alias| **alias != keeper).cloned());
+    }
```
richardpringle: There's gotta be a cleaner way to do this

tvc/src/commands/login.rs
```diff
@@ -72,17 +90,104 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
...
+        shell_println!(ctx)?;
+    }
+
+    // `remove_org` cleared `active_org` if the active profile was among the
+    // losers; its group's keeper inherits it.
+    if let Some(keeper) = active_repair {
+        config.set_active_org(&keeper)?;
+    }
+
+    config.save().await?;
```
richardpringle: Should we save after every delete? The config should be mutated, delete the file that was removed, then save the config. Either that or save the config before deleting any files. Please put more thought into the failure modes

tvc/src/commands/login.rs
```diff
@@ -212,35 +337,108 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<Outcome> {
...
-    config.save().await?;
-
     // A local delete does not touch the dashboard-registered API key, and we
     // can't tell whether it is still there, so hedge with "may" and give steps.
     let dashboard_url = dashboard_base_url(&removed.api_base_url);
 
-    Ok(Outcome::ProfileDeleted(ProfileDeleted {
-        alias,
+    Ok(ProfileDeleted {
+        alias: alias.to_string(),
```
richardpringle: Probably making the same mistake here. Return what needs to be returned to assemble the struct outside the function

tvc/src/commands/login.rs
```diff
@@ -212,35 +337,108 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<Outcome> {
...
         api_public_key,
+    })
+}
+
+/// Mark a profile as the default alias for its duplicated organization ID and
+/// clear the marker from the organization's other profiles. The marker only
+/// means something while duplicates exist, so a non-duplicated organization is
+/// refused.
+pub async fn run_set_default_alias(
+    _ctx: &mut StdCtx,
```
richardpringle: This is dumb, don't pass in `ctx` if you don't need it

tvc/src/commands/login.rs
```diff
@@ -212,35 +337,108 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<Outcome> {
...
+    args: SetDefaultAliasArgs,
+) -> Result<Outcome> {
+    let mut config = Config::load().await?;
```
richardpringle: Do all code paths need a config? Maybe pass in a mutable config from the outside

tvc/src/commands/login.rs
```diff
@@ -212,35 +337,108 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<Outcome> {
...
+            "Organization '{org_id}' only has the profile '{alias}'; \
+             a default alias is only needed while duplicate profiles exist."
+        );
+    }
+
+    for (other, org) in config.orgs.iter_mut() {
+        if org.id == org_id {
+            org.default_alias = *other == alias;
+        }
+    }
```
richardpringle: I know I usually say wait until you have three repetitions before you abstract, but I think it's an exception when there are effects. It looks like the same thing happens twice. It should happen in the same way and we should handle the failure modes in the exact same way

tvc/src/commands/login.rs
```diff
@@ -266,23 +465,57 @@ fn resolve_profile_alias(config: &Config, org: Option<String>) -> Result<String>
...
+    config: &'a Config,
+    aliases: &'a [String],
+) -> Vec<ProfileChoice<'a>> {
+    aliases
+        .iter()
+        .map(|alias| {
+            let org = config
+                .orgs
+                .get(alias)
+                .expect("alias was resolved from this config");
```
richardpringle: do not use expect, use `filter_map` instead

tvc/src/commands/login.rs
```diff
@@ -266,23 +465,57 @@ fn resolve_profile_alias(config: &Config, org: Option<String>) -> Result<String>
...
+}
+
 fn build_login_plan_interactive(
     ctx: &mut StdCtx,
     args: Args,
     config: &Config,
 ) -> Result<LoginPlan> {
     let org = match args.org {
-        Some(query) => OrgPlan::Existing(query),
+        Some(query) => OrgPlan::Existing(resolve_org_query(ctx, config, &query)?),
```
richardpringle: Is it possible to use `.map`, `.or_else`, and `.transpose` here instead of a match arm?

tvc/src/commands/login.rs
```diff
@@ -292,34 +525,160 @@ fn build_login_plan_interactive(
...
+            }
+
+            Ok(alias.clone())
+        }
+        _ if ctx.is_non_interactive() => {
+            let defaults: Vec<&str> = aliases
+                .iter()
+                .map(String::as_str)
+                .filter(|alias| config.orgs[*alias].default_alias)
+                .collect();
```
richardpringle: I don't like this, we have to loop over all the orgs to find the aliases? I feel like we need a way to validate the entire config and migrate it before doing anything else. If there are multiple orgs with different aliases and the same ID, only one should be marked as the default and that should be treated the alias to use. If there's no default on the first pass, we mark the "random" one then use that one forever. If the user goes in and makes multiple have the default alias, we should just pick up the first one in top down order of the toml. This should be solvable at deserialization time, not here.

tvc/src/commands/login.rs
```diff
@@ -292,34 +525,160 @@ fn build_login_plan_interactive(
...
+                .filter(|(_, org)| org.id == org_id)
+                .map(|(group_alias, _)| group_alias.as_str())
+                .collect();
+            group.sort_unstable();
+
+            if group.len() > 1 && ctx.is_non_interactive() {
+                let defaults: Vec<&str> = group
+                    .iter()
+                    .copied()
+                    .filter(|group_alias| config.orgs[*group_alias].default_alias)
```
richardpringle: See my other comment. This should be some sort of `find` not `filter`

tvc/src/commands/login.rs
```diff
@@ -292,34 +525,160 @@ fn build_login_plan_interactive(
...
+                    ),
+                    [] => return Err(no_default_alias_error(org_id, &group.join(", "))),
+                    marked => {
+                        return Err(multiple_default_aliases_error(org_id, &marked.join(", ")));
+                    }
+                }
+            }
+
+            Ok(alias.clone())
+        }
```
richardpringle: You're doing too much in each match arm. I seems like we could get a default alias from all the match arms and carry on with normal logic instead of having everything so deeply nested

tvc/src/commands/login.rs
```diff
@@ -292,34 +525,160 @@ fn build_login_plan_interactive(
...
+            }
+        }
+        _ => Ok(prompts::select(
+            &format!("Select profile for organization '{query}'"),
+            duplicate_alias_choices(config, &aliases),
+        )?
+        .alias
+        .to_string()),
+    }
+}
```
richardpringle: This function is all over the place. The match arms should be simplified

tvc/src/commands/login.rs
```diff
@@ -292,34 +525,160 @@ fn build_login_plan_interactive(
...
+         or run `tvc login` interactively to consolidate."
+    )
+}
+
+fn multiple_default_aliases_error(org_id: &str, marked_list: &str) -> anyhow::Error {
+    anyhow!(
+        "Multiple profiles are marked default_alias for organization '{org_id}': {marked_list}. \
+         Re-run `tvc profile set-default-alias --org <alias>` to leave exactly one."
+    )
+}
```
richardpringle: Don't do this. Make a struct with `thiserror@2.0.12` instead. It should be trivial to convert a `thiserror::Error` into an `anyhow::Error`
```rust
#[derive(displaydoc::Display, thiserror::Error)]
pub enum ProfileError {
    /// Multiple profiles are configured for organization '{org_id}': {alias_list}. Mark the one to keep with `tvc profile set-default-alias --org <alias>`, delete an extra with `tvc profile delete --org <alias> --yes`, or run `tvc login` interactively to consolidate.
    NoDefaultAlias { org_id: String, alias_list: String },

    /// Multiple profiles are marked default_alias for organization '{org_id}': {marked_list}. Re-run `tvc profile set-default-alias --org <alias>` to leave exactly one.
    MultipleDefaultAliases { org_id: String, marked_list: String },
}
```

tvc/src/commands/login.rs
```diff
@@ -292,34 +525,160 @@ fn build_login_plan_interactive(
...
                 &mut config,
                 &alias,
                 plan.api_base_url_override.as_deref(),
             );
-            let org_config = config.orgs.get(&alias).unwrap().clone();
+
+            let org_config = config
+                .orgs
+                .get(&alias)
+                .cloned()
```
richardpringle: ugh

tvc/src/commands/login.rs
```diff
@@ -292,34 +525,160 @@ fn build_login_plan_interactive(
...
+                .expect("plan alias was resolved against this config");
```
richardpringle: absolutely not, stop doing this

tvc/src/commands/login.rs
```diff
@@ -440,9 +800,35 @@ fn prompt_for_new_org_inputs(
...
     }
 
+    // One profile per organization: a second alias for the same ID would make
+    // resolution ambiguous again (TVC-159).
+    let already_configured = config
+        .orgs
+        .iter()
+        .filter(|(_, org)| org.id == id)
+        .map(|(alias, _)| alias)
+        .min();
```
richardpringle: Why? This is really dumb. why wouldn't you just use `find`. Or... if there's more than one configured org with the same ID, why aren't we prompting the user to just pick one and then we can merge the others

tvc/src/commands/login.rs
```diff
@@ -460,18 +846,23 @@ impl Display for OrgChoice {
...
-        if org_config.id == org {
-            return Some((alias, org_config));
-        }
+/// Every configured alias matching an org query, for callers to dispose of
+/// per their mode. An exact alias match is unambiguous and wins outright;
+/// otherwise all aliases registered for the query as an organization ID are
+/// returned, sorted for stable output.
+fn find_org_aliases(config: &Config, query: &str) -> Vec<String> {
+    if config.orgs.contains_key(query) {
+        return vec![query.to_string()];
```
richardpringle: Why would you do this? This is such a poor decision. We're not only allocating a new string but we're wrapping that in a `Vec`? I understand that you might need a vec, but everything else is borrowed and will live long enough. Put in _some_ effort to be efficient. `Vec<&'a str>` please

tvc/src/commands/login.rs
```diff
@@ -598,6 +990,56 @@ async fn find_or_generate_operator_key(
...
+                Some(&format!("operator-{org_alias}-backup.json")),
+            )?);
+
+            // Declining the overwrite skips the backup rather than failing:
+            // login has already succeeded and must not die here.
+            let proceed = !destination.exists()
+                || prompts::confirm(&format!("Overwrite {}?", destination.display()), false)?;
+
+            if proceed {
+                match backup_operator_key::back_up_key(org_alias, &local.key_path, &destination)
```
richardpringle: Why is there a stutter here. Why isn't this just `operator::back_up_key` or `operator_key::back_up`

tvc/src/commands/login.rs
```diff
@@ -697,6 +1139,26 @@ impl Display for ProfileDeleted {
     }
 }
 
+#[derive(Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DefaultAliasSet {
+    alias: String,
+    organization_id: String,
```
richardpringle: If an id is a UUID, the only time it should be represented as a string is when it is being converted in a display implementation or in (de)serialization. Parse, don't validate

tvc/src/commands/login.rs
```diff
@@ -812,4 +1298,30 @@ mod tests {
             extra: toml::Table::new(),
         }
     }
+
+    fn config_with_orgs(entries: &[(&str, &str)]) -> Config {
```
richardpringle: Pass in something that implements `IntoIterator<(&str, &str)>` so that maps, vecs, slices, arrays, and iterators can be used as an argument

tvc/tests/common.rs
richardpringle, line 1: use 2018 edition for new files. This should be `common.rs`

tvc/tests/common.rs
```diff
@@ -0,0 +1,105 @@
+//! Fixtures shared across the tvc integration-test binaries.
+//!
+//! Every test binary that declares `mod common;` compiles its own copy and
+//! uses a subset of these helpers, so unused items are expected.
+
+#![allow(dead_code)]
```
richardpringle, line 6: why?

tvc/tests/common/mod.rs
```diff
@@ -0,0 +1,105 @@
+//! Fixtures shared across the tvc integration-test binaries.
+//!
+//! Every test binary that declares `mod common;` compiles its own copy and
+//! uses a subset of these helpers, so unused items are expected.
+
+#![allow(dead_code)]
+
+use std::collections::HashMap;
+use std::fs;
+use std::path::{Path, PathBuf};
```
richardpringle: merge imports

tvc/tests/common/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+fn org_dir(home: &Path, alias: &str) -> PathBuf {
+    home.join(".config/turnkey/orgs").join(alias)
+}
+
+/// Write a v1 `tvc.config.toml` under `home` with one profile per
+/// `(alias, org_id)` pair, using the default per-alias key-file layout and a
+/// dead-port API base URL.
+pub fn write_profiles_config(home: &Path, profiles: &[(&str, &str)], active_org: Option<&str>) {
+    write_profiles_config_with_defaults(home, profiles, active_org, &[]);
+}
```
richardpringle: this is stupid, just call the inner function inline

tvc/tests/common/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+pub fn write_profiles_config_with_defaults(
+    home: &Path,
+    profiles: &[(&str, &str)],
+    active_org: Option<&str>,
+    default_aliases: &[&str],
+) {
+    let turnkey_dir = home.join(".config/turnkey");
+    fs::create_dir_all(&turnkey_dir).unwrap();
+
+    let orgs: HashMap<_, _> = profiles
```
richardpringle: How is the HashMap seeded? It would be bad if these things were always being written in different orders. An `IndexMap` would probably be better. Maybe just add a TODO, since we don't want to increase scope

Cargo.toml
```diff
@@ -78,7 +79,7 @@ x509-parser = { version = "0.14.0", default-features = false }
 
 # Serialization formats
 borsh = { version = "1.0", features = ["std", "derive"], default-features = false }
-toml = { version = "0.8", default-features = false, features = ["parse", "display"] }
+toml = { version = "0.8", default-features = false, features = ["parse", "display", "preserve_order"] }
```
richardpringle: 🤔 this feature should probably only be turned on in the cli

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,268 @@
+//! Operator key backup command - copies a local operator key file to a
+//! user-chosen destination.
+
+use crate::commands::Run;
+use crate::commands::login::resolve_org_query;
+use crate::config::turnkey::{Config, OrgQuery, StoredQosOperatorKey};
+use crate::outcome::Outcome;
+use crate::output::StdCtx;
+use crate::prompts::{self, error_required_in_non_interactive};
```
richardpringle: merge imports

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,268 @@
...
+use anyhow::{Context, Result, anyhow, bail};
+use clap::Args as ClapArgs;
+use serde::Serialize;
+use std::fmt::{self, Display, Formatter};
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
```
richardpringle: merge imports

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// Organization alias or ID whose operator key to back up.
+    /// Defaults to the active organization.
+    #[arg(long, env = "TVC_ORG", value_name = "ORG", value_parser = OrgQuery::from_str)]
+    org: Option<OrgQuery>,
+    /// Destination file for the backup copy.
+    #[arg(short, long, value_name = "PATH", env = "TVC_OPERATOR_KEY_BACKUP_OUT")]
+    output: Option<PathBuf>,
+    /// Overwrite the destination if it already exists.
+    #[arg(long)]
+    force: bool,
```
richardpringle: Is this type of option used anywhere else? Are all use-cases consistent? (could be `--force` or `--overwrite`)

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+#[derive(Default, Serialize)]
+#[cfg_attr(test, derive(Debug))]
+#[serde(rename_all = "camelCase")]
+pub struct OperatorKeyBackedUp {
+    alias: String,
+    public_key: String,
+    source_path: String,
+    backup_path: String,
```
richardpringle: These should be parsed types instead of plain strings. The same is true for `StoredQosOperatorKey` (where the `public_key` comes from)

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+impl OperatorKeyBackedUp {
+    /// The paths are projected to display strings here: the payload is a
+    /// serialization shape, not a working set of paths.
+    pub(crate) fn new(alias: String, public_key: String, source: &Path, backup: &Path) -> Self {
```
richardpringle: You shouldn't be able to construct an `Outcome` type outside of the file where it's defined. Why doesn't the `back_up` function just produce `OperatorKeyBackedUp`. Then the `back_up` call in the `login` flow can have a field that's `Option<OperatorKeyBackedUp>`.

Further, these (`source` and `backup` should be owned and parsed types if owned types is what we need)

tvc/src/cli.rs
```diff
@@ -299,6 +303,7 @@ impl Commands {
             },
             Commands::Profile { command } => match command {
                 ProfileCommands::Delete(_) => "profile delete",
+                ProfileCommands::SetDefaultAlias(_) => "profile set-default-alias",
```
richardpringle: Question, is there no way to get this string out of the `Commands` type instead of manually doing this?

tvc/src/cli.rs
```diff
@@ -414,6 +423,7 @@ impl AppCommands {
 impl KeysCommands {
     fn name(&self) -> &'static str {
         match self {
+            KeysCommands::BackupOperatorKey(_) => "keys backup-operator-key",
```
richardpringle, line 425: same here

tvc/src/operator.rs
```diff
@@ -500,7 +500,7 @@ fn find_hosted_operator(
     match matches.as_slice() {
         [] => Ok(None),
         [(name, record)] => Ok(Some((
-            org.id.clone(),
+            org.id.to_string(),
```
richardpringle: Are we doing this because this is now a `UUID`? Why not continue propagating the proper parsed type

tvc/tests/common.rs
richardpringle: There should be a comment above this annotation describing why it's needed

tvc/src/commands/operator/create.rs
```diff
@@ -100,11 +100,11 @@ pub async fn run(_ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
     let mut config = Config::load().await?;
     let (alias, configured_org_id) = config
         .active_org_config()
-        .map(|(alias, org)| (alias.clone(), org.id.as_str()))
+        .map(|(alias, org)| (alias.clone(), org.id.to_string()))
```
richardpringle: this seems wrong... Is this because it's a UUID? Why doesn't the `ensure_authenticated_org` take in a UUID instead of a string? I understand limiting the scope of the change, but I think propagating the change is more beneficial. When I say "X should be a UUID", that means it should carry that type information from "edge to edge" meaning, input to output.

tvc/src/config/turnkey/mod.rs
```diff
@@ -14,11 +17,13 @@ pub use api_key::{KeyCurve, StoredApiKey};
...
+use indexmap::IndexMap;
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
+use std::convert::Infallible;
 use std::fmt::{self, Display, Formatter};
-use std::path::Path;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
```
richardpringle: merge imports

tvc/src/config/turnkey/mod.rs
```diff
@@ -41,9 +46,11 @@ pub struct Config {
     /// The currently active organization alias
     #[serde(default)]
     pub active_org: Option<String>,
-    /// Map of org alias -> org config
+    /// Map of org alias -> org config, in config-file order. The order is
+    /// preserved so saves don't churn the file and so "first profile in the
+    /// config" is well-defined for default-alias repair.
```
richardpringle:
```suggestion
    /// Map of org alias -> org config, in config-file order.
```

tvc/src/config/turnkey/mod.rs
```diff
@@ -325,6 +340,12 @@ pub struct OrgConfig {
     /// Durable local and hosted operator metadata.
     #[serde(default)]
     pub operators: Vec<OperatorRecord>,
+    /// Marks the profile that org-ID resolution follows while several profiles
+    /// share this organization ID ("default_alias", not "default": profiles are
+    /// commonly literally aliased `default`). Only meaningful — and only
+    /// serialized — while such duplicates exist.
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub default_alias: bool,
```
richardpringle: Just marking this for myself.... I'm not convinced this should be `pub`. It might need a getter. It's also possible to hack this so that we don't have to turn every field into a getter by having one field that's not `pub` set to a ZST that's also not `pub`..

tvc/src/config/turnkey/mod.rs
```diff
@@ -363,12 +384,59 @@ impl OrgConfig {
...
+        self.api_key_path == dir.join(API_KEY_FILE)
+            && self
+                .operators
+                .iter()
+                .filter_map(|operator| match &operator.kind {
+                    OperatorRecordKind::Local(local) => Some(&local.key_path),
+                    _ => None,
+                })
+                .all(|key_path| *key_path == operator_key)
+    }
```
richardpringle, line 634: This seems fragile. We're suggesting that the layout is default by checking that all the local operator keys exist in a specific place? This seems inherently wrong

tvc/src/config/turnkey/mod.rs
```diff
@@ -468,15 +543,110 @@ impl Config {
...
+                .map(|(alias, org)| (alias.as_str(), org))
+                .into_iter()
+                .collect(),
+            OrgQuery::Id(id) => self
+                .orgs
+                .iter()
+                .filter(|(_, org)| org.id == *id)
+                .map(|(alias, org)| (alias.as_str(), org))
+                .collect(),
+        }
```
richardpringle: I think there's a pattern that's being used here that's wrong. It's because of the way the `Config` has maps keyed off of alias. Nothing should be keyed off of alias. We might need to go back to the drawing board here. Aliases should only exist at the edges (inputs and outputs).

tvc/src/config/turnkey/mod.rs
```diff
@@ -468,15 +543,110 @@ impl Config {
...
+                |mut groups, (alias, org)| {
+                    groups.entry(org.id).or_default().push(alias);
+                    groups
+                },
+            )
+            .into_iter()
+            .filter(|(_, aliases)| aliases.len() > 1)
+            .map(|(id, aliases)| (id, aliases.into_iter().map(String::from).collect()))
+            .collect()
+    }
```
richardpringle, line 827: If we do the above, this disappears

tvc/src/config/turnkey/mod.rs
```diff
@@ -468,15 +543,110 @@ impl Config {
...
+            .iter()
+            .map(|(alias, org)| {
+                let legacy = legacy_org_dir(alias)?;
+                let migrates =
+                    legacy != default_org_dir(org.id)? && org.has_default_layout_at(&legacy);
+                Ok(migrates.then(|| (alias.clone(), org.id)))
+            })
+            .filter_map(Result::transpose)
+            .collect()
+    }
```
richardpringle, line 837: 🤔 do we really need this? In theory, you should be able to put operator keys anywhere. This doesn't seem right

tvc/src/config/turnkey/mod.rs
```diff
@@ -468,15 +543,110 @@ impl Config {
...
+
+    /// Make `winner` the only profile marked `default_alias` among those
+    /// registered for `org_id`.
+    pub(crate) fn mark_sole_default_alias(&mut self, org_id: Uuid, winner: &str) {
+        self.orgs
+            .iter_mut()
+            .filter(|(_, org)| org.id == org_id)
+            .for_each(|(alias, org)| org.default_alias = alias == winner);
+    }
```
richardpringle, line 837: This goes away with the other comment further up

tvc/src/config/turnkey/mod.rs
```diff
@@ -468,15 +543,110 @@ impl Config {
...
+                    .find(|alias| self.orgs.get(*alias).is_some_and(|org| org.default_alias))
+                    .or(aliases.first())?;
+                Some((org_id, winner.clone()))
+            })
+            .collect();
+
+        winners
+            .into_iter()
+            .for_each(|(org_id, winner)| self.mark_sole_default_alias(org_id, &winner));
+    }
```
richardpringle, line 843: This also goes away with the other comment further up

tvc/src/config/turnkey/mod.rs
```diff
@@ -106,7 +113,10 @@ mod disk {
     }
 
     pub(super) fn from_toml(content: &str) -> Result<Config> {
-        DiskConfig::from_toml(content)?.into_current()
+        let mut config = DiskConfig::from_toml(content)?.into_current()?;
+        // Like the v0 migration, repairs are lazy: the next save persists them.
+        config.normalize_default_aliases();
```
richardpringle: I think this should also disappear. But we also need this to be fault tolerant here, I think. This is where we could to something like save the original file as `.backup` if and only if we're migrating

tvc/src/config/turnkey/mod.rs
richardpringle: Seems like this is going to disappear if other comments are followed

tvc/src/config/turnkey/mod.rs
```diff
@@ -491,13 +661,28 @@ impl Config {
...
+        // removed profile's group to a single survivor.
+        let mut survivors: Vec<_> = self
+            .orgs
+            .values_mut()
+            .filter(|org| org.id == removed.id)
+            .collect();
+
+        if let [survivor] = survivors.as_mut_slice() {
+            survivor.default_alias = false;
+        }
```
richardpringle: This should disappear with the other comment

tvc/src/commands/login.rs
```diff
@@ -1,10 +1,11 @@
 //! Login command for authenticating with Turnkey.
 
 use crate::client::build_turnkey_client;
+use crate::commands::keys::backup_operator_key::{OperatorKeyBackedUp, back_up};
```
richardpringle: merge imports

tvc/src/commands/login.rs
```diff
@@ -17,18 +18,22 @@ use serde::Serialize;
 use std::collections::BTreeSet;
 use std::fmt::{self, Display, Formatter};
 use std::io::BufRead;
+use std::path::PathBuf;
+use std::str::FromStr;
```
richardpringle: merge imports

tvc/src/commands/login.rs
```diff
@@ -72,17 +93,168 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
...
+) -> Result<()> {
+    for (alias, org_id) in profiles {
+        let source = legacy_org_dir(&alias)?;
+        let target = default_org_dir(org_id)?;
+
+        match tokio::fs::rename(&source, &target).await {
+            Ok(()) => {}
+            // A crash between a previous run's rename and its save leaves the
+            // files already moved; treat the move as done and just rewrite.
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound && target.is_dir() => {}
```
richardpringle, line 133: I don't understand this failure mode. Why would a rename lead to a `NotFound` error?

tvc/src/commands/login.rs
```diff
@@ -147,9 +317,25 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<Outcome> {
...
+/// Permanently delete profile `alias`: remove it from the config registry,
+/// delete its key files when they use the default per-org directory layout
+/// (custom key paths are left on disk with a warning), then save the config.
+/// The save comes last so that dying between the file cleanup and the save
+/// leaves the profile listed and the deletion retryable; the file deletion
+/// tolerates an already-missing directory for exactly that retry.
+async fn delete_profile(
+    ctx: &mut StdCtx,
+    config: &mut Config,
+    alias: String,
```
richardpringle: Just making sure that something like this won't exist after the bigger refactor

tvc/src/commands/login.rs
```diff
@@ -160,188 +346,420 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<Outcome> {
...
-            Ok(()) => Some(dir),
-            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+    // The default layout stores both key files in one per-org directory —
+    // id-keyed today, alias-keyed for profiles created before TVC-55 — so a
+    // default profile is removed by deleting that directory. Custom
+    // (hand-edited) key paths are left untouched with a warning, since the
+    // user placed them deliberately and they may live outside our config tree.
+    let owned_dir = [default_org_dir(removed.id)?, legacy_org_dir(&alias)?]
+        .into_iter()
+        .find(|dir| removed.has_default_layout_at(dir));
```
richardpringle, line 299: This is a fragile way of checking for defaults. Is this because the key would be the only thing that's in the directory? 

We shouldn't be checking for the default directory, that doesn't make sense. We should be checking if the absolute path is somewhere that's about to be moved and fixing after the move. Only deleting once we've confirmed the new state. I think that means we shouldn't be doing any renames because we can end up in invalid states.

tvc/src/commands/login.rs
richardpringle: Definitely pass the config in from the outside of any free function

richardpringle (commented):
More changes needed

tvc/src/commands/app/create.rs
```diff
@@ -97,21 +106,21 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
...
+/// Decide whether to reuse a known operator for the manifest set.
 ///
 /// Pure and mode-agnostic: it performs no I/O and knows nothing about
 /// interactivity. The caller resolves [`OperatorReuse::MultipleCandidates`] by
 /// prompting (interactive) or erroring (non-interactive).
 fn decide_operator_reuse(
     no_reuse: bool,
     manifest_set_params: Option<&OperatorSetParams>,
-    saved_ids: &[String],
+    candidates: &[OperatorCandidate],
```
emostov, line 126: Lets update the agents.md to say to prefer to not use strings and parse strings at the edge

tvc/src/commands/deploy/approve.rs
```diff
@@ -403,9 +403,26 @@ impl TryFrom<Args> for ArgsWithResolvedOperatorSeedSource {
...
+}
+
+impl fmt::Display for ApprovingOperator {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match &self.name {
+            Some(name) => write!(f, "{name} ({})", self.id),
+            None => fmt::Display::fmt(&self.id, f),
+        }
+    }
+}
```
richardpringle, line 420: I think I abstracted this somewhere else

tvc/src/commands/deploy/init.rs
```diff
@@ -26,20 +27,32 @@ placeholder template is written."#;
...
-#[command(about, long_about = None)]
+#[command(
+    about,
+    long_about = None,
+    group(
+        ArgGroup::new("app_id_source")
+            .args(["app_id", "from_deployment"])
+            .multiple(false)
+    )
+)]
```
richardpringle, line 38: This is a real change. Instead of having the `DeployConfig` potentially have a placeholder, we enforce specifying an `app-id`. 

Why? 

Deployments don't exist without an app, every deployment has a `belongs-to` relationship with an app. Sure there are other required fields, but they can be unique to a deployment.

Before we were filling in the app-id for the user by using the last app-id... I believe this implicit behaviour could lead to confusion in the case where someone has many live apps. It's better to have the user copy the id and always supply the argument. 

Please feel free to disagree! 

I spent a long time thinking about this, but you (dear reviewer) may still have insights that  I didn't think of!

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,295 @@
...
+Source:     {}
+Backup:     {}
+
+The backup contains the PRIVATE key. Store it somewhere safe - a password
+manager or an encrypted offline drive - never in source control or chat.
+
+To restore: copy the backup file back to the source path above, then run
+`tvc login`."#,
+            self.alias,
+            self.public_key,
```
richardpringle, line 187: Double check how this displays

tvc/src/commands/login.rs
```diff
@@ -73,19 +94,81 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
...
+async fn migrate_legacy_key_directories(
+    ctx: &mut StdCtx,
+    config: &mut Config,
+    profiles: Vec<(String, Uuid)>,
+) -> Result<()> {
+    for (alias, org_id) in profiles {
+        let source = legacy_org_dir(&alias)?;
+        let target = default_org_dir(org_id)?;
+
+        match tokio::fs::rename(&source, &target).await {
```
richardpringle, line 129: I don't think this is the best way to handle this... It needs to be more durable. We need both copies to exist with a `.backup` at the end or something. Otherwise we could end up in a broken state

tvc/src/commands/login.rs
```diff
@@ -161,227 +288,310 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<Outcome> {
...
-struct ProfileChoice<'a> {
-    alias: &'a str,
-    org_id: &'a str,
-    is_active: bool,
+/// One picker row for an organization: its aliases (or bare ID) plus the
+/// active marker. `name` carries the first alias for output echoing.
+struct OrgRow {
+    id: Uuid,
+    name: Option<String>,
+    display: String,
```
richardpringle, line 382: This looks like a code-smell... carrying around a `String` just to "maybe" display it? 

...I'll have to see the usage

richardpringle: If you look at the callsite, it would be better to carry around the `names` `Vec`

Is this just meant for migrating?

tvc/src/commands/login.rs
```diff
@@ -107,31 +190,50 @@ pub async fn run_delete(ctx: &mut StdCtx, args: DeleteArgs) -> Result<Outcome> {
...
+            )
+        }
+        None => {
+            // Reached only in interactive mode; a non-interactive run without
+            // --org is rejected up front before we get here.
+            if config.orgs.is_empty() {
+                bail!("No organization logins to delete.");
+            }
+
+            let row = prompts::select("Select organization to delete", org_rows(&config))?;
```
richardpringle, line 210: Need to test this manually with multiple aliases

tvc/src/commands/login.rs
richardpringle, line 1: I think a lot needs to be tested here

tvc/src/config/turnkey/mod.rs
```diff
@@ -1,24 +1,37 @@
...
+//!   the alias map, active org, and key paths
+//! - `orgs/<org-id>/api_key.json` - Default location for API keys
+//! - `orgs/<org-id>/operator.json` - Default location for operator keys
 //!
-//! Key paths are stored in the config so users can customize storage locations.
+//! Aliases are strictly an edge concept: the serialized `[aliases]` table maps
+//! human names to organization UUIDs, and nothing else in the config is keyed
+//! by name. Key paths are stored in the config so users can customize storage
+//! locations; directories keyed by the legacy `orgs/<alias>/` layout remain
+//! readable and are migrated by interactive login.
```
richardpringle, line 13: This should be the only migration path

tvc/src/config/turnkey/mod.rs
```diff
@@ -27,48 +40,193 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
-    /// The currently active organization alias
+    /// The currently active organization.
+    #[serde(default)]
+    pub active_org: Option<Uuid>,
+    /// Org registry keyed by organization ID, in config-file order.
     #[serde(default)]
-    pub active_org: Option<String>,
-    /// Map of org alias -> org config
+    pub orgs: IndexMap<Uuid, OrgConfig>,
+    /// Edge-side names for organizations.
```
richardpringle: What does "edge-side" mean

tvc/src/config/turnkey/mod.rs
```diff
@@ -27,48 +40,193 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
     #[serde(default)]
-    pub orgs: HashMap<String, OrgConfig>,
-    /// Map of org alias -> last created app ID (for convenience)
+    pub aliases: Aliases,
+    /// Last created app ID per organization (for convenience).
     #[serde(default)]
-    pub last_created_app_id: HashMap<String, String>,
-    /// Map of org alias -> last manifest set operator IDs (for convenience)
+    pub last_created_app_id: HashMap<Uuid, Uuid>,
+    /// Last manifest set operator IDs per organization (for convenience).
```
richardpringle, line 67: I may have made this irrelevant...

See my other comment about being explicit with app-ids

tvc/src/config/turnkey/mod.rs
```diff
@@ -27,48 +40,193 @@ const CONFIG_FILE: &str = "tvc.config.toml";
...
     #[serde(default)]
-    pub last_operator_ids: HashMap<String, Vec<String>>,
+    pub last_operator_ids: HashMap<Uuid, Vec<Uuid>>,
     /// Unrecognized top-level fields retained across supported config rewrites.
     #[serde(default, flatten)]
     pub extra: toml::Table,
 }
 
+/// Edge-side names for UUID-identified resources.
```
richardpringle: `edge-side` doesn't mean anything to people. "user-facing" would probably be better

tvc/src/config/turnkey/mod.rs
richardpringle, line 1: I don't think there are enough tests for migration. Migrating from v0 -> v2 and v1 -> v2. What happens when there are interruptions mid-migration? 

This needs better testing

tvc/src/config/turnkey/qos_operator_key.rs
```diff
@@ -2,14 +2,100 @@
...
+    ($ty:ident) => {
+        impl Debug for $ty {
+            fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+                write!(f, concat!(stringify!($ty), "({})"), hex::encode(self.0))
+            }
+        }
+    };
+}
+
+impl_hex_debug!(QosOperatorPublicKey);
```
richardpringle, line 81: Is this used anywhere else?

tvc/src/config/placeholder.rs
richardpringle, line 1: handwritten, don't roast me

tvc/src/config/placeholder.rs
```diff
@@ -0,0 +1,676 @@
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let borrowed = Cow::<'de, str>::deserialize(deserializer)?;
+
+        let mut chars = borrowed.chars();
+
+        match (chars.next(), chars.last()) {
+            (Some('<'), Some('>')) => Err(D::Error::custom(
+                "String be surrounded by placeholder brackets '<' and '>'",
```
richardpringle, line 197:
```suggestion
                "String cannot be surrounded by placeholder brackets '<' and '>'",
```

tvc/src/config/placeholder.rs
```diff
@@ -0,0 +1,676 @@
...
+    }
+}
+
+impl<P> From<&'static str> for StringWithPlaceholder<P> {
+    fn from(value: &'static str) -> Self {
+        NonPlaceHolderString::from(value).into()
+    }
+}
+
+impl<'de> Deserialize<'de> for NonPlaceHolderString {
```
richardpringle, line 186: Should leave a comment here. What happens in the deserialization implementation paired with the above is that first `NonPlaceHolderString` attempts to deserialize. If surrounding `<` and `>` are detected, serialization fails, and the derived untagged enum deserialization continue with the `Placeholder` variant (which does the actual check). 

If we had specialization, we would do this different (and make sure that the string is not an exact match), alas we do not and therefore do not have the exact match information here.

tvc/src/operator/mod.rs
```diff
@@ -0,0 +1,550 @@
...
+};
+
+use crate::approvals::ValidatedManifest;
+use crate::client::build_client;
+use crate::config::turnkey::{Config, OperatorKind, OperatorRecordKind, StoredQosOperatorKey};
+use crate::local_operator_key::{
+    LocalOperatorSeedSource, resolve_local_operator, resolve_registered_local_operator,
+    select_local_operator,
+};
+use crate::pair::Signer;
```
richardpringle, line 24: merge

tvc/src/operator/hosted.rs
```diff
@@ -550,19 +288,43 @@ pub(crate) fn resolve_hosted_operator_encrypt_key(
...
+    /// Assemble a hosted signer from its finished dependencies. The caller
+    /// has already verified that `auth` is authenticated against the
+    /// operator's organization.
+    pub(super) fn new(operator: ResolvedHostedOperator, auth: AuthenticatedClient) -> Self {
+        Self { operator, auth }
+    }
+}
+
+impl Signer for HostedSigner {
+    fn sign(&self, message: Vec<u8>) -> SignerFuture<'_, Result<Vec<u8>>> {
```
richardpringle: message should be borrowed here. The intent should be created before the async context then moved into it.

tvc/src/operator/hosted.rs
```diff
@@ -0,0 +1,689 @@
...
+    name: String,
+    wallet: HostedOperatorWallet,
+    path: String,
+}
+
+/// Valid wallet selections for hosted operator creation.
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum HostedOperatorWallet {
+    New(String),
+    Existing(Uuid),
```
richardpringle, line 45: Need docs for the variants

tvc/src/operator/hosted.rs
```diff
@@ -0,0 +1,689 @@
...
+    pub(crate) fn encrypt_public_key(&self) -> &OperatorPublicKey {
+        &self.encrypt_public_key
+    }
+
+    pub(crate) fn sign_public_key(&self) -> &OperatorPublicKey {
+        &self.sign_public_key
+    }
+
+    pub(crate) fn composite_public_key(&self) -> Vec<u8> {
+        let mut composite = Vec::with_capacity(130);
```
richardpringle, line 160: Should be a constant defined at the top of the file

tvc/src/operator/hosted.rs
```diff
@@ -0,0 +1,689 @@
...
+    match (hosted.next(), hosted.next()) {
+        (Some(sole), None) => Ok(sole),
+        (None, _) => Err(SelectHostedOperatorError::NoHostedOperator),
+        (Some(_), Some(_)) => Err(SelectHostedOperatorError::MultipleHostedOperators),
+    }
+}
+
+/// Find the hosted registry record with `operator_id` in the active
+/// organization, if any, validated for use.
+pub(super) fn find_hosted_operator(
```
richardpringle, line 241: Why wouldn't this just be a method on the org? 
Then you could do something like: 

```rust
let org = // get active org, handle errors in the context of the call-site, maybe prompt for org-id
let operators = org.hosted_operators()
// match on the single value or prompt for selection
```

For different call sites, it may or may not make sense to prompt the user to select a specific operator. We're violating layered architecture principles here

tvc/src/operator/mod.rs
```diff
@@ -0,0 +1,550 @@
...
+    Ok(ResolvedOperator {
+        name: Some(operator.name.clone()),
+        operator_id: resolved_operator_id,
+        signer: Box::new(resolve_registered_local_operator(local.key_path.clone()).await?),
+    })
+}
+
+fn parse_uuid(value: &str, field: &str) -> Result<Uuid> {
+    Uuid::parse_str(value).map_err(|_| anyhow!("{field} must be a UUID"))
+}
```
richardpringle, line 289: whyyyyy this should be inlined at all call-sites

tvc/src/operator/mod.rs
```diff
@@ -0,0 +1,550 @@
...
+
+/// Best-effort read of the active org's operator public key under its
+/// default backend, as qos composite hex, for prompt and template defaults.
+/// The local backend reads the sole registered key file; the hosted backend
+/// joins the sole record's stored points. `None` on any miss.
+pub(crate) async fn default_operator_public_key() -> Option<String> {
```
richardpringle, line 295: This should be a method on `Config`

tvc/src/operator/mod.rs
richardpringle, line 1: This file is poorly written and has too many free functions. Some of these functions would be better as methods on the `Config` type. We shouldn't be calling `Config::load` at all in this file

tvc/src/pair.rs
```diff
@@ -77,37 +80,35 @@ impl LocalPair {
...
         })
     }
-}
 
-impl Pair for LocalPair {
-    fn sign(&self, message: Vec<u8>) -> PairFuture<'_, anyhow::Result<Vec<u8>>> {
+    /// Decrypt the given ciphertext. Decryption needs the private key
+    /// material, so it lives on the concrete local pair rather than
+    /// [`Signer`].
+    pub async fn decrypt(&self, ciphertext: Vec<u8>) -> anyhow::Result<Zeroizing<Vec<u8>>> {
```
richardpringle, line 87: ciphertext should be borrowed

tvc/tests/deploy_approve.rs
```diff
@@ -181,68 +204,68 @@ fn explicit_seed_does_not_load_malformed_config() {
...
+//         .arg("fixtures/manifest.json")
+//         .arg("--manifest-id")
+//         .arg("11111111-1111-4111-8111-111111111111")
+//         .arg("--dangerous-skip-interactive")
+//         .assert()
+//         .failure()
+//         .stderr(predicate::str::contains(
+//             "saved operator ID 'not-a-uuid' is not a UUID",
+//         ));
+// }
```
richardpringle, line 243: Deleteme, operator ids are uuids now, right?

PR #226, opened 2026-07-30
feat(tvc): offer operator key backup during login onboarding

richardpringle:
## Backup nudge for fresh operator keys

Stacked on #225. When login generates a fresh operator key interactively, it now explains the key exists only on this machine and offers a backup (default Yes) via the shared `back_up_key` core — destination prompt, overwrite confirm (decline skips, doesn't fail login), and the full backed-up report. Backup errors degrade to a warning: login has already succeeded and must not die here. Declining (or any skip) prints the pointer to `tvc keys backup-operator-key`; re-logins with an existing key get a single tip line, no prompts.

The plan's "login hook isn't e2e-testable" gap turned out to be closable: a one-shot mock whoami server (the `error_output.rs` pattern) carries PTY login past credential verification, so the accept path (byte-identical backup on disk), decline path, and existing-key tip are all covered end-to-end.

Closes TVC-53.

### DIFF
--- a/tvc/src/commands/login.rs
+++ b/tvc/src/commands/login.rs
@@ -1,6 +1,7 @@
 //! Login command for authenticating with Turnkey.
 
 use crate::client::build_turnkey_client;
+use crate::commands::keys::backup_operator_key;
 use crate::config::turnkey::{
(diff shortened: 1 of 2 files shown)


richardpringle:
Folded into #224 — everything ships on the tvc-159 branch.

PR #228, opened 2026-07-31
feat(tvc): show a clear upgrade hint when the backend rejects an outdated client

richardpringle:
Follows #218 (X-TVC-CLIENT-VERSION stamping + `version` subcommand). Handles the backend's rejection of an outdated client:

- Classifies the `TVC_CLIENT_VERSION_TOO_OLD` rejection as `client_version_too_old` and renders a terse `error: `tvc` version too old` with an upgrade `hint:`.
- Moves the raw HTTP detail to a `debug!` at the CLI boundary (recoverable via `RUST_LOG=tvc=debug`), keeping it out of the human error line.
- Adds `#[instrument(skip_all)]` spans across the client boundary and API-hitting commands for a clean debug call path.

Closes ENG-4082.

### DIFF
--- a/tvc/src/errors.rs
+++ b/tvc/src/errors.rs
@@ -9,7 +9,7 @@
 //! before they cross the CLI output boundary.
 
 use reqwest::StatusCode;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use turnkey_client::TurnkeyClientError;
 
 /// Cap on rendered error messages. Large enough for any real API error body
@@ -59,6 +59,9 @@ pub enum ErrorCode {
     NotFound,
     /// Any other non-success HTTP status, or a failed/unexpected activity.
     ApiError,
+    /// The backend rejected this tvc release as older than the minimum
+    /// version it supports. Remediation: upgrade tvc.
+    ClientVersionTooOld,
     /// An activity needs more approvals.
     ApprovalRequired,
     /// A connect/timeout/DNS failure — the request never reached the server.
@@ -98,6 +101,25 @@ pub fn classify(error: &anyhow::Error) -> Classification {
     Classification::new(ErrorCode::CommandError, None)
 }
 
+/// A remediation hint for recognized error causes, rendered by the human
+/// output layer beneath the error line. Today only the backend's
+/// client-version rejection produces one: the hint is the server's own
+/// message (it names the running and minimum versions), so the reader does
+/// not have to dig it out of the raw response body in the chain.
+pub fn hint(error: &anyhow::Error) -> Option<String> {
+    error.chain().find_map(|cause| {
+        let envelope = client_version_rejection(cause.downcast_ref::<TurnkeyClientError>()?)?;
+
+        Some(if envelope.message.is_empty() {
+            "this tvc release is older than the minimum version the backend supports; \
+             upgrade tvc to the latest release"
+                .to_string()
+        } else {
+            envelope.message
+        })
+    })
+}
+
 /// Render an error's complete cause chain like anyhow's alternate
 /// `{error:#}` display (links joined with `": "`), then truncate the result to
 /// [`MAX_ERROR_MESSAGE_CHARS`].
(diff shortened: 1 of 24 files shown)


tvc/src/errors.rs
```diff
@@ -133,18 +168,58 @@ fn truncate_message(message: String) -> String {
...
+/// [`hint`] so they cannot disagree on what counts as one.
+fn client_version_rejection(error: &TurnkeyClientError) -> Option<ApiErrorEnvelope> {
+    let TurnkeyClientError::UnexpectedHttpStatus(_, body) = error else {
+        return None;
+    };
+
+    serde_json::from_str::<ApiErrorEnvelope>(body)
+        .ok()
+        .filter(|envelope| envelope.turnkey_error_code == TVC_CLIENT_VERSION_TOO_OLD)
+}
```
daniilrrr, line 206: you added a new TurnkeyErrorCode to the proto in mono https://github.com/tkhq/mono/pull/7433/changes#diff-9fef28f00e081664b0755492c653dbb674b3ddc653df1471a908d607728c4c71R52

there should be a way to sync this new TVC error into our code, right? 

...

ah it looks like we don't currently consume `proto/external/errors/v1/errors.proto` into codegen. So that would be a bigger lift than is necessary here to parse one new error.

LLM tells me we could at least parse the unique error code from the proto as part of codegen and match on that, instead of the string. Up to you if it's worth doing here:
___
## 1. Generate only TurnkeyErrorCode

This is the smallest and least disruptive approach.

The proto is already synced, but codegen does not compile it because codegen/src/main.rs:59 only names public_api.proto.

Add it as a codegen root:

```rust
const PUBLIC_API_PROTO_PATH: &str =
    "proto/services/coordinator/public/v1/public_api.proto";
const EXTERNAL_ERRORS_PROTO_PATH: &str =
    "proto/external/errors/v1/errors.proto";

tonic_build::configure()
    // existing configuration...
    .compile_protos(
        &[PUBLIC_API_PROTO_PATH, EXTERNAL_ERRORS_PROTO_PATH],
        &[INCLUDE_PROTO_PATH],
    )
    .unwrap();
```
That should generate:
```
generated::external::errors::v1::TurnkeyErrorCode
```
Then move raw-body interpretation into turnkey_client:
```rust
#[derive(Deserialize)]
struct ApiErrorEnvelope {
    #[serde(default)]
    turnkey_error_code: String,
}

impl TurnkeyClientError {
    pub fn turnkey_error_code(&self) -> Option<TurnkeyErrorCode> {
        let Self::UnexpectedHttpStatus(_, body) = self else {
            return None;
        };

        let envelope = serde_json::from_str::<ApiErrorEnvelope>(body).ok()?;

        TurnkeyErrorCode::from_str_name(&envelope.turnkey_error_code)
    }
}
```
TVC then uses the generated variant:
```rust
if error.turnkey_error_code()
    == Some(TurnkeyErrorCode::TvcClientVersionTooOld)
{
    // Render upgrade guidance.
}
```
Benefits:

- No duplicated "TVC_CLIENT_VERSION_TOO_OLD" constant.
- Proto sync automatically adds new Rust enum variants.
- No breaking change to TurnkeyClientError.
- Raw-body parsing is centralized in the client crate.

richardpringle: I don't think it's worth doing here. But I think it _is_ worth doing. I'll create a ticket

tvc/src/errors.rs
```diff
@@ -133,18 +168,60 @@ fn truncate_message(message: String) -> String {
     )
 }
 
+/// The `turnkeyErrorCode` the backend attaches when it rejects a tvc release
+/// older than the minimum version it supports (the client-version gate).
+/// Must match the `TurnkeyErrorCode` enum value name in Turnkey's public
+/// error proto.
+// TODO(TVC-269): generate `TurnkeyErrorCode` from the synced errors.proto and
+// match on the enum variant instead of this string.
+const TVC_CLIENT_VERSION_TOO_OLD: &str = "TVC_CLIENT_VERSION_TOO_OLD";
```
richardpringle, line 177: @daniilrrr FYI

daniilrrr (approved):

PR #229, opened 2026-08-03
chore(tvc): add parse-don't-validate to AGENTS.md

richardpringle:
Rework the "validate or resolve inputs once" bullet to name the principle and add a trigger for validation disguised as parsing (`is_*`/`check_*`/`verify_*`/`validate_*` names, `bool`/`Result<(), E>` returns).

### DIFF
--- a/tvc/AGENTS.md
+++ b/tvc/AGENTS.md
@@ -79,9 +79,16 @@ Default guidance for coding-agent runs in this repository.
   forward-compatible identifiers as strings or dedicated domain wrappers.
   Compare typed values internally and convert them to strings only when an
   external wire type requires it.
-- Validate or resolve inputs once, then pass a narrow validated type into
-  creation or transformation code. Prefer an infallible transformation after
-  validation over a helper that repeats validation deeper in the call stack.
+- Parse, don't validate: run the fallible check once, at the boundary, and
+  return a narrow domain type that carries the proof. Downstream code takes
+  that type and transforms it infallibly instead of repeating the check deeper
+  in the call stack. Watch for validation disguised as parsing: before writing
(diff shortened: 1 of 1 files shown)


tvc/AGENTS.md
```diff
@@ -79,9 +79,16 @@ Default guidance for coding-agent runs in this repository.
   forward-compatible identifiers as strings or dedicated domain wrappers.
   Compare typed values internally and convert them to strings only when an
   external wire type requires it.
-- Validate or resolve inputs once, then pass a narrow validated type into
-  creation or transformation code. Prefer an infallible transformation after
-  validation over a helper that repeats validation deeper in the call stack.
+- Parse, don't validate: run the fallible check once, at the boundary, and
```
daniilrrr, line 82: imo "parse don't validate" is too abstract for LLMs to implement effectively, and might require a Skill or more concrete "if this do that" scaffolding as you have done here. Or there is insufficient training data on it, idk 

maybe this will help, we'll see i suppose

daniilrrr (approved):

Avi-D-coder (approved):
Nice, I have been struggling with this generation of model over engineering everything like this.

They seem to need really explicit guidance like this to not be overly defensive. Kinda sad

tvc/AGENTS.md
richardpringle: It can regurgitate the blogpost principles extremely well. It's not too abstract, but isn't enough on its own. 

Fable seems to think that this will help. I'm also a little skeptical though. I definitely don't think it'll hurt. I have my own version of this that's waaaaaay more detailed but the style is too personal to force everyone's agents to use it

PR #231, opened 2026-08-05
Add replica count support to tvc deploy create

emostov:
## What

`tvc deploy create` now supports specifying a desired replica count, defaulting to 1:

- `--replicas <N>` flag / `TVC_REPLICAS` env var, overriding the config file like the other flags
- `replicas` field in the deploy config file; `tvc deploy init` templates include `"replicas": 1`
- The create intent always sends an explicit replica count (previously omitted, leaving it to the server default)
- A count of 0 is rejected client-side with a clear validation error

## Notes

- Config files written before the `replicas` field existed keep working (serde default of 1).
- Seeding a config from an existing deployment defaults to 1 replica since the fetched deployment doesn't carry its desired replica count.

## Testing

- `cargo test` (all suites pass), `cargo clippy --all-targets --all-features -D warnings` clean.
- New tests: replicas default, flag-overrides-file, zero rejection, intent wiring, backward-compat deserialization.

### DIFF
--- a/tvc/src/commands/deploy/create.rs
+++ b/tvc/src/commands/deploy/create.rs
@@ -142,6 +142,11 @@ struct Overrides {
     /// This is usually the port your server listens on for external requests.
     #[arg(long, env = "TVC_PUBLIC_INGRESS_PORT")]
     pub public_ingress_port: Option<u16>,
+
+    /// Desired replica count for the deployment. When omitted, the server
+    /// applies its default.
+    #[arg(long, env = "TVC_REPLICAS")]
+    pub replicas: Option<u32>,
 }
 
 struct ResolvedDeployInputs {
@@ -305,6 +310,9 @@ fn apply_overrides(config: &mut DeployConfig, overrides: &Overrides) {
     if let Some(v) = overrides.public_ingress_port {
         config.public_ingress_port = v;
     }
+    if let Some(v) = overrides.replicas {
+        config.replicas = Some(v);
+    }
 }
 
 fn invalid_deploy_config_error(errors: DeployConfigValidationErrors) -> anyhow::Error {
@@ -364,8 +372,7 @@ fn build_create_intent(
         health_check_type: deploy_config.health_check_type,
         health_check_port: deploy_config.health_check_port as u32,
         public_ingress_port: deploy_config.public_ingress_port as u32,
-        // Not yet configurable via the CLI; the server applies its default.
-        replicas: None,
+        replicas: deploy_config.replicas,
     }
 }
 
@@ -623,6 +630,54 @@ mod tests {
         assert!(resolved.pivot_args.is_empty());
         // Pull-secret placeholder cleared in flag-only mode.
         assert_eq!(resolved.pivot_container_encrypted_pull_secret, None);
+        // Replicas stays unset when not specified.
+        assert_eq!(resolved.replicas, None);
+    }
+
+    #[test]
+    fn replicas_flag_overrides_file_value() {
+        let file = write_config(&file_config()); // file leaves replicas unset
+        let overrides = Overrides {
+            replicas: Some(3),
+            ..Default::default()
+        };
+        let args = Args {
+            config_file: Some(file.path().to_path_buf()),
+            overrides,
+            ..Default::default()
+        };
+        let resolved = run_resolve(&args).unwrap();
+        assert_eq!(resolved.replicas, Some(3));
+    }
+
+    #[test]
+    fn zero_replicas_is_rejected() {
+        let file = write_config(&file_config());
+        let overrides = Overrides {
+            replicas: Some(0),
+            ..Default::default()
+        };
+        let args = Args {
+            config_file: Some(file.path().to_path_buf()),
+            overrides,
+            ..Default::default()
+        };
+        let err = run_resolve(&args).unwrap_err();
+        assert!(err.to_string().contains("replicas"), "{err}");
+    }
+
+    #[test]
+    fn create_intent_carries_replica_count() {
+        let mut config = file_config();
+        config.replicas = Some(4);
+        let intent = build_create_intent(&config, "image".into(), None);
+        assert_eq!(intent.replicas, Some(4));
+    }
+
+    #[test]
+    fn create_intent_omits_replicas_when_unset() {
+        let intent = build_create_intent(&file_config(), "image".into(), None);
+        assert_eq!(intent.replicas, None);
     }
 
     #[test]
--- a/tvc/src/config/deploy.rs
+++ b/tvc/src/config/deploy.rs
@@ -44,6 +44,10 @@ pub struct DeployConfig {
     pub health_check_type: TvcHealthCheckType,
     pub health_check_port: u16,
     pub public_ingress_port: u16,
+    /// Desired replica count for the deployment. When unset, the create
+    /// request omits it and the server applies its default.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub replicas: Option<u32>,
 }
 
 /// Build a config seeded from an existing deployment, for use as a template.
@@ -109,6 +113,9 @@ impl TryFrom<TvcDeployment> for DeployConfig {
             health_check_type,
             health_check_port,
             public_ingress_port,
+            // Not recoverable from the fetched deployment; leave unset so the
+            // server applies its default.
+            replicas: None,
         })
     }
 }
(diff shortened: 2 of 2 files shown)


tvc/src/commands/deploy/create.rs
```diff
@@ -141,6 +141,10 @@ struct Overrides {
     /// This is usually the port your server listens on for external requests.
     #[arg(long, env = "TVC_PUBLIC_INGRESS_PORT")]
     pub public_ingress_port: Option<u16>,
+
+    /// Desired replica count for the deployment (defaults to 1).
+    #[arg(long, env = "TVC_REPLICAS")]
+    pub replicas: Option<u32>,
```
daniilrrr, line 149: you can set the `default_value` here in the clap annotation to avoid downstream setter code 


```suggestion
    #[arg(long, env = "TVC_REPLICAS", default_value_t=1)]
    pub replicas: u32,
```

tvc/src/config/deploy.rs
```diff
@@ -44,6 +44,16 @@ pub struct DeployConfig {
...
+    #[serde(default = "default_replicas")]
+    pub replicas: u32,
+}
+
+/// Default replica count for new deployments.
+pub const DEFAULT_REPLICAS: u32 = 1;
+
+fn default_replicas() -> u32 {
+    DEFAULT_REPLICAS
 }
```
daniilrrr: rm this and move to Clap annotation

tvc/src/commands/deploy/create.rs
emostov: Opted out of including a default - right now we have this field as optional and manage server side

tvc/src/config/deploy.rs
emostov: Opted out of including a default - right now we have this field as optional and manage server side

tvc/src/config/deploy.rs
```diff
@@ -109,6 +113,9 @@ impl TryFrom<TvcDeployment> for DeployConfig {
             health_check_type,
             health_check_port,
             public_ingress_port,
+            // Not recoverable from the fetched deployment; leave unset so the
```
daniilrrr, line 116: why isn't it recoverable? I think this would be useful for `tvc deployment copy` one day

tvc/src/commands/deploy/create.rs
daniilrrr: (non-blocking, you could merge as is)

i see. Then we can use a type here to eliminate the need to validate later

```suggestion
    use std::num::NonZeroU32;
    ...

    #[arg(long, env = "TVC_REPLICAS")]
    pub replicas: Option<NonZeroU32>,
```
we'd have to update a few other places 

___
LLM says:

 - DeployConfig::replicas becomes Option<NonZeroU32>
  - remove the InvalidReplicas validation check/error
  - convert back to u32 only when building the generated API intent in `build_create_intent()` with .map(NonZeroU32::get)
  - update the tests, including checking that JSON containing "replicas": 0 fails to deserialize

emostov: Done in febf89ee4d4d172cd7c37d417d69efc3f3b78500: `--replicas` and the config field are now `Option<NonZeroU32>`, the `InvalidReplicas` validation error is gone, and the value converts back to `u32` only in `build_create_intent` (the generated intent field is `Option<u32>`). Zero is now rejected at both parse boundaries — clap errors on `--replicas 0` and serde errors on `"replicas": 0` in a config file — covered by `zero_replicas_is_rejected_by_clap` and `zero_replicas_is_rejected_at_parse_time`. The omit-when-unset/server-default behavior is unchanged.

tvc/src/config/deploy.rs
emostov: It's not recoverable because the fetched `TvcDeployment`/`TvcContainerSpec` don't carry a desired replica count — that field only exists on the separate `DeploymentStatus` resource (`desiredReplicas`, returned by the app/deploy status endpoints), and this `TryFrom` is a pure conversion with no API access. Clarified the comment to say so in f63ada349912c9d02a706bafe3dc80f73fc6a870. If the deployment proto grows a replicas field later, the exhaustive destructure here fails to compile and forces the copy decision — so a future `tvc deployment copy` can pick it up then (or fetch status alongside).

emostov: Implemented in d4dc868561ff18b5a171b82964b60aa3863e43b6: `tvc deploy init --from-deployment` now also queries the deployment's status (`get_app_status` → `desiredReplicas`) and seeds `replicas` in the written config when the status reports a positive count. The `TryFrom<TvcDeployment>` conversion stays pure/network-free; the command enriches the seeded config from the status. If the status is unavailable or reports zero/unspecified, `replicas` stays unset with a stderr warning (server default applies) instead of failing the seeding. Covered by `positive_desired_replica_count_is_seeded_from_status`, `zero_desired_replica_count_is_not_seeded`, and `deployment_missing_from_status_seeds_nothing` in `commands/deploy/init.rs`.

emostov: Follow-up: the full PR is now squashed into signed commit `7b43054adca6c92a2b0488fdac2add4ec33ca93e`; this supersedes the earlier `d4dc868` reference. The status-enrichment behavior and tests are unchanged.

tvc/src/commands/deploy/create.rs
daniilrrr: cc @richardpringle we're __don't parse, validate__ing :)

tvc/src/commands/deploy/init.rs
```diff
@@ -86,8 +90,40 @@ async fn execute(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
...
+                        "deployment status reports no positive desired replica count; \
+                         leaving replicas unset (the server default applies)",
+                    )?,
+                },
+                Err(error) => ctx.shell().human().warn(format!(
+                    "failed to fetch deployment status; \
+                     leaving replicas unset (the server default applies): {error}"
+                ))?,
+            }
+            config
```
daniilrrr: i tried to use `inspect_err()` here like we do in `deploy/status.rs` to flatten the `match status_response` logic and make it appear cleaner, and didn't succeed as much as expected. but I still think this is a smidge better


```suggestion
            if let Ok(status_response) = auth
                .client
                .get_app_status(GetAppStatusRequest {
                    organization_id: org_id,
                    app_id: config.app_id.clone(),
                })
                .await
                .inspect_err(|error| {
                    let write_result = ctx.shell().human().warn(format!(
                        "failed to fetch deployment status; leaving replicas unset \
                         (the server default applies): {error}"
                    ));

                    if let Err(write_error) = write_result {
                        debug!(%write_error, "failed to write deployment-status warning");
                    }
                })
            {
                match status_response
                    .app_status
                    .map(sanitize_app_status)
                    .as_ref()
                    .and_then(|app_status| desired_replicas(app_status, &deploy_id))
                {
                    Some(count) => config.replicas = Some(count),
                    None => ctx.shell().human().warn(
                        "deployment status reports no positive desired replica count; \
                         leaving replicas unset (the server default applies)",
                    )?,
                }
            }
            config
```

tvc/src/config/deploy.rs
```diff
@@ -44,6 +45,11 @@ pub struct DeployConfig {
     pub health_check_type: TvcHealthCheckType,
     pub health_check_port: u16,
     pub public_ingress_port: u16,
+    /// Desired replica count for the deployment. When unset, the create
+    /// request omits it and the server applies its default. Zero is
+    /// unrepresentable: it is rejected while parsing the config file.
```
daniilrrr: rm 
```
    /// Zero is
    /// unrepresentable: it is rejected while parsing the config file.
```
redundant from reading type imo

daniilrrr (dismissed):
approved with nits

tvc/src/commands/deploy/init.rs
emostov: Implemented in final squashed commit `3b2714844b9f04ba346e1ea81a1df96d096cd35f`: flattened the successful status path with `if let Ok(...).inspect_err(...)`, matching `deploy/status.rs`. A status-fetch failure remains best-effort; if writing its warning also fails, that secondary output error is debug-logged rather than aborting config seeding. Existing missing/zero status behavior is unchanged.

tvc/src/config/deploy.rs
emostov: Implemented in final squashed commit `3b2714844b9f04ba346e1ea81a1df96d096cd35f`: removed the redundant zero-value explanation; `Option<NonZeroU32>` already carries that constraint.

daniilrrr (dismissed):

PR #232, opened 2026-08-05
Sync protos from mono

emostov:
## What

Full proto sync from mono per the documented `sync/rust-sdk` process, plus `make generate`.

Notable upstream changes picked up:
- `replicas` field on `CreateTvcDeploymentIntent` (CLI support lands in a stacked PR, #231)
- New activity types (secrets import, swap v2, earn fee claims, etc.)
- New `immutable/data/v18`, earn and product protos

## Codegen fix

The synced protos introduce `google.protobuf.Timestamp` fields (`ActivityFilter`, `ActivityCursor`), which prost maps to `prost_types::Timestamp` with no serde support, breaking the blanket serde derive. Codegen now maps `google.protobuf.Timestamp` to a small vendored `Timestamp` in `turnkey_client::well_known` that (de)serializes as an RFC 3339 string per the proto3 JSON mapping, using the already-present `chrono` dependency. This avoids taking on `pbjson-types` and its build-time codegen machinery for a single type.

`tvc deploy create` sets the new `replicas` intent field to `None` so behavior is unchanged; the stacked PR makes it configurable.

## Testing

`cargo build --locked`, full `cargo test` (including new unit tests for the vendored Timestamp: RFC 3339 round-trip, Z/offset parsing, negative-nanos and malformed-input rejection), and `cargo clippy --all-targets --all-features -D warnings` all clean.


Cargo.lock
```diff
@@ -4114,6 +4114,7 @@ name = "turnkey_client"
 version = "0.13.1"
 dependencies = [
  "base64 0.22.1",
+ "chrono",
```
richardpringle, line 4117: 👍

client/src/lib.rs
```diff
@@ -66,6 +66,7 @@ pub use turnkey_api_key_stamper::{TurnkeyP256ApiKey, TurnkeySecp256k1ApiKey};
 pub mod generated;
 
 pub mod retry;
+pub mod well_known;
```
richardpringle, line 69: Not blocking, but I feel like we could come up with a better name. Maybe `common`? That's pretty standard... unless this is a google naming thing, in which case it's fine

client/src/well_known.rs
daniilrrr, line 1: is `well_known` a common phrase in this context? I think `custom_types.rs` is clearer here

tvc/src/commands/deploy/create.rs
```diff
@@ -362,6 +362,8 @@ fn build_create_intent(
         health_check_type: deploy_config.health_check_type,
         health_check_port: deploy_config.health_check_port as u32,
         public_ingress_port: deploy_config.public_ingress_port as u32,
+        // Not yet configurable via the CLI; the server applies its default.
```
daniilrrr, line 365: rm comment

daniilrrr (approved):
lgtm

client/src/well_known.rs
```diff
@@ -0,0 +1,125 @@
+//! Hand-vendored `google.protobuf` well-known types used by the generated code.
```
richardpringle, line 1: Why are we "vendoring" this instead of just putting the protobuf types through the code generator? 

(the explanation should actually go on in the doc-comment here)

richardpringle: Or is this hand-written as opposed to "vendored". "vendored" implies copying someone else's code, so I just want to make sure we aren't hand-writing these implementations and calling them hand-vendored. 

If I had to guess... we've hand (well... LLM) written the conversions. I think a better way to deal with that is have the conversions in a separate file that is hand-written and not "vendored" and the type definitions stay in `src/generated`.

client/src/well_known.rs
```diff
@@ -0,0 +1,125 @@
...
+        let dt = DateTime::parse_from_rfc3339(s).map_err(serde::de::Error::custom)?;
+        Ok(dt.with_timezone(&Utc).into())
+    }
+}
+
+impl<'de> serde::Deserialize<'de> for Timestamp {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
```
richardpringle, line 68: nit (aka can choose to ignore)


```suggestion
impl<'de> serde::Deserialize<'de> for Timestamp {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        struct TimestampVisitor;

        impl serde::de::Visitor<'_> for TimestampVisitor {
            type Value = Timestamp;

            fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
                formatter.write_str("an RFC 3339 date string")
            }

            fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
            where
                E: serde::de::Error,
            {
                let dt = DateTime::parse_from_rfc3339(s).map_err(serde::de::Error::custom)?;
                Ok(dt.with_timezone(&Utc).into())
            }
        }

```

Doesn't actually matter at all but this is kind of a convention established by serde

client/src/well_known.rs
```diff
@@ -0,0 +1,125 @@
+//! Hand-vendored `google.protobuf` well-known types used by the generated code.
+
+use chrono::{DateTime, Utc};
+
+/// `google.protobuf.Timestamp`.
+///
+/// (De)serializes as an RFC 3339 string per the proto3 JSON mapping.
+#[derive(Clone, PartialEq, ::prost::Message)]
```
richardpringle, line 8:
```suggestion
#[derive(Copy, Clone, PartialEq, ::prost::Message)]
```

client/src/well_known.rs
richardpringle: I'm going to review/have been reviewing this file as though parts of it are handwritten

client/src/well_known.rs
```diff
@@ -0,0 +1,125 @@
...
+        };
+        assert!(serde_json::to_string(&ts).is_err());
+    }
+
+    #[test]
+    fn rejects_malformed_strings() {
+        assert!(serde_json::from_str::<Timestamp>("\"not a date\"").is_err());
+        assert!(serde_json::from_str::<Timestamp>("12345").is_err());
+    }
+}
```
richardpringle, line 125: Should we have a full test from proto? I'm not actually too concerned about this since there's plenty of coverage, but might be nice to have here...

client/Cargo.toml
```diff
@@ -13,6 +13,7 @@ categories = ["api-bindings"]
 
 [dependencies]
 mime.workspace = true
+chrono.workspace = true
```
richardpringle, line 16: 👍

codegen/src/main.rs
```diff
@@ -70,6 +70,9 @@ fn main() {
         .type_attribute(".google.rpc", SERDE_DERIVE)
         .field_attribute("google.rpc.Status.details", "#[serde(skip)]")
         .type_attribute("google.rpc", SERDE_CAMEL_CASE)
+        // prost_types::Timestamp has no serde support; our vendored Timestamp
+        // (de)serializes as an RFC 3339 string per the proto3 JSON mapping.
```
richardpringle, line 74: Ahhhh this is the reason. Why didn't we just go with a newtype wrapper then? This is more me being pedantic than practical, but in theory, it's our job to make sure that a type that we didn't pass through codegen requires a different type of testing. With a newtype wrapper, you rely on `prost_types` (owned, maintained, and probably proptested by the tokio team) and all we have to do is delegate to their `prost::Message` implementation (I know the proc-macro works, but that's not the point), and write our own serialization method. 

It's more about the pattern than the details for this specific types that's really simple

richardpringle: (if you agree with me, let's create a ticket instead of block the PR on this)

tvc/src/commands/deploy/create.rs
richardpringle, line 365: Again, not blocking. I would remove this comment though. When it does become configurable via the cli, no one is going to know to change this comment.

richardpringle (approved):
Nothing blocking. Please read my comments though!

proto/immutable/activity/v1/activity.proto
```diff
@@ -5950,6 +6155,18 @@ message CreateTvcAppResult {
...
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "The unique identifier for the TVC share set"}
+  ];
+  repeated string share_set_operator_ids = 6 [
+    (google.api.field_behavior) = REQUIRED,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "The unique identifiers of the share set operators"}
+  ];
+  uint32 share_set_threshold = 7 [
+    (google.api.field_behavior) = REQUIRED,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "The required number of approvals for the share set"}
+  ];
```
annieke, line 6169: lol yay my change made it in

annieke (approved):
another one

PR #235, opened 2026-08-07
tvc: propagate Vivosuite trace context

daniilrrr:
## Summary

- Propagates upstream W3C trace context onto TVC API requests.
- Connects Vivosuite test, phase, and command spans to Coordinator request spans.
- Uses the official OpenTelemetry propagator with bounded header inputs.
- Fails open: missing or malformed trace context is omitted and never blocks a request.

Supports TVC-265 and tkhq/mono#7539.

### DIFF
--- a/tvc/src/client.rs
+++ b/tvc/src/client.rs
@@ -3,7 +3,11 @@
 use crate::config::turnkey::{Config, StoredApiKey};
 use crate::errors::MissingResource;
 use anyhow::{Context, Result, anyhow, bail};
-use reqwest::header::{HeaderMap, HeaderValue};
+use opentelemetry::{propagation::TextMapPropagator, trace::TraceContextExt};
+use opentelemetry_http::{HeaderExtractor, HeaderInjector};
+use opentelemetry_sdk::propagation::TraceContextPropagator;
+use reqwest::header::{HeaderMap, HeaderName, HeaderValue, InvalidHeaderValue};
+use thiserror::Error;
 use tracing::{debug, instrument};
 use turnkey_api_key_stamper::TurnkeyP256ApiKey;
 use turnkey_client::{
@@ -21,6 +25,8 @@ const ENV_ORG_ID: &str = "TVC_ORG_ID";
 const ENV_API_BASE_URL: &str = "TVC_API_BASE_URL";
 const ENV_API_KEY_PUBLIC: &str = "TVC_API_KEY_PUBLIC";
 const ENV_API_KEY_PRIVATE: &str = "TVC_API_KEY_PRIVATE";
+const ENV_TRACEPARENT: &str = "TVC_OTEL_TRACEPARENT";
+const ENV_TRACESTATE: &str = "TVC_OTEL_TRACESTATE";
 const DEFAULT_API_BASE_URL: &str = "https://api.turnkey.com";
 
 /// An authenticated Turnkey client with organization context.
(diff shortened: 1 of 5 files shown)


tvc/src/client.rs
```diff
@@ -129,23 +134,141 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
 /// enforces a minimum-version floor on it (enforce-when-present) to retire
 /// known-defective releases with an upgrade prompt.
 const TVC_CLIENT_VERSION_HEADER: &str = "X-TVC-CLIENT-VERSION";
+const TRACEPARENT_HEADER: HeaderName = HeaderName::from_static("traceparent");
+const TRACESTATE_HEADER: HeaderName = HeaderName::from_static("tracestate");
```
richardpringle, line 139: 🥇

tvc/src/client.rs
```diff
@@ -129,23 +134,141 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
...
+const MAX_TRACE_CONTEXT_HEADER_LEN: usize = 512;
 
-/// Build the Turnkey API client used for all tvc requests.
+enum BoundedHeaderValueError {
+    TooLong,
+    Invalid,
+}
```
richardpringle, line 148: 🥇

tvc/src/client.rs
```diff
@@ -129,23 +134,141 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
...
+    type Error = BoundedHeaderValueError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        if value.len() > MAX_LEN {
+            return Err(BoundedHeaderValueError::TooLong);
+        }
+
+        HeaderValue::from_str(value)
+            .map(Self)
+            .map_err(|_| BoundedHeaderValueError::Invalid)
```
richardpringle: You didn't want to wrap the error here? (don't need to, just wondering)

tvc/src/client.rs
```diff
@@ -129,23 +134,141 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
...
+    }
+}
+
+impl<const MAX_LEN: usize> From<BoundedHeaderValue<MAX_LEN>> for HeaderValue {
+    fn from(value: BoundedHeaderValue<MAX_LEN>) -> Self {
+        value.0
+    }
+}
+
+type TraceContextHeaderValue = BoundedHeaderValue<MAX_TRACE_CONTEXT_HEADER_LEN>;
```
richardpringle, line 173: 👍

tvc/src/client.rs
```diff
@@ -129,23 +134,141 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
...
+                Ok(value) => Some(value),
+                Err(BoundedHeaderValueError::TooLong) => {
+                    debug!(name, "ignoring oversized propagated trace header");
+                    None
+                }
+                Err(BoundedHeaderValueError::Invalid) => {
+                    debug!(name, "ignoring invalid propagated trace header");
+                    None
+                }
+            };
```
richardpringle: nit: This should just be a free function, there's no need for it to be a closure

tvc/src/client.rs
```diff
@@ -129,23 +134,141 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
...
+                    parse_header("traceparent", traceparent)?.into(),
+                );
+                map
+            }
+            (None, Some(_)) => {
+                debug!("ignoring propagated tracestate without traceparent");
+                return None;
+            }
+            (None, None) => return None,
+        };
```
richardpringle:
```suggestion
        let carrier = [(TRACEPARENT_HEADER, traceparent), (TRACESTATE_HEADER, tracestate)]
            .into_iter()
            .filter_map(|(name, header)| TraceContextHeaderValue::try_from(header).ok().map(|header| (name, header)))
            .collect::<HeaderMap>()?
```

you can expand the `filter_map` to include errors and just use the header name in the `debug!`

tvc/src/client.rs
```diff
@@ -129,23 +134,141 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
...
+            return None;
+        }
+
+        let had_tracestate = carrier.contains_key(TRACESTATE_HEADER);
+        let mut headers = HeaderMap::new();
+        propagator.inject_context(&context, &mut HeaderInjector(&mut headers));
+
+        if headers
+            .get(TRACESTATE_HEADER)
+            .is_some_and(|value| value.is_empty())
```
richardpringle: Make sure `tracestate` isn't empty before you get to this point with `let tracestate = tracestate.filter(|val| !val.is_empty())`

tvc/src/client.rs
```diff
@@ -129,23 +134,141 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
...
+        {
+            headers.remove(TRACESTATE_HEADER);
+        }
+        if had_tracestate && !headers.contains_key(TRACESTATE_HEADER) {
+            debug!("ignoring invalid propagated tracestate");
+        }
```
richardpringle: If you follow the advice from the above comment, this check isn't necessary.

tvc/src/client.rs
```diff
@@ -129,23 +134,141 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
...
+}
+
+fn tvc_client_headers(traceparent: Option<&str>, tracestate: Option<&str>) -> HeaderMap {
     let mut headers = HeaderMap::new();
     headers.insert(
         TVC_CLIENT_VERSION_HEADER,
         HeaderValue::from_static(env!("CARGO_PKG_VERSION")),
     );
 
+    if let Some(trace_context) = TraceContextHeaders::from_values(traceparent, tracestate) {
```
richardpringle: is `from_values` used anywhere else? If not, this should just be inlined

tvc/src/client.rs
```diff
@@ -129,23 +134,141 @@ async fn load_credentials_from_config() -> Result<(String, String, String, Strin
...
+///
+/// Trace propagation is best-effort: malformed context is omitted and never
+/// prevents the authenticated API client from being constructed.
+fn build_turnkey_client_with_trace_context(
+    stamper: TurnkeyP256ApiKey,
+    api_base_url: &str,
+    traceparent: Option<&str>,
+    tracestate: Option<&str>,
+) -> Result<TurnkeyClient<TurnkeyP256ApiKey>> {
+    let headers = tvc_client_headers(traceparent, tracestate);
```
richardpringle: Not a comment for a change, as this pattern exists everywhere in this repo, but it's an antipattern. The type system is strong enough that it's really easy to call `tvc_client_headers` _outside_ of `build_turnkey_client_with_trace_context`. It simplifies the dependency chain and is much more flexible in case there's ever the need to build the client with different headers.

tvc/src/client.rs
```diff
@@ -154,6 +277,27 @@ pub(crate) fn build_turnkey_client(
...
+    let traceparent = read_env_var(ENV_TRACEPARENT);
+    let tracestate = read_env_var(ENV_TRACESTATE);
+
+    build_turnkey_client_with_trace_context(
+        stamper,
+        api_base_url,
+        traceparent.as_deref(),
+        tracestate.as_deref(),
+    )
+}
```
richardpringle: Same comment as above. This is an anti-pattern and couples code. 

We want to separate resource acquisition from business logic. The function should receive ready-to-use dependencies, so it will have fewer failure modes, no hidden setup or I/O, less coupling to a concrete construction method, and is easier to test and reuse. Failures also stay at the wiring boundary, where they can be handled with better context.

richardpringle (commented):
I left some comments that aren't blockers... but the more we write code like this, the more work it is to undo it. I would suggest we land this then once verified, you spend the time to unwrap some of these coupled free functions.

I can't approve yet as there's real dependency addition here

tvc/src/client.rs
daniilrrr: wtf i did update this

daniilrrr: resolving, updated

tvc/src/client.rs
daniilrrr: applied. 

didn't apply in `normalize_trace_context_headers()` because the W3C propagator will inject a `tracestate` automatically for 
 - No input tracestate
  - Malformed non-empty tracestate
  - Whitespace-only tracestate

which we want to avoid

tvc/src/client.rs
daniilrrr: [see other comment](https://github.com/tkhq/rust-sdk/pull/235#discussion_r3769897940)

tvc/src/client.rs
daniilrrr: no longer relevant with refactored code imo, and the different combos have to be handled differently 

```
The valid input combinations are:
-  traceparent + tracestate → forward both
- traceparent only → forward traceparent
- tracestate only → forward neither
- neither → forward neither
```

  traceparent can stand alone; tracestate cannot.

tvc/src/client.rs
daniilrrr: done

tvc/src/client.rs
daniilrrr: fixed

tvc/src/client.rs
daniilrrr: updated to 
```
pub(crate) fn build_turnkey_client(
    stamper: TurnkeyP256ApiKey,
    api_base_url: &str,
    headers: HeaderMap,
) -> Result<TurnkeyClient<TurnkeyP256ApiKey>> 
```

let me know if that's what you meant

tvc/src/client.rs
daniilrrr: updated to 
```
pub(crate) fn build_turnkey_client(
    stamper: TurnkeyP256ApiKey,
    api_base_url: &str,
    headers: HeaderMap,
) -> Result<TurnkeyClient<TurnkeyP256ApiKey>> 
```

let me know if that's what you meant

richardpringle: That's exactly what I meant! But it doesn't look like this was updated

richardpringle (approved):

daniilrrr:
self-blocking on merging until I resolve a discussion about how we do backend tracing and resolve `tracecontext` vs existing `b3` propagator

PR #236, opened 2026-08-10
feat(tvc): operator key backup command and login onboarding nudge

richardpringle:
Extracted from #224 (first of the stack).

## Typed operator public key

`StoredQosOperatorKey.public_key` becomes `QosOperatorPublicKey`, a 130-byte composite (`encrypt_public ‖ sign_public`) parsed once at the file boundary; hex stays the display and serialization form, so disk and JSON shapes are unchanged. The backup command and login nudge below carry the typed key instead of re-validating strings.

## TVC-52 — `tvc keys backup-operator-key`

Copies the org's operator key file byte-for-byte after validating it parses (unknown fields survive — no re-serialize). Interactive runs prompt for the destination and confirm overwrites; non-interactive requires `--output`/`--overwrite`. Output includes the private-key warning, storage guidance, and the manual restore path (restore command deferred). Org resolution follows login's `find_org` (alias or org ID, else the active org). Adopts the `Run` trait from #223.

## TVC-53 — onboarding backup nudge

When login generates a fresh operator key interactively, it explains the key exists only on this machine and offers a backup (default Yes) via the shared core; backup errors degrade to a warning so login never dies. Declining prints the pointer to the standalone command; re-logins with an existing key get a single tip line. A one-shot mock whoami server carries PTY login past credential verification, so the accept/decline/tip paths are all covered end-to-end.

Closes TVC-52. Closes TVC-53.


emostov (dismissed):
dep review only approval

tvc/src/commands/login.rs
```diff
@@ -599,6 +602,57 @@ async fn find_or_generate_operator_key(
...
     )?;
 
+    // Onboarding nudge for the freshly generated key. JSON mode already
+    // forces non-interactive; the TTY check keeps piped runs from hanging on
+    // the prompt.
+    if !ctx.is_non_interactive() && prompts::stdin_can_prompt() {
+        shell_println!(ctx)?;
+        shell_println!(
+            ctx,
+            "This key exists only on this machine; if it's lost you cannot \
```
daniilrrr: (human) add a "WARNING" or similar callout here? kinda a big deal imo

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,282 @@
...
+
+impl Run for Args {
+    type Outcome = OperatorKeyBackedUp;
+
+    async fn run(self, ctx: &mut StdCtx) -> Result<OperatorKeyBackedUp> {
+        // Non-interactive mode cannot prompt for the destination; reject
+        // before loading config or resolving the organization.
+        if ctx.is_non_interactive() && self.output.is_none() {
+            return Err(error_required_in_non_interactive("--output"));
+        }
```
daniilrrr: **(AGENT) blocking — piped stdin falls through to a prompt that cannot run**

`ctx.is_non_interactive()` is only the `--non-interactive` flag plus JSON mode (`output.rs:204`); it says nothing about whether stdin is a TTY. So a piped run without the env var reaches `prompts::text` at line 63 and dies in `inquire`:

```
$ tvc keys backup-operator-key < /dev/null
error: The input device is not a TTY        (exit 1)
```

instead of `--output is required in non-interactive mode`. `deploy/approve.rs:121` guards with `ctx.is_non_interactive() || !stdin_can_prompt()`, and so does the login nudge added in this same PR (`login.rs:608`) — this is the one call site that doesn't.

Binding the decision once also fixes the overwrite branch (see the comment on line 78), which has the same hole:

```suggestion
        // Reject before loading config or resolving the organization when
        // there is no way to prompt for the destination: --non-interactive,
        // JSON mode, or a non-TTY stdin.
        let can_prompt = !ctx.is_non_interactive() && prompts::stdin_can_prompt();
        if !can_prompt && self.output.is_none() {
            return Err(error_required_in_non_interactive("--output"));
        }
```

Needs `prompts::stdin_can_prompt` added to the `use` at line 9.

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,282 @@
...
+        if destination.exists() && !self.overwrite {
+            if ctx.is_non_interactive() {
+                bail!(
+                    "destination {} already exists; pass --overwrite to replace it",
+                    destination.display()
+                );
+            }
+
+            prompts::confirm_or_bail(&format!("Overwrite {}?", destination.display()), "backup")?;
+        }
```
daniilrrr: **(AGENT) blocking — same TTY hole on the overwrite path**

Second half of the fix on line 42. As written, a piped run whose `--output` already exists skips the `bail!` (the flag isn't set) and calls `confirm_or_bail`, which fails with `The input device is not a TTY` rather than the actionable `pass --overwrite to replace it`.

Reusing `can_prompt` from the suggestion above:

```suggestion
        if destination.exists() && !self.overwrite {
            if !can_prompt {
                bail!(
                    "destination {} already exists; pass --overwrite to replace it",
                    destination.display()
                );
            }

            prompts::confirm_or_bail(&format!("Overwrite {}?", destination.display()), "backup")?;
        }
```

Apply together with the line 42 suggestion — neither compiles alone.

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,282 @@
...
+        tokio::fs::create_dir_all(parent)
+            .await
+            .with_context(|| format!("failed to create backup directory: {}", parent.display()))?;
+    }
+
+    // Written with default (umask) permissions, matching
+    // `StoredQosOperatorKey::save`; tightening both is tracked by TVC-241.
+    tokio::fs::write(&destination, &bytes)
+        .await
+        .with_context(|| format!("failed to write backup: {}", destination.display()))?;
```
daniilrrr, line 168: **(AGENT) blocking — the backup of a private key is created world-readable**

Deferring `StoredQosOperatorKey::save`'s permissions to TVC-241 is reasonable; deferring them *here* is less so. This command's entire purpose is writing private key material to a new path the user chose, and the output text three lines later tells them to treat the file as a secret ("The backup contains the PRIVATE key. Store it somewhere safe"). At umask defaults that file is typically `0644`. The code should agree with the advice it prints, and it's a local fix rather than a cross-cutting one:

```suggestion
    // The backup holds private key material, so create it 0600 rather than at
    // umask defaults. TVC-241 tracks the same tightening for the primary key
    // file, which this command reads but does not write.
    let mut options = tokio::fs::OpenOptions::new();
    options.write(true).create(true).truncate(true);
    #[cfg(unix)]
    options.mode(0o600);
    let write_context = || format!("failed to write backup: {}", destination.display());
    options
        .open(&destination)
        .await
        .with_context(write_context)?
        .write_all(&bytes)
        .await
        .with_context(write_context)?;
```

Needs `use tokio::io::AsyncWriteExt;` and, under `#[cfg(unix)]`, `use std::os::unix::fs::OpenOptionsExt;`. Happy to be told this belongs in TVC-241 with the other one instead — but then it's worth saying so in the doc comment, since a reader currently sees "matching `save`" and may not realize that means 0644.

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,282 @@
...
+        _ => anyhow::Error::new(e)
+            .context(format!("failed to read operator key: {}", source.display())),
+    })?;
+
+    let key: StoredQosOperatorKey = serde_json::from_slice(&bytes).with_context(|| {
+        format!(
+            "operator key at {} is not a valid operator key file",
+            source.display()
+        )
+    })?;
```
daniilrrr, line 154: **(AGENT) medium — "parsed to validate it" is a much weaker claim than it reads as**

The doc at line 95 says the source "is parsed to validate it." What's actually checked is 130 bytes of hex on the *public* key. `private_key` is still an unparsed `String` — which means the one field that determines whether the backup is restorable is never looked at:

```
$ cat operator.json     # public_key: 130 random bytes, private_key: "not-even-hex"
$ tvc keys backup-operator-key --output bk.json
Operator key backed up!            (exit 0)
```

A backup that cannot be restored reports success in green. Per AGENTS.md's parse-don't-validate rule this is the remaining unparsed field on the DTO, and the parser already exists (`HexSeed` / `LocalPair::from_hex_seed`, `pair.rs:30`/`:67`) — plus `LocalPair::public_key()` returns the same composite encoding, so the round-trip check is nearly free:

```suggestion
    let key: StoredQosOperatorKey = serde_json::from_slice(&bytes).with_context(|| {
        format!(
            "operator key at {} is not a valid operator key file",
            source.display()
        )
    })?;

    // A backup is only worth reporting as one if it can be restored, so prove
    // the stored seed derives the stored public key before copying.
    let derived = LocalPair::from_hex_seed(&key.private_key).with_context(|| {
        format!(
            "operator key at {} has an unusable private key",
            source.display()
        )
    })?;
    if QosOperatorPublicKey::try_from(derived.public_key().as_slice())? != key.public_key {
        bail!(
            "operator key at {} is inconsistent: its private key does not derive its public key",
            source.display()
        );
    }
```

Needs `use crate::pair::{LocalPair, Pair};`. If you'd rather keep `back_up` cheap and I/O-only, the alternative is to soften the doc comment — but then the success message probably shouldn't be as confident as it is.

tvc/src/commands/keys/backup_operator_key.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        let source = &org_config.select_local_record(alias)?.key_path;
+
+        let destination: PathBuf = self.output.map(Ok).unwrap_or_else(|| {
+            prompts::text(
+                "Backup file path",
+                Some(&format!("operator-{alias}-backup.json")),
+            )
+            .map(Into::into)
+        })?;
```
daniilrrr: **(AGENT) nit — the default destination is relative, so it lands in `$PWD`**

`operator-{alias}-backup.json` with no directory resolves against the current working directory, which for a dev running `tvc` is very often a git repo. The success text then says "never in source control" — while the default we handed them may well be inside one. A `$HOME`-rooted default (or `~/.config/turnkey/backups/`) would match the advice we print. Same default is used by the login nudge (`login.rs:619`), so worth deciding once.

tvc/src/config/turnkey/qos_operator_key.rs
```diff
@@ -2,14 +2,100 @@
...
+///
+/// The bytes are `qos_p256::P256Public::to_bytes()`'s composite encoding —
+/// `encrypt_public ‖ sign_public`, two 65-byte uncompressed SEC1 points —
+/// but tvc never reads the halves, so it deliberately doesn't model the
+/// split. Hex is the display and serialization form.
+///
+/// TODO(TVC-270): the proper home for this type is qos_p256, as a bytemuck
+/// repr-backed wire form of `P256Public` (which already holds the two keys
+/// as separate fields, as does the Turnkey API's proto); move it upstream
+/// and store that type here instead.
```
daniilrrr, line 27: **(AGENT) medium — this is the third representation of an operator public key, and the weakest of the three**

The crate already has, for the same concept:

- `operator::OperatorPublicKey(PublicKey)` (`operator.rs:46`) — one 65-byte SEC1 point, hex `FromStr`/`Display`, and it validates the `0x04` prefix **and** that the bytes are a real point on P-256 (`InvalidEncoding` / `InvalidPoint`).
- `ResolvedHostedOperator::composite_public_key()` (`operator.rs:313`) — builds exactly this `encrypt ‖ sign` 130-byte composite, returned as an untyped `Vec<u8>`.

So the new type checks strictly less than the sibling sitting next to it (length only — no prefix check, no point-on-curve check) with the validating machinery already in scope. Preferred shape, which subsumes both:

```rust
pub struct QosOperatorPublicKey { encrypt: OperatorPublicKey, sign: OperatorPublicKey }
```

and `composite_public_key()` returns that instead of `Vec<u8>`.

Separately, "tvc never reads the halves, so it deliberately doesn't model the split" isn't accurate — `operator.rs` reads them (`encrypt_public_key()`, `sign_public_key()`) and produces the composite from them. If the unification is really waiting on TVC-270, the doc should say *not yet* rather than *deliberately*:

```suggestion
/// A stored QOS operator public key.
///
/// The bytes are `qos_p256::P256Public::to_bytes()`'s composite encoding —
/// `encrypt_public ‖ sign_public`, two 65-byte uncompressed SEC1 points. The
/// halves are not modelled here yet; the hosted-operator path in
/// [`crate::operator`] carries them as a validated pair instead, and only
/// length and charset are checked on this side. Hex is the display and
/// serialization form.
///
/// TODO(TVC-270): the proper home for this type is qos_p256, as a bytemuck
/// repr-backed wire form of `P256Public` (which already holds the two keys
/// as separate fields, as does the Turnkey API's proto); move it upstream,
/// store that type here, and fold the two representations together.
```

tvc/src/config/turnkey/qos_operator_key.rs
```diff
@@ -30,11 +116,7 @@ impl StoredQosOperatorKey {
         let key: StoredQosOperatorKey = serde_json::from_str(&content)
             .with_context(|| format!("failed to parse operator key: {}", path.display()))?;
```
daniilrrr, line 117: **(AGENT) medium — worth a deliberate call: this now gates paths that never read the field**

Typing the field makes `load` fail for any `public_key` that isn't 130-byte hex, including on paths that only need the seed. `local_operator_key.rs:88` loads the file and then reads `private_key` and nothing else, so with a legacy or hand-edited public key you get:

```
$ tvc deploy approve --dangerous-skip-interactive --skip-post --manifest ...
error: failed to parse operator key: .../operator.json: must be 130 bytes, got 3 at line 1 column 23
```

Approving is blocked by a field the approve path doesn't use, and the same applies to `login`. Real exposure looks low — `login` is the only writer and always emits 130 bytes — so I think this break is fine and arguably desirable (AGENTS.md: surface malformed persisted values). Flagging it so it's a decision rather than a side effect: `operator.json` is a compatibility boundary, and this is a behaviour change to it.

(`app/init.rs` and friends swallow it via `.ok()??` and silently omit the key from the template — pre-existing, not introduced here.)

tvc/src/config/turnkey/qos_operator_key.rs
```diff
@@ -2,14 +2,100 @@
...
+    ($ty:ident) => {
+        impl Debug for $ty {
+            fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+                write!(f, concat!(stringify!($ty), "({})"), hex::encode(self.0))
+            }
+        }
+    };
+}
+
+impl_hex_debug!(QosOperatorPublicKey);
```
daniilrrr, line 81: **(AGENT) nit — a `macro_rules!` with exactly one call site**

The rename-safety rationale is real, but macro plus its three-line doc comment is longer than what it generates, and AGENTS.md says extract on the third occurrence rather than in anticipation. If `StoredQosOperatorKey`'s private key later gets the same treatment, the macro earns its keep and can come back:

```suggestion
/// Hex, not the raw byte array, so the value stays greppable in logs and test
/// failures.
impl Debug for QosOperatorPublicKey {
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        write!(f, "QosOperatorPublicKey({})", hex::encode(self.0))
    }
}
```

Tradeoff acknowledged: this loses the compile-time rename check. Fine either way — just noting the current version is a macro serving one caller.

tvc/src/config/turnkey/qos_operator_key.rs
```diff
@@ -2,14 +2,100 @@
...
+///
+/// TODO(TVC-270): the proper home for this type is qos_p256, as a bytemuck
+/// repr-backed wire form of `P256Public` (which already holds the two keys
+/// as separate fields, as does the Turnkey API's proto); move it upstream
+/// and store that type here instead.
+#[derive(Clone, Copy, PartialEq, Eq, SerializeDisplay, DeserializeFromStr)]
+pub struct QosOperatorPublicKey([u8; QOS_OPERATOR_PUBLIC_KEY_LEN]);
+
+/// Error returned when parsing a [`QosOperatorPublicKey`].
+#[derive(Debug, displaydoc::Display, thiserror::Error)]
```
daniilrrr, line 32: **(AGENT) nit — error style diverges from its sibling**

This uses `displaydoc::Display`; `operator::OperatorPublicKeyParseError` (`operator.rs:49`) models the same thing with `#[error("...")]` attributes. Both idioms exist in the crate (`approvals.rs`, `deploy/approve.rs` use displaydoc), so this isn't wrong — but two parse-error types for two flavours of the same value reading differently is the kind of thing that makes the pair harder to unify later. Whichever way, worth matching the one you're most likely to merge with.

tvc/src/config/turnkey/qos_operator_key.rs
```diff
@@ -2,14 +2,100 @@
...
+impl Default for QosOperatorPublicKey {
+    fn default() -> Self {
+        // Not `Self(Default::default())`: std's array `Default` stops at 32
+        // elements, so the repeat expression delegates per element instead.
+        Self([Default::default(); QOS_OPERATOR_PUBLIC_KEY_LEN])
+    }
+}
+
 /// Operator key stored in operator.json
 #[derive(Debug, Clone, Serialize, Deserialize)]
```
daniilrrr, line 94: **(AGENT) nit — `Debug` derive on a struct holding a private key**

`private_key: String` is a hex master seed, and the derive will print it in full. Nothing currently `{:?}`-prints this struct so there's no live leak, and it predates this PR — but this commit is the one that took the type seriously (it also dropped `has_public_key` from the debug log two lines down, which reads as the same instinct). `HexSeed` next door already solves this the right way (`pair.rs:45`: manual `Debug` printing `HexSeed(..)`); either drop the derive here or store the seed as `HexSeed` and let it handle redaction.

tvc/src/commands/login.rs
```diff
@@ -599,6 +602,57 @@ async fn find_or_generate_operator_key(
...
+                        shell_eprintln!(ctx, "WARNING: backup failed: {error:#}")?;
+                        None
+                    }
+                }
+            } else {
+                None
+            }
+        } else {
+            None
+        };
```
daniilrrr: **(AGENT) blocking — "must not die here" only holds for backup errors, not for the prompts**

The comment on line 623 commits to login surviving this block, and the `back_up` error path honours it. The prompts don't: `prompts::confirm` and `prompts::text` return `Result`, and every one of them is `?`-propagated out of `find_or_generate_operator_key` into `execute_login`. Escape or Ctrl-C at "Back up your operator key now?" or at "Backup file path" therefore aborts login with a non-zero exit and no `Successfully logged in!` output — for a run that had already saved the config (line 336) and both key files (line 589) and genuinely succeeded.

Routing the whole attempt through one error sink makes the comment true, and folds in the missing `is_dir` guard (see the note on line 629):

```suggestion
        // Everything below is advisory: a prompt the user escapes out of and a
        // backup that fails both degrade to a warning, because the config and
        // both key files are already saved by this point and the login outcome
        // must still land.
        let attempt: Result<Option<OperatorKeyBackedUp>> = async {
            if !prompts::confirm("Back up your operator key now?", true)? {
                return Ok(None);
            }

            let destination: PathBuf = prompts::text(
                "Backup file path",
                Some(&format!("operator-{org_alias}-backup.json")),
            )?
            .into();

            if destination.is_dir() {
                bail!(
                    "destination {} is a directory; include a file name",
                    destination.display()
                );
            }

            if destination.exists()
                && !prompts::confirm(&format!("Overwrite {}?", destination.display()), false)?
            {
                return Ok(None);
            }

            back_up(org_alias.to_string(), local.key_path.clone(), destination)
                .await
                .map(Some)
        }
        .await;

        let backed_up = match attempt {
            Ok(report) => report,
            Err(error) => {
                shell_eprintln!(ctx, "WARNING: backup skipped: {error:#}")?;
                None
            }
        };
```

Needs `OperatorKeyBackedUp` added to the `back_up` import on line 4. A PTY test that sends `\x1b` (or closes the pty) at the destination prompt and still expects `Successfully logged in!` would lock this in — the three existing tests all drive the happy paths.

tvc/src/commands/login.rs
```diff
@@ -599,6 +602,57 @@ async fn find_or_generate_operator_key(
...
+            )?
+            .into();
+
+            // Declining the overwrite skips the backup rather than failing:
+            // login has already succeeded and must not die here.
+            let proceed = !destination.exists()
+                || prompts::confirm(&format!("Overwrite {}?", destination.display()), false)?;
+
+            if proceed {
+                match back_up(org_alias.to_string(), local.key_path.clone(), destination).await {
```
daniilrrr: **(AGENT) medium — the nudge re-implements `Args::run`'s prompt logic, and the two have already drifted**

`back_up` is shared, but the ~15 lines above it (default destination, existence check, overwrite confirmation) are a second copy of `backup_operator_key.rs:62-86`, and they already disagree:

| | standalone | login nudge |
|---|---|---|
| directory destination | rejected up front (`:70`) | not checked |
| overwrite declined | `confirm_or_bail` → error | skip, keep going |

The second row is a deliberate and correct difference. The first isn't: type a directory at login's prompt and you get `Overwrite /some/dir?` → yes → `WARNING: backup failed: <raw io error>`. That's exactly the drift the duplication predicts, and there will be more of it the next time either prompt changes.

Suggest lifting a `pub(crate) fn prompt_for_backup_destination(alias: &str) -> Result<Option<PathBuf>>` into `backup_operator_key.rs` next to `back_up` — it owns the default name, the `is_dir` rejection, and the overwrite question, and returns `None` for "user declined". Both call sites then differ only in how they treat `None`. (The suggestion on line 616 patches the `is_dir` symptom inline; this is the version that stops it recurring.)

tvc/src/commands/login.rs
```diff
@@ -566,6 +568,7 @@ async fn find_or_generate_operator_key(
     if let Some(operator_key) = StoredQosOperatorKey::load(&local.key_path).await? {
         debug!("using existing operator key");
         shell_println!(ctx, "Using existing operator key.")?;
+        shell_println!(ctx, "Tip: back it up with `tvc keys backup-operator-key`.")?;
```
daniilrrr, line 572: **(AGENT) nit — unconditional tip on every re-login**

Correctly suppressed in JSON mode (`shell_println!` goes through the human sink), but it still prints on every `--non-interactive` human run and on every interactive re-login, forever, with no notion of whether a backup already exists. As a stopgap for TVC-53 that's defensible — there's no backup state to check against. Two smaller options if you want to cut the nagging: gate it on `prompts::stdin_can_prompt()` so scripted runs stay quiet, or record the last backup path in the local operator record and only nudge when it's absent.

tvc/tests/common.rs
```diff
@@ -0,0 +1,102 @@
+//! Fixtures shared across the tvc integration-test binaries.
+//!
+//! Each test binary that declares `mod common;` compiles its own copy and
+//! uses a subset of these helpers, so unused items are expected in every
+//! build of this file; the crate-level allow keeps those builds quiet.
+
+#![allow(dead_code)]
```
daniilrrr: **(AGENT) nit — good addition, but the existing copies weren't migrated onto it**

`write_profiles_config` is very nearly `non_interactive.rs:31` (`write_config`) — same v1 header, same `OrgConfig`, same `OperatorRecord::local`, same `LOCAL_API_BASE_URL` constant — and `login.rs:35` / `auth_env.rs:112` each have another variant. `non_interactive.rs` is the file that gained a test in this same commit and still doesn't `mod common;`.

Net effect right now is a fourth copy rather than one shared fixture. Worth either migrating those three in this PR (mechanical, and `write_profiles_config` already covers their shapes) or noting the follow-up, otherwise the shared module drifts from the copies it was meant to replace.

tvc/tests/pty.rs
```diff
@@ -21,6 +28,73 @@ fn spawn(args: &str) -> PtySession {
...
+        .env_remove("TVC_NON_INTERACTIVE");
+
+    rexpect::session::spawn_command(cmd, Some(TIMEOUT_MS))
+        .unwrap_or_else(|e| panic!("spawn failed: {e}\n  cmd: {bin} {}", args.join(" ")))
+}
+
+/// One-shot mock Turnkey API that answers the whoami query, enough to carry
+/// login past its credential verification and into the operator-key flow.
+/// Same shape as `tests/error_output.rs::spawn_json_server`.
+fn spawn_whoami_server() -> (String, JoinHandle<()>) {
```
daniilrrr, line 51: **(AGENT) nit — this helper's home is `tests/common.rs`**

The doc comment already says it: "Same shape as `tests/error_output.rs::spawn_json_server`." Now that `tests/common.rs` exists in this very commit, the duplicate should move there rather than be documented as a duplicate. `common.rs` is `#![allow(dead_code)]` precisely so per-binary unused helpers are fine.

Also worth a line noting the one-shot contract — the thread `accept()`s exactly once, so a login that made a second request would see connection-refused rather than a hang. That's the right call for these tests, it's just invisible to the next person adding one.

Unrelated to the mechanics: carrying PTY login past credential verification is a genuinely useful unlock, and having all three onboarding paths covered end-to-end is more than most CLI onboarding flows get.

daniilrrr: (human) agree

tvc/tests/common.rs
daniilrrr: (human) I think migrating in this PR is fine, but non-blocking if you choose not to

also i commented out this `![allow(dead_code)]` locally and `cargo check` passed so I think it's fine to remove?

tvc/src/config/turnkey/qos_operator_key.rs
daniilrrr: (human) this might be slop but try the command it highlighted and see for yourself if you're happy

tvc/src/config/turnkey/qos_operator_key.rs
daniilrrr: (human) super agree and I was concerned about storing the `private_key` as String throughout this PR

tvc/src/config/turnkey/qos_operator_key.rs
daniilrrr: (human) shrug, keep if you want

tvc/src/config/turnkey/qos_operator_key.rs
daniilrrr: (human) +1, we already use `#[error("...")]` / `thiserror` throughout so I would appreciate sticking to a convention (and maybe adding this convention to AGENTS.md). If you prefer a diff convention then lets do it in a diff PR

tvc/src/config/turnkey/qos_operator_key.rs
daniilrrr: (human) I like 
`pub struct QosOperatorPublicKey { encrypt: OperatorPublicKey, sign: OperatorPublicKey }`
conceptually but leave it to you to do in this PR if you want

I do think that we now have several fragmented "versions" of keys and it's harder to keep straight overall, and unifying or refactoring types would be nice

tvc/src/commands/login.rs
daniilrrr: (human) I think this is a good shout

tvc/src/commands/login.rs
daniilrrr: (human) - sure, esp if they've drifted

tvc/src/commands/login.rs
daniilrrr: (human) I think this is fine, up to you

tvc/src/commands/keys/backup_operator_key.rs
daniilrrr: (human) I'm not familiar with Unix conventions but this seems fine, albeit not blocking imo

tvc/src/commands/keys/backup_operator_key.rs
daniilrrr: (human) I think this is good type tightening for the private key. I also think that maybe we could store it as something besides a `String` for additional redaction-safety for logging

tvc/src/commands/keys/backup_operator_key.rs
daniilrrr: (human) good call imo

tvc/src/commands/keys/backup_operator_key.rs
daniilrrr: (human) good call

tvc/src/commands/keys/backup_operator_key.rs
daniilrrr: (human) agree

daniilrrr (commented):
trying something a little different with this review

I started human-reviewing the PR with Claude Opus 5 locally, and it flagged several edge cases off the bat. I am not very familiar with this code, so I thought the most productive thing would be to demarcate the pure-AGENT comments, and add my own human review reply on each. Feel free to override any of the comments, but have a look. 

I did not intend to add slop to your reading workload, but rather use the agent to review a variety of paths and edge cases

**(AGENT) acceptance checklist — the five commands that misbehave today**

Every one of these was reproduced against `2d1908d` with the script at the bottom. Note the framing: two of them must keep exiting non-zero and change only their *message*, and one must start failing where it currently succeeds — "must not fail" isn't the criterion for all five.

---

### 1. No `--output`, stdin is a pipe, `TVC_NON_INTERACTIVE` unset

```
$ tvc keys backup-operator-key < /dev/null
error: The input device is not a TTY          exit=1
```

**Required:** still `exit=1`, but `error: --output is required in non-interactive mode`. The exit code is already correct; the message is the defect. Covers the blocking comment on `backup_operator_key.rs:42`.

### 2. `--output` already exists, stdin is a pipe, `TVC_NON_INTERACTIVE` unset

```
$ echo previous > bk.json
$ tvc keys backup-operator-key --output bk.json < /dev/null
error: The input device is not a TTY          exit=1
```

**Required:** still `exit=1`, with `error: destination bk.json already exists; pass --overwrite to replace it`, and `bk.json` still containing `previous`. Covers the blocking comment on `backup_operator_key.rs:78`.

### 3. Permissions of a written backup

```
$ tvc keys backup-operator-key --output bk2.json    # TVC_NON_INTERACTIVE=1
$ ls -l bk2.json
-rw-r--r--   bk2.json
```

**Required:** `-rw-------`, under the default `umask 022` this repo's CI and a stock dev shell both use. Covers the blocking comment on `backup_operator_key.rs:129`. Worth asserting the mode in the test rather than eyeballing it — `std::os::unix::fs::PermissionsExt::mode() & 0o777 == 0o600`.

### 4. An operator key whose private key cannot derive its public key

```
$ cat operator.json    # public_key: 130 valid hex bytes, private_key: "not-even-hex"
$ tvc keys backup-operator-key --output bk3.json
Operator key backed up!                       exit=0
```

**Required:** non-zero exit naming the source path, and **no** `bk3.json` written. This is the inverse of the others — it currently succeeds and must start failing. Covers the medium comment on `backup_operator_key.rs:112`. If you decide to keep `back_up` I/O-only instead, this check comes off the list and the doc comment on line 95 changes instead — but then say so on that thread, because as written the command reports success for a backup that cannot be restored.

### 5. `login`: cancelling the backup prompt kills a login that already succeeded

```
$ printf '\x04' | script -q /dev/null tvc login --org alias-a
...
Operator Key Generated!
? Back up your operator key now? <canceled>
error: Operation was canceled by the user     exit=1
```

with `operator.json` **already on disk** and the config already saved. Re-running `tvc login` immediately afterwards prints `Successfully logged in!` and `exit=0`, which is what makes it clear the first run had genuinely succeeded before the prompt killed it.

**Required:** `exit=0` and `Successfully logged in!` in the output, with a warning line about the skipped backup. Covers the blocking comment on `login.rs:616`. The probe uses Ctrl-D because it scripts cleanly; Ctrl-C and Esc reach the same `InquireError` and the same `?`, so one PTY test covers all three. Both prompts need it — the destination prompt on line 617 has the identical problem, so please cover cancelling at *either* one.

---

### Not on this list, deliberately

`tvc deploy approve --dangerous-skip-interactive --skip-post --manifest ...` against an `operator.json` with a short `public_key` currently fails with `must be 130 bytes, got 3`. That's the subject of the medium comment on `qos_operator_key.rs:116`, and it's a **decision**, not a defect — if you accept the compatibility break (which I'd lean towards) it should keep failing, so there's nothing to verify. It only becomes an acceptance criterion if you decide to preserve compatibility for the seed-only paths.

---

<details>
<summary>Reproduction script (run from the repo root after <code>cargo build -p tvc</code>)</summary>

Needs `openssl`, `python3`, and `script(1)`. Builds a throwaway `$HOME` with a real p256 API key and a one-shot whoami server on port 8791, so `login` gets past credential verification — same trick as `tests/pty.rs::spawn_whoami_server`.

```bash
#!/usr/bin/env bash
# Acceptance probes for PR #236. Run from the repo root after `cargo build -p tvc`.
set -uo pipefail

TVC=./target/debug/tvc
FIX=$(mktemp -d)
ORG="$FIX/.config/turnkey/orgs/alias-a"
mkdir -p "$ORG"

# A real p256 API key, so `login` gets past credential loading.
openssl ecparam -name prime256v1 -genkey -noout -out "$FIX/k.pem" 2>/dev/null
PRIV=$(openssl ec -in "$FIX/k.pem" -text -noout 2>/dev/null \
  | awk '/priv:/{f=1;next}/pub:/{f=0}f' | tr -d ' :\n')
PUB=$(openssl ec -in "$FIX/k.pem" -pubout -conv_form compressed -outform DER 2>/dev/null \
  | xxd -p | tr -d '\n' | tail -c 66)
python3 - "$ORG" "$PRIV" "$PUB" <<'PY'
import json, os, secrets, sys
org, priv, pub = sys.argv[1], sys.argv[2].rjust(64, "0"), sys.argv[3]
json.dump({"public_key": pub, "private_key": priv, "curve": "p256"},
          open(os.path.join(org, "api_key.json"), "w"))
# A well-formed operator key: 130-byte public, 32-byte seed.
json.dump({"public_key": secrets.token_hex(130), "private_key": secrets.token_hex(32)},
          open(os.path.join(org, "operator.json"), "w"))
PY

cat > "$FIX/.config/turnkey/tvc.config.toml" <<EOF
version = 1
active_org = "alias-a"

[orgs.alias-a]
id = "org-e2e"
api_key_path = "$ORG/api_key.json"
api_base_url = "http://127.0.0.1:8791"
default_operator_kind = "local"

[[orgs.alias-a.operators]]
name = "default"
kind = "local"
key_path = "$ORG/operator.json"
EOF

echo "### 1. no --output, stdin is a pipe, TVC_NON_INTERACTIVE unset"
HOME="$FIX" $TVC keys backup-operator-key < /dev/null 2>&1; echo "   exit=$?"

echo
echo "### 2. --output exists, stdin is a pipe, TVC_NON_INTERACTIVE unset"
echo previous > "$FIX/bk.json"
HOME="$FIX" $TVC keys backup-operator-key --output "$FIX/bk.json" < /dev/null 2>&1; echo "   exit=$?"

echo
echo "### 3. permissions of a written backup"
rm -f "$FIX/bk2.json"
HOME="$FIX" TVC_NON_INTERACTIVE=1 $TVC keys backup-operator-key --output "$FIX/bk2.json" >/dev/null 2>&1
ls -l "$FIX/bk2.json" | awk '{print "   "$1"  "$NF}'

echo
echo "### 4. operator key whose private key cannot derive its public key"
python3 -c "
import json,secrets
json.dump({'public_key':secrets.token_hex(130),'private_key':'not-even-hex'},
          open('$ORG/operator.json','w'))"
rm -f "$FIX/bk3.json"
HOME="$FIX" TVC_NON_INTERACTIVE=1 $TVC keys backup-operator-key --output "$FIX/bk3.json" 2>&1 | head -2
echo "   exit=${PIPESTATUS[0]}"
python3 -c "
import json,secrets
json.dump({'public_key':secrets.token_hex(130),'private_key':secrets.token_hex(32)},
          open('$ORG/operator.json','w'))"

echo
echo "### 5. login: cancel the backup prompt after the key is already saved"
python3 - <<'PY' &
from http.server import BaseHTTPRequestHandler, HTTPServer
B = b'{"organizationId":"org-e2e","organizationName":"E2E Org","userId":"user-1","username":"e2e"}'
class H(BaseHTTPRequestHandler):
    def do_POST(self):
        self.rfile.read(int(self.headers.get("content-length", 0)))
        self.send_response(200)
        self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(B)))
        self.end_headers()
        self.wfile.write(B)
    def log_message(self, *a): pass
HTTPServer(("127.0.0.1", 8791), H).serve_forever()
PY
SERVER=$!
sleep 1
rm -f "$ORG/operator.json"
OUT=$(printf '\x04' | HOME="$FIX" script -q /dev/null $TVC login --org alias-a 2>&1)
CODE=$?
echo "   exit=$CODE"
echo "$OUT" | tr -d '\r' | tail -1 | sed 's/^/   /'
echo "$OUT" | grep -q "Successfully logged in" \
  && echo "   reported success: yes" || echo "   reported success: NO"
test -f "$ORG/operator.json" \
  && echo "   operator key on disk: yes (login had already succeeded)" \
  || echo "   operator key on disk: no"
kill $SERVER 2>/dev/null
rm -rf "$FIX"
```

</details>

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: I think I disagree with all the robots here. This should be a read and copy. We shouldn't have to think about perms for backup, just copy the original file's perms. Also, `copy` as atomic and allows the OS to handle the copy in the most efficient way

richardpringle: b6f864800

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: Yeah, we also don't use zeorize in here. I think this is a larger refactor though

tvc/src/config/turnkey/qos_operator_key.rs
richardpringle: Yeah the suggestion here is to do what's already captured in TVC-270. The key is that we should use `bytemuck` with `repr(packed)` but it's a new dependency.  I don't think that work belongs in this PR.

tvc/src/config/turnkey/qos_operator_key.rs
richardpringle: I think I left a comment on this in my own PR. I agree with your agent here

richardpringle: Oh... changed this and realized why it was doing what it was doing. I actually prefer the original code. Derive implementations are usually safer. 

The better way to do this would be to create an `AsHex<const N: usize>([u8; N])` wrapper that encodes the array for both `Debug` and `Display`. I'll create a ticket, this doesn't need to happen now

tvc/src/config/turnkey/qos_operator_key.rs
richardpringle: Will create a ticket

tvc/tests/common.rs
richardpringle: 54dea51e6

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: Tracked in https://linear.app/turnkey/issue/TVC-278/

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: Tracked in https://linear.app/turnkey/issue/TVC-278/

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: Tracked in https://linear.app/turnkey/issue/TVC-278/

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: Tracked in https://linear.app/turnkey/issue/TVC-278/

tvc/src/commands/login.rs
richardpringle: Tracked in https://linear.app/turnkey/issue/TVC-278/

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: d253f72e

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: d253f72e

tvc/src/commands/login.rs
richardpringle: 8460f9a9

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: d253f72e

tvc/src/commands/login.rs
richardpringle: 0fe438bf

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: Re-signed: 91ddd644

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: Re-signed: 91ddd644

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: Re-signed: 91ddd644

tvc/src/commands/login.rs
richardpringle: Re-signed: 268091e5

tvc/src/commands/login.rs
richardpringle: Re-signed: 268091e5

tvc/src/commands/login.rs
richardpringle: Done: 91ddd644 (shared prompt_for_backup_destination), 268091e5 (login call site)

tvc/src/commands/keys/backup_operator_key.rs
richardpringle: Tracked in https://linear.app/turnkey/issue/TVC-283/

tvc/src/config/turnkey/qos_operator_key.rs
richardpringle: Tracked in https://linear.app/turnkey/issue/TVC-284/

tvc/src/commands/login.rs
richardpringle: Tracked in https://linear.app/turnkey/issue/TVC-282/

tvc/tests/pty.rs
richardpringle: Tracked in https://linear.app/turnkey/issue/TVC-280/

tvc/src/config/turnkey/qos_operator_key.rs
richardpringle: Keeping the break: login is the only writer of operator.json and always emits the full 130-byte composite, so anything else is hand-edited or corrupt and should surface loudly rather than ride along.

tvc/src/config/turnkey/qos_operator_key.rs
richardpringle: Tracked in https://linear.app/turnkey/issue/TVC-285/

tvc/src/config/turnkey/qos_operator_key.rs
```diff
@@ -2,14 +2,100 @@
...
+}
+
+impl_hex_debug!(QosOperatorPublicKey);
+
+/// The nil key. Exists for the payload-enumeration tests on [`crate::outcome::Outcome`],
+/// which construct every outcome shape via `Default`.
+impl Default for QosOperatorPublicKey {
+    fn default() -> Self {
+        // Not `Self(Default::default())`: std's array `Default` stops at 32
+        // elements, so the repeat expression delegates per element instead.
```
daniilrrr, line 88: dont think this comment is still relevant?

tvc/src/local_operator_key.rs
```diff
@@ -147,7 +147,8 @@ mod tests {
         fs::write(
             &path,
             serde_json::to_string(&StoredQosOperatorKey {
-                public_key: "unused".to_string(),
+                // The resolve path only reads the seed; a nil key stands in.
```
daniilrrr, line 150: rm redundant comment

tvc/tests/common.rs
```diff
@@ -0,0 +1,100 @@
+//! Fixtures shared across the tvc integration-test binaries.
+//!
+//! Each test binary that declares `mod common;` compiles its own copy and
+//! uses a subset of these helpers, so unused items are expected in every
+//! build of this file; the crate-level allow keeps those builds quiet.
```
daniilrrr, line 5: last phrase in the paragraph is definitely false and should be removed

i dont think any of this comment is true, I see every func being used and no dead code

tvc/src/commands/keys/backup_operator_key.rs
daniilrrr: this is my biggest concern in the PR, because it's an easy fix to delay, forget about, then accidentally log or leak private key info. Any chance we can store it as `HexSeed` in this PR? If not, can TVC-278 be like an immediate follow ticket?

richardpringle: Yeah, I actually already did a bunch of TVC-278 in here, just not this one - so can follow up right after. The issue that's described here is that we don't properly validate the key before backing it up, but it's a corner case. We shouldn't be backing up any keys that weren't created by the cli to begin with. You'd have to go in and manually manipulate the key. 

I agree that it should happen and we don't want to keep the key as a string, etc etc... but all these things existed before this ticket. I'm not making the security posture worse with this backup mechanism

daniilrrr (approved):

r-n-o (approved):
+1 from a dependency-reviewers POV. No new deps ✅

tvc/src/config/turnkey/qos_operator_key.rs
richardpringle: f6294086

tvc/src/local_operator_key.rs
richardpringle: f6294086

tvc/tests/common.rs
richardpringle: f6294086

PR #237, opened 2026-08-10
refactor(tvc): seal operator backends behind a Signer port

richardpringle:
Second in the stack, extracted from #224 — based on #236; retarget to main when it merges.

- Operator semantics move behind one seam: `trait Signer { sign, public_key }` in `pair.rs`, with `LocalPair` and `HostedSigner` adapters (the hosted signer owns its `AuthenticatedClient` and signs via `sign_raw_payload`).
- `operator.rs` splits into `operator/mod.rs` (resolution, `ResolvedOperator`, manifest approval) and `operator/hosted.rs` (creation, registry resolution, the API-signing adapter).
- `approve_manifest` no longer takes a ctx or matches on kind — `OperatorCtx`, `ResolvedOperatorKind`, `is_hosted()`, and `ValidatedHostedOperatorRecord` are gone (the latter absorbed into `ResolvedHostedOperator`, keys parsed once at resolution).
- `resolve_operator` preserves main's precedence exactly (rustdoc-documented, no crossover between backends) and now takes a `SignerRequirement { Any, OfflineApproval }`, so `--skip-post` with a hosted operator is refused before any credentials load — pinned by the existing `hosted_operator_rejects_skip_post_before_api_activity` test.
- Local operator selection becomes `OrgConfig::select_local_operator`, returning the `(record, local)` pair with a typed `SelectLocalOperatorError`; callers add org context via `with_context`.
- `OrgConfig::select_local_record` and the default-kind gate are gone — file-management commands no longer check the org's default backend, an accepted behavior change.
- Known message-level deviations (outcomes identical): selector errors are reworded with call-site org context, and `--skip-post` + hosted ID + broken credentials now surfaces the credential error.

### DIFF
--- a/tvc/src/commands/deploy/approve.rs
+++ b/tvc/src/commands/deploy/approve.rs
@@ -7,7 +7,7 @@ use crate::{
     config::turnkey::Config,
     errors::MissingResource,
     local_operator_key::LocalOperatorSeedSource,
-    operator::{OperatorCtx, ResolvedOperator, resolve_operator},
+    operator::{SignerRequirement, resolve_operator},
     outcome::Outcome,
     output::StdCtx,
     pair::HexSeed,
@@ -481,24 +481,22 @@ async fn run_with_resolved_inputs(
             }
         });
 
-    let operator = resolve_operator(inputs.operator_seed_source, inputs.operator_id).await?;
-    if inputs.skip_post && operator.is_hosted() {
-        bail!("--skip-post is only supported for local operators");
-    }
-    let auth = if operator.is_hosted() {
-        Some(build_client().await?)
+    let requirement = if inputs.skip_post {
+        SignerRequirement::OfflineApproval
     } else {
-        None
+        SignerRequirement::Any
     };
-    let approval = sign_and_write_approval(
-        &operator,
-        &OperatorCtx {
-            auth: auth.as_ref(),
-        },
-        inputs.approval_out.as_deref(),
-        &inputs.manifest,
-    )
-    .await?;
+    let operator =
+        resolve_operator(inputs.operator_seed_source, inputs.operator_id, requirement).await?;
+
+    let approval = operator.approve_manifest(&inputs.manifest).await?;
+
+    // Reporting the approval (inline payload or file path) is the terminal
+    // outcome's job; `--approval-out` additionally persists it here.
+    if let Some(path) = inputs.approval_out.as_deref() {
+        let json = serde_json::to_string_pretty(&approval)?;
+        write_file(path, &json).await?;
+    }
 
     match inputs.post_target {
         Some(target) => {
@@ -554,25 +552,6 @@ async fn load_manifest(
         }
         (None, None) => bail!("a manifest source is required"),
     }
-}
-
-/// Sign the manifest and, when `--approval-out` is set, write the approval to
-/// that file. Reporting the approval (inline payload or file path) is the
-/// terminal outcome's job.
-async fn sign_and_write_approval(
-    operator: &ResolvedOperator,
-    operator_ctx: &OperatorCtx<'_>,
-    approval_out: Option<&Path>,
-    manifest: &ValidatedManifest<'_>,
-) -> anyhow::Result<Approval> {
-    let approval = operator.approve_manifest(operator_ctx, manifest).await?;
-
-    if let Some(path) = approval_out {
-        let json = serde_json::to_string_pretty(&approval)?;
-        write_file(path, &json).await?;
-    }
-
-    Ok(approval)
 }
 
 #[instrument(skip_all, fields(manifest_id = %plan.manifest_id, operator_id = %plan.operator_id, deploy_id = ?plan.deploy_id))]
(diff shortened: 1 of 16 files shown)


tvc/src/commands/deploy/approve.rs
```diff
@@ -481,24 +481,22 @@ async fn run_with_resolved_inputs(
...
-            auth: auth.as_ref(),
-        },
-        inputs.approval_out.as_deref(),
-        &inputs.manifest,
-    )
-    .await?;
+    let operator =
+        resolve_operator(inputs.operator_seed_source, inputs.operator_id, requirement).await?;
+
+    let approval = operator.approve_manifest(&inputs.manifest).await?;
```
annieke, line 492: i like this

tvc/src/operator/hosted.rs
```diff
@@ -550,27 +262,51 @@ pub(crate) fn resolve_hosted_operator_encrypt_key(
...
 }
 
-fn hosted_signing_intent(sign_with: String, manifest_hash: &[u8]) -> SignRawPayloadIntentV2 {
-    SignRawPayloadIntentV2 {
-        sign_with,
-        payload: hex::encode(manifest_hash),
-        encoding: PayloadEncoding::Hexadecimal,
-        hash_function: HashFunction::Sha256,
+impl Signer for HostedSigner {
+    fn sign(&self, message: &[u8]) -> SignerFuture<'_, Result<Vec<u8>>> {
```
annieke, line 310: i'm not sure if this is the ultimate API for hosted operators, since my plan is to rip out the `SignRawPayload` and replace it with an activity which is dedicated to approving a manifest and signing in an enclave only that shape

tvc/src/operator/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+    name: Option<String>,
+    /// Always present for hosted operators and optional for local operators.
+    operator_id: Option<Uuid>,
+    signer: Box<dyn Signer>,
+}
+
+impl ResolvedOperator {
+    pub(crate) fn name(&self) -> Option<&str> {
+        self.name.as_deref()
+    }
```
annieke: fyi operator names are not enforced to be unique per org, and i do have it as optional in the cli with a static default

annieke (commented):
pretty coarse review since i didn't look at the first stack too deeply. overall looks like a worthy cleanup; only potential issue is the generic `sign` for hosted operators

tvc/src/operator/hosted.rs
richardpringle: That's okay, I think we can absolutely change it if we need to, but we should have something that's generic that captures the shared logic between types of operators.

tvc/src/operator/mod.rs
richardpringle: Should we put a static default in here?

tvc/src/commands/keys/re_encrypt_local_share.rs
```diff
@@ -129,7 +129,9 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> anyhow::Result<Outcome> {
 async fn build_re_encrypted_share_output(
     quorum_key_metadata: &QuorumKeyMetadata,
     provision_bundle: &ProvisionBundle,
-    operator_pair: &dyn Pair,
+    // Concretely local: re-encryption decrypts the share, and decryption
+    // needs local key material.
+    operator_pair: &LocalPair,
```
annieke, line 134: v helpful note

tvc/src/operator/mod.rs
annieke: it's these:
https://github.com/tkhq/rust-sdk/blob/3a4dc456433defea2a8e454772b7e267fe311ca8/tvc/src/commands/operator/create.rs#L21-L22

tvc/src/operator/hosted.rs
annieke: what do you propose changing this to if we migrate to an `APPROVE_TVC_MANIFEST` activity? rip out this function as part of the implementation?

i won't block on this but would add a plan in the comments since it's planned work

annieke (commented):
hesitant about the `sign` function on the hosted signer (see comment). not blocking but let's document a path forward for `Signer` for removing the generic `sign_raw_payload` call since that's already planned.

tvc/src/operator/mod.rs
richardpringle: Ah, this code isn't super clear. The only place that function was used was for displaying an error. I got rid of the function and I think it's fine to keep the property of the type as is for now. If it's clear to you that it should be different, open a PR? 😅

tvc/src/operator/hosted.rs
richardpringle: Seems like the right abstraction is 
```rust
trait SignManifest {
    fn sign(&self, manifest: &VerifiedManifest<'_>) -> Vec<u8>; 
    // ...
}
```

Since we are moving from more generic to more specific, I think it make sense to leave this as is for now and make the change when we know what the final shape of the activity is (and that `VerifiedManifest` contains all the necessary data)

annieke (approved):

PR #239, opened 2026-08-10
feat(tvc): hosted-operator parity for login, create defaults, and approvals

richardpringle:
Third in the stack, extracted from #224 — based on #237; retarget when it merges. This is the fix for login being impossible when an org has no local operator.

## Sole-hosted resolution

`select_hosted_operator` (typed `SelectHostedOperatorError`, sibling of the local selector) resolves the sole hosted registry record. `resolve_operator`'s hosted-default arm no longer demands `--operator-id`: with none given it refuses `--skip-post` before selection or credentials, then resolves the sole hosted record — no crossover to the local key, precedence otherwise unchanged. Dead `OperatorRecord::operator_kind()` deleted. Per review comments on #224: hosted registry iteration is now `OrgConfig::hosted_operators()`, and both hosted lookups go through it.

## Registered hosted operators in defaults

`Config::default_operator_public_key()` (a method per review feedback) honors the org's default backend — local reads the sole key file, hosted joins the sole record's stored points — replacing the three copies of the local-only helper in `app init`, `app create`, and `keys init-local-quorum-key`. `Config::known_operator_candidates()` unions registered hosted operators (config order) with the last `app create`'s saved IDs, deduplicated; `app create` reuses over that union by default (`--no-operator-reuse` still mints), and `deploy approve` sources its posting candidate from it. Flagged behavior change: a local-default org with a registered hosted record and no saved IDs now auto-posts via the hosted identity instead of bailing.

## Hosted-default login

`execute_login` matches on `default_operator_kind` with no crossover: local finds or generates the registered key file (backup nudge unchanged); hosted reports the registered identity and generates nothing. The JSON outcome flattens a `LoggedInOperator` enum — local keeps its exact pre-hosted fields plus an additive `operatorKind` tag; hosted emits `operatorName`/`operatorId`/composite `operatorPublicKey` and no `operatorKeyPath`. Covered by complete-JSON serde tests and a mock-whoami pty login test.

## Local-only commands explain themselves

For hosted-only orgs, `keys backup-operator-key` explains the private key is held by Turnkey and cannot be exported, and `keys re-encrypt-local-share` points at `tvc deploy provision`; typed errors are preserved through the chains. Shared `write_hosted_only_config` fixture in `tests/common.rs`.

One review comment deliberately deferred: "no `Config::load` in the operator module" — `resolve_operator` keeps its lazy load because `explicit_seed_does_not_load_malformed_config` pins that explicit-seed approvals must not depend on a well-formed config file; documented in its rustdoc.

### DIFF
--- a/tvc/src/commands/app/create.rs
+++ b/tvc/src/commands/app/create.rs
@@ -86,13 +86,18 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
         OperatorReuse::KeepConfig => {}
         OperatorReuse::Reuse(operator) => apply_operator_reuse(ctx, &mut app_config, operator)?,
         OperatorReuse::MultipleCandidates(operators) => {
-            if ctx.is_non_interactive() {
+            // Picking a candidate needs a prompt, which is impossible under
+            // --non-interactive, JSON mode, or a non-TTY stdin.
+            let can_prompt = !ctx.is_non_interactive() && prompts::stdin_can_prompt();
+
+            if !can_prompt {
                 bail!(
                     "multiple operator IDs are known for the active org; \
                      set manifestSetParams.existingOperatorIds in your config to reuse one, \
                      or pass --no-operator-reuse to create a new operator"
                 );
             }
+
             let operator = prompts::select("Select operator to reuse", operators)?;
             apply_operator_reuse(ctx, &mut app_config, operator)?;
         }
--- a/tvc/src/commands/app/create.rs
+++ b/tvc/src/commands/app/create.rs
@@ -4,8 +4,9 @@ use crate::{
     client::build_client,
     config::{
         app::{AppConfig, AppConfigValidationErrors, OperatorSetParams},
-        turnkey::{self, StoredQosOperatorKey},
+        turnkey,
     },
+    operator::OperatorCandidate,
     outcome::Outcome,
     output::{Ctx, StdCtx},
     prompts, shell_println,
@@ -31,11 +32,12 @@ pub struct Args {
     #[arg(short = 'c', long, value_name = "PATH", env = "TVC_APP_CONFIG")]
     pub config_file: PathBuf,
 
-    /// Create a new operator instead of reusing the most recently created one.
+    /// Create a new operator instead of reusing a known one.
     ///
-    /// By default `app create` reuses the operator from your last `app create`
-    /// (the same local operator key) rather than minting a new operator ID each
-    /// time. Pass this to force creating a new operator.
+    /// By default `app create` reuses a known operator — a registered hosted
+    /// operator, or the one from your last `app create` — rather than minting
+    /// a new operator ID each time. Pass this to force creating a new
+    /// operator.
     #[arg(long, env = "TVC_NO_OPERATOR_REUSE")]
     pub no_operator_reuse: bool,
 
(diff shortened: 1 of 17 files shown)


tvc/src/commands/app/create.rs
```diff
@@ -64,28 +66,35 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> Result<Outcome> {
...
         app_config.manifest_set_params.as_ref(),
-        &saved_ids,
+        &candidates,
     ) {
         OperatorReuse::KeepConfig => {}
-        OperatorReuse::Reuse(id) => apply_operator_reuse(ctx, &mut app_config, id)?,
-        OperatorReuse::MultipleCandidates(ids) => {
+        OperatorReuse::Reuse(operator) => apply_operator_reuse(ctx, &mut app_config, operator)?,
+        OperatorReuse::MultipleCandidates(operators) => {
             if ctx.is_non_interactive() {
```
richardpringle: Do I need `can_prompt`?

tvc/src/commands/keys/re_encrypt_local_share.rs
```diff
@@ -113,7 +114,20 @@ pub async fn run(ctx: &mut StdCtx, args: Args) -> anyhow::Result<Outcome> {
...
+    let operator_pair = resolve_local_operator(operator_seed_source)
+        .await
+        .map_err(|error| {
+            // Decryption needs local key material a hosted-only org does not
+            // have; point at the hosted counterpart of this operation.
+            match error.downcast_ref::<SelectLocalOperatorError>() {
+                Some(SelectLocalOperatorError::NoLocalOperator) => error.context(
+                    "re-encrypting a local share needs the local operator key the share was \
+                     encrypted to; hosted operators re-encrypt through Turnkey with \
+                     `tvc deploy provision`",
```
richardpringle, line 126: @annieke can you confirm that this is accurate?

tvc/src/commands/login.rs
```diff
@@ -703,10 +729,42 @@ pub struct LoggedIn {
...
+
+/// Local, like every org `login` itself creates; exists for [`LoggedIn`]'s
+/// `Default`, which the outcome payload-enumeration tests construct.
+impl Default for LoggedInOperator {
+    fn default() -> Self {
+        Self::Local {
+            operator_public_key: QosOperatorPublicKey::default(),
+            operator_key_path: String::new(),
+        }
+    }
```
richardpringle, line 767: this was a weird one. I believe it's because all the `Outcome`s need `Default`

tvc/src/commands/keys/re_encrypt_local_share.rs
annieke: yup!

tvc/tests/common.rs
```diff
@@ -1,4 +1,6 @@
-//! Fixtures shared across the tvc integration-test binaries.
+//! Fixtures shared across the tvc integration-test binaries. Each binary
+//! compiles its own copy and uses a subset, so unused items are expected.
+#![allow(dead_code)]
```
richardpringle, line 3: this is actually needed this time

tvc/src/commands/app/create.rs
```diff
@@ -31,11 +32,12 @@ pub struct Args {
...
     ///
-    /// By default `app create` reuses the operator from your last `app create`
-    /// (the same local operator key) rather than minting a new operator ID each
-    /// time. Pass this to force creating a new operator.
+    /// By default `app create` reuses a known operator — a registered hosted
+    /// operator, or the one from your last `app create` — rather than minting
+    /// a new operator ID each time. Pass this to force creating a new
+    /// operator.
     #[arg(long, env = "TVC_NO_OPERATOR_REUSE")]
     pub no_operator_reuse: bool,
```
annieke, line 42: just a thought to myself: this flag feels kind of like a relic of testing to me and is confusing when left in. not within scope of this PR to remove it but just jotting thoughts

tvc/src/commands/login.rs
```diff
@@ -361,15 +362,44 @@ async fn execute_login(ctx: &mut StdCtx, mut config: Config, plan: LoginPlan) ->
...
+                .select_local_operator()
+                .with_context(|| format!("org '{alias}'"))?;
+            let operator_key = find_or_generate_operator_key(ctx, &alias, local).await?;
+
+            LoggedInOperator::Local {
+                operator_public_key: operator_key.public_key,
+                operator_key_path: local.key_path.display().to_string(),
+            }
+        }
+        OperatorKind::Hosted => {
```
annieke, line 382: note for myself: if we default to hosted we would need its version of `find_or_create_hosted_operator` and probably need to re-order the api key creation step when i make this change

tvc/src/operator.rs
```diff
@@ -163,8 +163,13 @@ pub(crate) enum SignerRequirement {
 ///    operator — even when the org's default operator kind is local.
 /// 3. Otherwise the active org's `default_operator_kind` decides, and never
 ///    crosses over: `local` resolves the sole local record (reconciling a
-///    given ID against its configured one); `hosted` refuses — an ID is
-///    required, and resolution never falls back to the local key.
+///    given ID against its configured one); `hosted` resolves the sole
+///    hosted record when no ID names one, and never falls back to the
+///    local key.
```
annieke, line 168: v helpful comment

tvc/src/operator.rs
```diff
@@ -220,7 +225,26 @@ pub(crate) async fn resolve_operator(
     if org.default_operator_kind == OperatorKind::Hosted {
         match operator_id {
             Some(id) => bail!("hosted operator ID '{id}' was not found in org '{alias}'"),
-            None => bail!("--operator-id is required to approve with a hosted operator"),
+            None => {
+                // A hosted default cannot satisfy an offline approval;
+                // refuse before selection or credentials.
+                if requirement == SignerRequirement::OfflineApproval {
+                    bail!("--skip-post is only supported for local operators");
+                }
```
annieke, line 233: 💯

tvc/tests/common.rs
```diff
@@ -59,6 +61,58 @@ pub fn write_profiles_config(home: &Path, profiles: &[(&str, &str)], active_org:
     .unwrap();
 }
 
+/// Write a v1 `tvc.config.toml` under `home` whose sole, active organization
+/// defaults to the hosted backend and registers exactly one operator, hosted
+/// — the fixture for commands that need local key material a hosted-only org
+/// does not have. The record carries a real generated composite key split
+/// into its two points, the way `operator create` stores it.
+pub fn write_hosted_only_config(home: &Path, alias: &str, org_id: &str) {
```
annieke, line 69: 🙏

annieke (approved):
this change wasn't easy to scan so def used AI to parse; lmk if you wanted readability feedback because I mostly focused on functionality.

this is a big improvement to the original local-only defaulting (THANK YOU for this), and to my understanding
- hosted defaulting only works if there's one hosted operator (this is ok, maybe one day we can set a default but i don't even love that idea since defaulting code gets so complex)
- multiple operators requires explicit id or interactive selection (👍 )
- an org that's set to `hosted` will default to the sole hosted operator if none is specified

all this behavior is an improvement to me. reviewing mostly just highlighted how spaghetti original defaulting was 🥲

PR #245, opened 2026-08-13
refactor(tvc): load config once at dispatch and pass it into commands

richardpringle:
Removes every `Config::load()` call site. The CLI builds the config once in `Commands::run` (loading `tvc.config.toml`, or writing the default on first run) and passes it down: command endpoints take an owned `Config` (matching the `Run` trait), shared helpers (`build_client`, `resolve_operator`, `resolve_local_operator`, ...) take `&Config`.

Behavior changes that fall out of the top-level load, encoded in the updated integration tests:
- every command now requires `HOME` and a well-formed config before dispatch, including env-auth (CI) runs and offline paths (`deploy approve --operator-seed --skip-post`, `deploy init`)
- a malformed config file fails fast with a parse error instead of being ignored on paths that never read it
- first run writes the default config file

### DIFF
--- a/tvc/src/config/turnkey/mod.rs
+++ b/tvc/src/config/turnkey/mod.rs
@@ -23,8 +23,8 @@ use thiserror::Error;
 use tracing::debug;
 use uuid::Uuid;
 
-const CONFIG_DIR: &str = ".config/turnkey";
-const CONFIG_FILE: &str = "tvc.config.toml";
+pub const CONFIG_DIR: &str = ".config/turnkey";
+pub const CONFIG_FILE: &str = "tvc.config.toml";
 const ORGS_DIR: &str = "orgs";
 const API_KEY_FILE: &str = "api_key.json";
 const OPERATOR_KEY_FILE: &str = "operator.json";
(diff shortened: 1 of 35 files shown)


tvc/tests/deploy_approve.rs
```diff
@@ -161,7 +161,10 @@ fn hosted_operator_rejects_skip_post_before_api_activity() {
 }
 
 #[test]
-fn explicit_seed_does_not_load_malformed_config() {
+fn malformed_config_fails_before_dispatch_even_with_explicit_seed() {
```
richardpringle, line 189: This test changed. It doesn't make sense to be able to run some commands but not others. The majority of the cli depends on a well-formed config, you shouldn't be able to circumvent that.

tvc/src/config/turnkey/mod.rs
```diff
@@ -400,25 +400,8 @@ pub fn default_operator_key_path(alias: &str) -> Result<PathBuf> {
...
-            config_path = %path.display(),
-            active_org = ?config.active_org,
-            org_count = config.orgs.len(),
-            "loaded tvc config"
-        );
-
-        Ok(config)
+    pub fn from_toml(content: &str) -> Result<Self> {
+        disk::from_toml(content)
     }
```
daniilrrr, line 434: what does this change do? 

 and it looks like the existing code moved to `cli.rs`?

daniilrrr (dismissed):
great change

tvc/src/config/turnkey/mod.rs
richardpringle: It was just part of the `load` call. I know load the file contents in cli.rs and pass them to this file. Why? The `Config` type shouldn't be coupled to where it's stored. The only place that we care about that is in the bootstrapping of the cli application.

daniilrrr (approved):

PR #250, opened 2026-08-14
Improve client transport behavior

emostov:
Closes ENG-4611

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -3,7 +3,8 @@ use serde::Serialize;
 use serde::de::DeserializeOwned;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-use reqwest::header::CONTENT_TYPE;
+use reqwest::StatusCode;
+use reqwest::header::{CONTENT_TYPE, LOCATION};
 use thiserror::Error;
 
 use generated::Activity;
@@ -72,6 +73,9 @@ pub use retry::RetryConfig;
 const TURNKEY_RUST_SDK_USER_AGENT: &str =
     concat!("turnkey-rust-client/", env!("CARGO_PKG_VERSION"));
 
+/// Maximum number of redirects followed for a single request.
+const MAX_REDIRECTS: usize = 10;
+
 #[derive(Debug, Error)]
 pub enum TurnkeyClientError {
     #[error("Client builder is missing its API key. Call .api_key(...) to configure it.")]
@@ -128,6 +132,9 @@ pub enum TurnkeyClientError {
     #[error("Maximum number of attempts reached (after {0} retries)")]
     ExceededRetries(usize),
 
+    #[error("Refused to follow redirect (status {0}) to {1}")]
+    RefusedRedirect(u16, String),
+
     #[error("Stamper error")]
     StamperError(#[from] StamperError),
 }
@@ -155,7 +162,7 @@ impl<S: Stamp> TurnkeyClientBuilder<S> {
             api_key: None,
             base_url: None,
             retry_config: None,
-            reqwest_builder: reqwest::Client::builder(),
+            reqwest_builder: reqwest::Client::builder().redirect(same_origin_redirect_policy()),
             timeout: None,
         }
     }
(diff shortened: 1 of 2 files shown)


client/src/lib.rs
```diff
@@ -231,6 +238,40 @@ impl<S: Stamp> Default for TurnkeyClientBuilder<S> {
     }
 }
 
+/// The redirect policy applied to every client built by [`TurnkeyClientBuilder`].
+///
+/// Requests carry the API key stamp and the full activity payload. `reqwest` only strips
+/// `Authorization` and `Cookie` when a redirect crosses origins, so without this policy a
+/// redirect could hand a valid stamp and its signed body to another host. Redirects are
+/// therefore only followed back to the origin of the *original* request, which also stops a
+/// server from walking the client off-origin one same-origin-looking hop at a time.
```
r-n-o, line 247: A stamp isn't sensitive like normal bearer tokens are and this comment implies that it is.

I'm not against defense-in-depth though, so the code can stay.

r-n-o (commented):
I'm sort of "meh" about this diff. The code as-is doesn't seem to be vulnerable to anything bad given X-Stamp headers aren't sensitive.

In fact this behavior (following cross-origin redirect) could be helpful if you're configuring your client to target a TVC app, and the TVC app then responds with a 307 back to api.turnkey.com? (I don't think that's how we want to implement co-signers but...this diff sort of closes this option forever). Another thing that 307 can be helpful for is domain migration or multi-region. Likely we'll implement this at the DNS level though. Anyway.

Bottom-line: if we need to merge this because security reasons, sure let's do it. But a) it's extra code to maintain b) the behavior change doesn't address immediate security concerns afaict c) I have questions in my mind around whether it closes possibilities down the road or not.

lreading-turnkey:
> I'm sort of "meh" about this diff. The code as-is doesn't seem to be vulnerable to anything bad given X-Stamp headers aren't sensitive.
> 
> In fact this behavior (following cross-origin redirect) could be helpful if you're configuring your client to target a TVC app, and the TVC app then responds with a 307 back to api.turnkey.com? (I don't think that's how we want to implement co-signers but...this diff sort of closes this option forever). Another thing that 307 can be helpful for is domain migration or multi-region. Likely we'll implement this at the DNS level though. Anyway.
> 
> Bottom-line: if we need to merge this because security reasons, sure let's do it. But a) it's extra code to maintain b) the behavior change doesn't address immediate security concerns afaict c) I have questions in my mind around whether it closes possibilities down the road or not.

Your concern is fair, especially for future TVC designs. I also agree that the current wording overstates the sensitivity of X-Stamp. 

I do see it as a sane strategy for defense in depth though, not an urgent MITM mitigation. From what I can tell, this implementation only blocks 307/308 to other origins. I like that idea because a redirect can include the whole body _as well as_ a valid signature for that specific payload. 

If TVC needs a more permissive cross-origin redirect later, we can either bake that into the SDK as an opt-in, or they could override the default with the _reqwest_ builder. I think making it opt-in is the safer decision.

I went a _very_ different path in https://github.com/tkhq/sdk/pull/1490 - which blocks all redirects wholesale. After thinking this through more, I think that's the wrong move.

Does that sound reasonable, or are you still leaning towards not merging this @r-n-o?

r-n-o:
> a redirect can include the whole body as well as a valid signature for that specific payload

I think that would be a valid concern in standard APIs which rely on symmetric auth (bearer tokens), or have private data in POST bodies. Given Turnkey's threat model I'm less concerned; in fact, avoiding this entire class of concerns (MITM or MITM-adjacent essentially) is the whole reason why we designed [Stamping](https://docs.turnkey.com/api-reference/overview/stamps#stamps) the way we did, with asymmetric crypto at its core, and why we don't trust anything outside of secure enclaves (if something sensitive is POSTed to turnkey, it's encrypted client-side, to an enclave-held key! See [e2e channels](https://docs.turnkey.com/security/enclave-secure-channels#enclave-to-end-user-secure-channels)). So I'm a tad "sad" that we have to take on extra code and complexity here.

That being said, I understand the defense-in-depth argument, and I'm not opposed to merging this. It's not that bad, and I don't think TVC has real plans to use 307/308 in its design yet. That was just an reaction off-the-cuff to show that we _are_ closing some doors here. But I'm not sure these are doors we'd care about in the first place. And you're right that we can re-open this possibility later behind specific opt-in flags if required.

r-n-o (approved):

PR #253, opened 2026-08-24
feat(client): minimal singular import_secret / export_secret wrappers

emostov:
## Summary

Adds two client methods for single-secret import and export. Each folds the whole flow into one call; callers pass only business inputs

Tested demo locally - requires secrets enabled for org since it is still in private beta.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4179,8 +4179,10 @@ version = "0.15.0"
 dependencies = [
  "base64 0.22.1",
  "chrono",
+ "hex",
  "http 0.2.12",
  "mime",
+ "p256 0.13.2",
  "prost 0.12.6",
  "prost-types 0.12.6",
  "reqwest",
@@ -4190,6 +4192,7 @@ dependencies = [
  "thiserror 2.0.12",
  "tokio",
  "turnkey_api_key_stamper",
+ "turnkey_enclave_encrypt",
  "wiremock",
 ]
 
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -495,15 +495,13 @@ impl<S: Stamp> TurnkeyClient<S> {
         static_properties: BTreeMap<String, String>,
         signer_quorum_public_key: &QuorumPublicKey,
     ) -> Result<ActivityResult<String>, TurnkeyClientError> {
-        let timestamp_ms = self.current_timestamp();
-
         // `init_import_secrets` is marked INTERNAL in the proto and has no
         // generated client method; submit it directly through `process_activity`.
         let init_activity = self
             .process_activity(
                 generated::external::activity::v1::InitImportSecretsRequest {
                     r#type: "ACTIVITY_TYPE_INIT_IMPORT_SECRETS".to_string(),
-                    timestamp_ms: timestamp_ms.to_string(),
+                    timestamp_ms: self.current_timestamp().to_string(),
                     organization_id: organization_id.clone(),
                     parameters: Some(InitImportSecretsIntent {
                         encryption_suite: TransportEncryptionSuite::EnclaveEncryptV1,
@@ -534,8 +532,11 @@ impl<S: Stamp> TurnkeyClient<S> {
         // `encrypt_secret_bundle` verifies the quorum signature over the target
         // and encrypts to it.
         let target_data = target_data(&target)?;
-        let secret_payload = ImportClient::new(signer_quorum_public_key)
-            .encrypt_secret_bundle(&plaintext, &target, &organization_id)?;
+        let secret_payload = ImportClient::new(signer_quorum_public_key).encrypt_secret_bundle(
+            &plaintext,
+            &target,
+            &organization_id,
+        )?;
         let target_public_key = hex::encode(*target_data.target_public);
 
         let static_properties = static_properties
@@ -546,7 +547,7 @@ impl<S: Stamp> TurnkeyClient<S> {
         let batch = self
             .import_secrets(
                 organization_id,
-                timestamp_ms,
+                self.current_timestamp(),
                 ImportSecretsIntent {
                     secrets: vec![ImportSecretParams {
                         name,
@@ -570,6 +571,10 @@ impl<S: Stamp> TurnkeyClient<S> {
 
     /// Exports secrets and returns their decrypted UTF-8 plaintexts, in the
     /// order of `secret_ids`.
+    ///
+    /// Takes a slice so that any number of secrets exports in a single
+    /// activity. The singular name is only because `export_secrets` is the
+    /// generated method this one is built on.
     pub async fn export_secret(
         &self,
         organization_id: String,
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -1,6 +1,7 @@
 #![doc = include_str!("../README.md")]
 use serde::Serialize;
 use serde::de::DeserializeOwned;
+use std::collections::BTreeMap;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use reqwest::StatusCode;
@@ -10,10 +11,21 @@ use thiserror::Error;
 use generated::Activity;
 use generated::ActivityResponse;
 use generated::ActivityStatus;
+use generated::ExportSecretParams;
+use generated::ExportSecretsIntent;
+use generated::ImportSecretParams;
+use generated::ImportSecretsIntent;
+use generated::InitImportSecretsIntent;
 use generated::external::data::v1::AppProof;
 use generated::google::rpc::Status;
+use generated::immutable::models::v1::{KeyValue, TransportEncryptionSuite};
+use generated::result::Inner as ActivityResultInner;
 
 use turnkey_api_key_stamper::{Stamp, StampHeader, StamperError};
+use turnkey_enclave_encrypt::errors::EnclaveEncryptError;
+use turnkey_enclave_encrypt::{
+    ExportClient, ImportClient, QuorumPublicKey, ServerTargetData, ServerTargetMsgV1,
+};
 
 /// Result of an activity request, containing both the typed result and activity metadata.
 ///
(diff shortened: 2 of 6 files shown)


emostov:
Rework in progress locally on the sandbox but the push credential expired mid-session, so the branch on GitHub still points at `377d287` (the singular batch-wrapper iteration). The follow-up commit that lives locally (`b1ff8b8`) rewires the two helpers to the shape the reviewer asked for:

```rust
pub async fn import_secret(
    &self,
    organization_id: String,
    name: Option<String>,
    plaintext: String,
    static_properties: BTreeMap<String, String>,
    signer_quorum_public_key: &QuorumPublicKey,
) -> Result<ActivityResult<String>, TurnkeyClientError>;

pub async fn export_secret(
    &self,
    organization_id: String,
    secret_id: String,
    signer_quorum_public_key: &QuorumPublicKey,
) -> Result<ActivityResult<String>, TurnkeyClientError>;
```

The helpers now handle timestamp, transport suite selection, target-key generation and validation, HPKE encryption/decryption, and payload serialization internally.

### Also on this local commit

- Rebased onto the current tip of `swag/sync-protos-from-mono-20260813` (`369da54`, the squashed force-push).
- `enclave_encrypt` gains `ServerSecretTargetData`, `EncryptedSecret`, `ImportClient::encrypt_secret_with_bundle`, `ExportClient::decrypt_secret`, `EnclaveEncryptClient::encrypt_secret`, and `EnclaveEncryptServer::publish_secret_target` — the minimum needed to keep the crypto plumbing inside `enclave_encrypt` and out of the client.
- `TurnkeyClientError::EnclaveEncrypt(EnclaveEncryptError)` added and classified explicitly in `tvc::errors` (`CommandError` bucket).
- Two wiremock roundtrip tests exercise the real crypto path: `import_secret_runs_init_encrypt_and_import` and `export_secret_decrypts_returned_payload`.

### Local validation

```
cargo fmt -- --check                     # clean
cargo clippy --all-targets --all-features -- -D warnings   # clean workspace-wide
cargo test --workspace                   # all suites pass (32/32 in turnkey_client, everything else green)
```

Will push as soon as the credential is refreshed.

emostov:
Reworked to the simpler-params shape locally on top of the rebase (`ccef5ca`), but the sandbox is missing the git push credential this session so the PR head hasn't moved yet.

### Local commit `452cbc6` (single commit sitting on top of the base `97c53d4`)

Replaces the two prior commits (`7d90ccc`, `ccef5ca`) with one self-contained commit. Public surface:

```rust
pub async fn import_secret(
    &self,
    organization_id: String,
    name: Option<String>,
    plaintext: String,
    static_properties: BTreeMap<String, String>,
    signer_quorum_public_key: &QuorumPublicKey,
) -> Result<ActivityResult<String>, TurnkeyClientError>;

pub async fn export_secret(
    &self,
    organization_id: String,
    secret_id: String,
    signer_quorum_public_key: &QuorumPublicKey,
) -> Result<ActivityResult<String>, TurnkeyClientError>;
```

`import_secret` runs `init_import_secrets`, HPKE-encrypts the plaintext to the signed ingress target, and submits `import_secrets`. `export_secret` generates a one-shot P-256 recipient key, submits `export_secrets`, and decrypts the returned payload into UTF-8. Timestamp, transport encryption suite, target-key generation/validation, and payload serialization are all internal.

### Files touched (7, +529/-8)

```
 Cargo.lock                    |   3 +
 client/Cargo.toml             |   3 +
 client/src/lib.rs             | 355 +++++++++++++++++++++++++++++++++++++++++
 enclave_encrypt/src/client.rs | 121 +++++++++++++-
 enclave_encrypt/src/lib.rs    |  15 ++
 enclave_encrypt/src/server.rs |  34 +++-
 tvc/src/errors.rs             |   6 +-
```

- `enclave_encrypt`: `ServerSecretTargetData`, `EncryptedSecret`, `ImportClient::encrypt_secret_with_bundle`, `ExportClient::decrypt_secret`, `EnclaveEncryptClient::encrypt_secret` primitive, `EnclaveEncryptServer::publish_secret_target` (for tests). No other changes.
- `client`: two singular helpers + two wiremock roundtrip tests that exercise the full crypto path.
- `tvc/errors`: new `TurnkeyClientError::EnclaveEncrypt` variant classified explicitly in the `CommandError` bucket (client-side crypto never reached the API); satisfies the exhaustive-match rule in `tvc/AGENTS.md`.

### Local validation (toolchain 1.95)

```
cargo fmt -- --check                                            # clean
cargo clippy --all-targets --all-features -- -D warnings        # clean workspace-wide
cargo test --workspace                                          # all suites pass, incl. two new roundtrip tests
```

### Push blocker

Same issue as before: the Valet sandbox does not have a persistent git credential and the GitHub App plugin here doesn't expose a `push_commits` / `create_content` action. As soon as the credential is refreshed I'll force-push `452cbc6` onto `agent/minimal-secret-import-export`.

Cargo.lock
richardpringle, line 1: 👍

richardpringle (dismissed):
deps review

figitaki (commented):
Reviewed with tests run locally (secrets tests pass, clippy clean). One question about the hardcoded empty user ID, plus a few non-blocking nits inline. The tests are strong: real HPKE round-trips, per-secret one-shot target keys, and enclave-side decrypt of the import payload.

client/src/lib.rs
```diff
@@ -411,6 +432,169 @@ impl<S: Stamp> TurnkeyClient<S> {
...
+            init_result.enclave_target_messages,
+            "init import target message",
+        )?;
+
+        // `encrypt_wallet_with_bundle` verifies the quorum signature over the
+        // target and encrypts to it. The user ID in a secrets ingress target is
+        // always empty: the flow is scoped to the organization, not to a user.
+        let target_data = target_data(&target)?;
+        let secret_payload = ImportClient::new(signer_quorum_public_key)
+            .encrypt_wallet_with_bundle(&plaintext, &target, &organization_id, "")?;
```
figitaki: This bakes in the assumption that secrets ingress targets always carry an empty `user_id`. `encrypt_wallet_with_bundle` verifies the signed bundle's `user_id` against this value, so if the signer ever stamps the initiating user into these targets, every `import_secret` call fails with `InvalidUser`.

For comparison, mono's e2e test (`src/go/e2e/public/secrets_test.go`) passes `targetData.UserId` from the bundle through instead of asserting it is empty. Since `target_data()` already parses the bundle here, passing `target_data.user_id` would be forward-compatible. Keeping `""` is a stricter contract assertion, which is also defensible — but it should be a deliberate choice. Is "always empty" guaranteed by the signer, or just current behavior?

client/src/lib.rs
```diff
@@ -411,6 +432,169 @@ impl<S: Stamp> TurnkeyClient<S> {
...
+    /// Imports one UTF-8 secret and returns the created secret ID.
+    pub async fn import_secret(
+        &self,
+        organization_id: String,
+        name: Option<String>,
+        plaintext: String,
+        static_properties: BTreeMap<String, String>,
+        signer_quorum_public_key: &QuorumPublicKey,
+    ) -> Result<ActivityResult<String>, TurnkeyClientError> {
+        let timestamp_ms = self.current_timestamp();
```
figitaki: Nit: this timestamp is reused for the `import_secrets` submit below, which happens after the init activity completes (possibly after polling). Harmless today, but a fresh `current_timestamp()` before the second submit would be more accurate.

client/src/lib.rs
```diff
@@ -411,6 +432,169 @@ impl<S: Stamp> TurnkeyClient<S> {
     }
 }
 
+impl<S: Stamp> TurnkeyClient<S> {
+    /// Imports one UTF-8 secret and returns the created secret ID.
+    pub async fn import_secret(
+        &self,
+        organization_id: String,
+        name: Option<String>,
+        plaintext: String,
```
figitaki, line 494: Nit: the plaintext (and the decrypted exports below) are not zeroized after use. This matches the crate's existing wallet import/export flows, so not a blocker — just noting it since these are long-lived application secrets.

client/src/lib.rs
```diff
@@ -411,6 +432,169 @@ impl<S: Stamp> TurnkeyClient<S> {
...
+            result: secret_id,
+            activity_id: batch.activity_id,
+            status: batch.status,
+            app_proofs: batch.app_proofs,
+        })
+    }
+
+    /// Exports secrets and returns their decrypted UTF-8 plaintexts, in the
+    /// order of `secret_ids`.
+    pub async fn export_secret(
```
figitaki, line 578: Nit: `import_secret` is truly singular but `export_secret` takes a slice and returns `Vec<String>`. I assume the generated `export_secrets` occupies the plural name — fine, but the asymmetry reads oddly. Worth a doc sentence explaining why, if a rename is off the table.

(One property I checked: trusting response order is safe here, because each payload is encrypted to its own one-shot target key — a reordered response fails HPKE decryption rather than silently mis-associating plaintexts.)

client/src/lib.rs
```diff
@@ -411,6 +432,169 @@ impl<S: Stamp> TurnkeyClient<S> {
...
+            ));
+        }
+
+        let plaintexts = secret_payloads
+            .into_iter()
+            .zip(&mut recipients)
+            .map(|(payload, recipient)| {
+                // Secret payloads are enclave-signed bundles over UTF-8 bytes,
+                // the same shape a wallet export bundle carries.
+                recipient.decrypt_wallet_mnemonic_phrase(payload, &organization_id)
```
figitaki: Nit: reusing `decrypt_wallet_mnemonic_phrase` / `encrypt_wallet_with_bundle` for secrets is functionally correct (both are just UTF-8 bytes) and the comments explain it, but a neutral-named alias in `turnkey_enclave_encrypt` (e.g. `decrypt_utf8` / `encrypt_utf8_with_bundle`) would keep call sites honest long-term. Fine as a follow-up.

emostov: Originally I had something like this, but was just trying to minimize diff

client/src/lib.rs
emostov: It will always be empty. The user ID is a vestige of how wallets import/export was designed but is not at all related to secret storage other then the fact it re-uses some of the primitives

client/src/lib.rs
emostov: [3a131b9](https://github.com/tkhq/rust-sdk/pull/253/commits/3a131b9f75d47a9615797e7d3b26051dc3b2261f)

client/src/lib.rs
emostov: This would cause a bit bigger diff so punting for now

client/src/lib.rs
emostov: fixed

client/src/lib.rs
emostov: Doc comment should now reflect it is batch

richardpringle (dismissed):

richardpringle (approved):

PR #254, opened 2026-08-24
feat(tvc): add YubiKey registry, provisioning, and key management

richardpringle:
Adds the config model and device-management core for TVC operator keys on YubiKeys, plus `tvc keys provision-yubikey` and `tvc keys delete-yubikey`. YubiKey does not become an operator backend yet (TVC-307).

- Top-level `[[yubikeys]]` registry: canonical hex serial + cached composite public key (device stays authoritative; provisioning refreshes a stale cache). Org records reference a serial only.
- `DeviceOps` boundary in `tvc/src/yubikey.rs`: idempotent/resumable provisioning, deletion that only touches verified `CN=QuorumOS` certificates. yubikey 0.8 cannot delete slot private keys, so delete is certificate-only and says so; tvc never PIV-resets.
- The direct `yubikey` dep is temporary until qos_client re-exports the crate; CI installs `libpcsclite-dev` (linked at build time on Linux).

Closes TVC-306.


tvc/src/commands/keys/delete_yubikey.rs
```diff
@@ -0,0 +1,161 @@
...
+                &format!("Delete the QuorumOS key material on YubiKey {serial}?"),
+                "deletion",
+            )?;
+        }
+
+        let mut devices = PcscDevices;
+        let deleted = devices.delete_qos_material(serial)?;
+
+        config.deregister_yubikey(serial);
+        let config_path = config_file_path()?;
```
richardpringle: Something seems off here. If we can't fetch the file-path, then we don't even attempt to save? My guess is that this same function is called in `config.save`. It should probably be made private and an error with the filename should be returned, then that error can be augmented with the device material message.

tvc/src/commands/keys/provision_yubikey.rs
```diff
@@ -0,0 +1,150 @@
...
+        // QuorumOS touch policy demands physical touches during generation.
+        if ctx.is_non_interactive() || !prompts::stdin_can_prompt() {
+            bail!(
+                "provisioning a YubiKey is interactive: the PIN is prompted and \
+                 the device must be touched during key generation"
+            );
+        }
+
+        let mut devices = PcscDevices;
+        let connected = devices.connected_serials()?;
```
richardpringle: This is an odd pattern. This should probably just be a module instead of a ZST if it isn't passed anywhere

tvc/src/commands/keys/provision_yubikey.rs
```diff
@@ -0,0 +1,150 @@
...
+        )?);
+
+        shell_println!(ctx, "Touch the YubiKey each time it blinks.")?;
+
+        let provisioned = devices.ensure_provisioned(serial, &pin)?;
+        let registration = config.register_yubikey(serial, provisioned.pair_public_key);
+
+        if registration != Registration::Unchanged {
+            let config_path = config_file_path()?;
+            config.save().await.with_context(|| {
```
richardpringle: same comment

tvc/src/commands/keys/provision_yubikey.rs
```diff
@@ -0,0 +1,150 @@
...
+
+/// Renders `; connected: a, b` for a mismatch error, or nothing when no
+/// device is present (the bare message already says it all).
+fn connected_list(connected: &[YubiKeySerial]) -> String {
+    if connected.is_empty() {
+        return String::new();
+    }
+
+    let serials: Vec<String> = connected.iter().map(ToString::to_string).collect();
+    format!("; connected: {}", serials.join(", "))
```
richardpringle: Best to follow good patterns

```suggestion
    let mut s = String::new();
    
    if connected.is_empty() {
        return s;
    }
    
    let mut iter = connected.iter();
    
    if let Some(first) = iter.next() {
        write!(&mut s, "; connected: {}", first);
    }
    
    iter.for_each(|next| write!(&mut s, ", {}", next));

    s
```

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,914 @@
...
+            .iter()
+            .try_for_each(|slot| self.delete_qos_certificate(serial, *slot))?;
+
+        Ok(DeletedMaterial { cleared_slots })
+    }
+}
+
+/// The machine's connected YubiKeys, reached over PC/SC through `qos_client`
+/// and the `yubikey` crate.
+pub(crate) struct PcscDevices;
```
richardpringle: Ah, I see now. This is used for the trait implementation. I guess this works.

richardpringle: But when I look further, it seems like we're conflating too many thing into one big trait instead of having multiple smaller traits. 

The `connected_serials` method really represents an effect. The effect should probably come from `Ctx`. A mock `Ctx` or `Ctx` that's meant for testing can output different serials. 

`open`ing a serial should just be a free function. 

`slot_status` shouldn't even be a function, we should just implement the `DeviceOps` on the actual `Yubikey` type

tvc/src/commands/keys/provision_yubikey.rs
```diff
@@ -0,0 +1,223 @@
...
+            Registration::Added => Some(format!(
+                r#"add this to tvc.config.toml:
+
+[[yubikeys]]
+serial = "{serial}"
+public_key = "{operator_public_key}""#
+            )),
+            Registration::Updated => Some(format!(
+                r#"replace public_key on the existing [[yubikeys]] entry whose serial is {serial} with "{operator_public_key}""#
+            )),
```
richardpringle: why do we have to do this manually?

tvc/src/config/turnkey/mod.rs
```diff
@@ -99,10 +114,35 @@ mod disk {
...
+                            Some((alias, &operator.name, record.serial))
+                        }
+                        _ => None,
+                    })
+            });
+
+            if let Some((alias, name, serial)) = dangling {
+                bail!(
+                    "organization '{alias}' operator '{name}' references YubiKey {serial}, \
+                     which is not in the yubikeys registry"
```
richardpringle: So this whole cli is bricked then, with no advise to the user on how to fix?

tvc/src/config/turnkey/mod.rs
```diff
@@ -275,6 +316,14 @@ impl OperatorRecord {
...
     Local(LocalOperatorRecord),
     Hosted(HostedOperatorRecord),
+    // The rename keeps the product spelling on the wire; snake_case would
+    // split it into "yubi_key". Deliberately added within disk schema v1:
+    // older v1 readers reject a config that uses this variant, which is
+    // acceptable because nothing writes it until YubiKey becomes a
+    // selectable backend (TVC-307) — that work owns a version bump if the
+    // incompatibility turns out to matter.
+    #[serde(rename = "yubikey")]
+    YubiKey(YubiKeyOperatorRecord),
```
richardpringle:
```suggestion
    Yubikey(YubiKeyOperatorRecord),
```

richardpringle: I've already fix this once, why did you put it back?

tvc/src/config/turnkey/yubikey.rs
```diff
@@ -0,0 +1,302 @@
...
+
+impl TryFrom<Vec<YubiKeyRegistryEntry>> for YubiKeyRegistry {
+    type Error = YubiKeyRegistryParseError;
+
+    fn try_from(entries: Vec<YubiKeyRegistryEntry>) -> Result<Self, Self::Error> {
+        let duplicate = entries.iter().enumerate().find(|(index, entry)| {
+            entries[..*index]
+                .iter()
+                .any(|earlier| earlier.serial == entry.serial)
+        });
```
richardpringle: Just use a `HashSet`... what are you doing?

tvc/src/config/turnkey/yubikey.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+/// The registered YubiKey devices, at most one entry per serial.
+///
+/// Construction is the proof of uniqueness: deserialization funnels through
+/// `TryFrom`, and the mutating methods preserve the invariant.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[cfg_attr(test, derive(PartialEq))]
+#[serde(try_from = "Vec<YubiKeyRegistryEntry>")]
+pub struct YubiKeyRegistry(Vec<YubiKeyRegistryEntry>);
```
richardpringle: Why is this not a `IndexMap<Serial, RegistryEntry>`? Then validation on the conversion from the Vec just has to make sure there are no dupes (insert should return `None` every time)

tvc/src/config/turnkey/yubikey.rs
richardpringle: Don't even need to do this if you use an `IndexMap`

tvc/src/output.rs
```diff
@@ -205,9 +211,25 @@ impl<W: Write, W2: Write> Ctx<W, W2> {
         Self {
             shell,
             non_interactive,
+            connected_yubikeys: Box::new(crate::yubikey::connected_serials),
         }
     }
 
+    /// Replace the YubiKey discovery effect, for tests.
+    #[expect(dead_code, reason = "the test seam for scripted discovery")]
```
richardpringle: Is this a hack for integration tests? Why can we not use `cfg(test)`?

richardpringle (commented):
It's super weird that you're asking to manually enter the yubikey config

tvc/src/commands/keys/provision_yubikey.rs
```diff
@@ -0,0 +1,212 @@
...
+                    let mut serials = connected.iter();
+
+                    if let Some(first) = serials.next() {
+                        write!(list, "; connected: {first}")
+                            .expect("writing to a String cannot fail");
+                    }
+
+                    serials
+                        .try_for_each(|next| write!(list, ", {next}"))
+                        .expect("writing to a String cannot fail");
```
richardpringle: `std::fmt::Display::fmt` can fail, so this can definitely fail

tvc/src/commands/keys/delete_yubikey.rs
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        // Interactive confirmation, literal about what deletion can and
+        // cannot do. A non-interactive run without --yes was rejected up
+        // front, so reaching here with !self.yes means we can prompt.
+        if !self.yes {
+            shell_eprintln!(ctx, "")?;
+            shell_eprintln!(
+                ctx,
+                "WARNING: This deletes the QuorumOS certificates on YubiKey {serial} and"
```
emostov, line 102: nit: Whats the point of deleting cert but not key?

richardpringle: I didn't see any apis to delete an individual key, only a function to reset the entire device

Cargo.toml
```diff
@@ -17,11 +17,13 @@ members = [
...
 
 [workspace.dependencies]
-# QuorumOS
+qos_client = { version = "0.13", default-features = false, features = ["smartcard"] }
 qos_core = { version = "0.13", default-features = false }
 qos_crypto = { version = "0.13", default-features = false }
 qos_nsm = { version = "0.13", default-features = false }
 qos_p256 = { version = "0.13", default-features = false }
+# Matches qos_client's own pin; goes away once qos_client re-exports the crate.
+yubikey = { version = "0.8", features = ["untested"], default-features = false }
```
emostov: did the re-export not work?

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,544 @@
...
+//! open a device once and drive inspection and mutation through that one
+//! handle.
+//!
+//! Provisioning follows the QuorumOS conventions from `qos_client`: two
+//! P-256 keys in the PIV signing and key-agreement slots, each under a
+//! self-signed `CN=QuorumOS` certificate, PIN policy "always" and touch
+//! policy "always". That certificate subject is how TVC recognizes its own
+//! material; a slot holding anything else is never touched. A slot only
+//! counts as empty when its emptiness is proven — no readable certificate
+//! AND no key per the PIV metadata command (firmware 5.3+); anything
```
emostov, line 17: Not sure what "firmware 5.3+" means - is that yubikey firmware?

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,544 @@
...
+use yubikey::{
+    Error as PivError, MgmKey, Serial, TouchPolicy, YubiKey,
+    certificate::Certificate,
+    piv::{self, SlotId},
+    reader::Context,
+};
+use zeroize::Zeroizing;
+
+/// The subject of the self-signed certificates `qos_client` provisions.
+// TODO: qos_client keeps this constant private; have it exported (along with
```
emostov, line 40: track in linear

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,544 @@
...
+            key_agreement: self.slot_status(QosSlot::KeyAgreement)?,
+        })
+    }
+
+    fn provision_slot(&mut self, slot: QosSlot, pin: &Pin) -> Result<(), DeviceError> {
+        qos_client::yubikey::generate_signed_certificate(
+            self,
+            slot.slot_id(),
+            pin.as_bytes(),
+            MgmKey::default(),
```
emostov, line 315: noting we assume default managment key. But they can override it later

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,544 @@
...
+            SlotStatus::QosProvisioned => {}
+            SlotStatus::Foreign { subject } => {
+                return Err(DeviceError::ForeignSlot { slot, subject });
+            }
+            SlotStatus::Empty | SlotStatus::KeyWithoutCertificate => {
+                return Err(DeviceError::ChangedSinceInspection { slot });
+            }
+        }
+
+        self.authenticate(MgmKey::default())
```
emostov, line 341: So theoretically best practice they should override their management to something random after provisioning. So this wouldn't work if it requires management key

emostov: Does make me wonder the value of deleting just the certificate. Either reset the key or keep using it - not sure the product case of wanting to delete the certificate

emostov (dismissed):
I'm fine with merging as is - but noting that 

- for best practice we should advise to set a random management key
- not sure the practical purpose of deleting just the certificate
- we may want an option in the future for provisioning multiple yubikeys with the same key; but don't need to do that now so we can keep things scoped down

tvc/src/config/turnkey/mod.rs
```diff
@@ -99,10 +114,36 @@ mod disk {
...
-                Self::V0(config) => migrate_v0(config),
-                Self::V1(config) => Ok(config),
+            let config = match self {
+                Self::V0(config) => migrate_v0(config)?,
+                Self::V1(config) => config,
+            };
+
+            // An organization operator may only reference a registered
+            // device; rejecting a dangling serial here keeps everything
+            // downstream of the parse free of that state.
```
annieke, line 124: this makes sense to me, no using random yubikeys

tvc/src/config/turnkey/mod.rs
```diff
@@ -99,10 +114,36 @@ mod disk {
...
+                        }
+                        _ => None,
+                    })
+            });
+
+            if let Some((alias, name, serial)) = dangling {
+                bail!(
+                    "organization '{alias}' operator '{name}' references YubiKey {serial}, \
+                     which is not in the yubikeys registry; edit tvc.config.toml to add the \
+                     matching [[yubikeys]] entry or remove this operator record"
```
annieke, line 142: why instruct the users to edit the config directly instead of use the provision command? or is the provision command only for new keys, and will require a direct edit if it's an existing one?

Cargo.toml
richardpringle: My change didn't get included in the release. I'm not sure what happened

tvc/src/yubikey.rs
richardpringle: yes

tvc/src/config/turnkey/mod.rs
```diff
@@ -275,6 +317,7 @@ impl OperatorRecord {
 pub enum OperatorRecordKind {
     Local(LocalOperatorRecord),
     Hosted(HostedOperatorRecord),
+    Yubikey(YubiKeyOperatorRecord),
```
annieke, line 320: this makes sense to me to be its own kind

tvc/src/yubikey.rs
richardpringle: Good question. I guess there was no "delete key" command for older yubikeys, but I hadn't thought about just provisioning a new key in the old slot. Deleting the certificate tells us not to use the key. 

From a product perspective, it might make sense to say "if you want to do anything, you're going to have to use ykman" instead of offering this stuff through the tvc-cli. 

But I would make the case that we either offer both provisioning and deletion/reseting or neither. There should be parity

tvc/src/yubikey.rs
annieke, line 315: i didn't really review this file in depth, but while asking codex about yubikey best practice it called out this usage of the default management key and highlighted that 
> Yubico recommends changing the PIN, PUK, and management key from their universal defaults before operational use. [Yubico PIV credential guidance](https://docs.yubico.com/yesdk/users-manual/application-piv/pin-puk-mgmt-key.html).
That produces an awkward conflict:
> - Change the management key before provisioning, and this command cannot provision.
> - Change it afterward, and delete-yubikey cannot delete the certificates.
> - Leave it at the default, and anyone with physical/PCSC access can perform PIV administrative operations using a publicly known key.
>
> Before production use, I’d add support for a custom or PIN-protected management key—without storing its raw value in tvc.config.toml. Yubico supports generating a random management key and protecting it with the PIN. [ykman PIV access commands](https://docs.yubico.com/software/yubikey/tools/ykman/PIV_Commands.html).

maybe this is addressed in the followups so just putting this here

annieke (commented):
- new `kind = "yubikey"` makes sense to me
- so does using `serial` to identify the yubikey and also keeping track of `public_key` in the registry
- requiring a yubikey to be in registry before using ✅ 

the big picture of all the parts i know about feel right; future commands we can think about more deeply as we iterate. yubikey specifics i def am lacking on so just forwarded codex's thoughts.

tvc/src/config/turnkey/mod.rs
richardpringle: Provision is only for new at the moment. I want to get the happy path merged, then I can fix this

tvc/src/yubikey.rs
richardpringle: yeah, let's do in a follow up. We need a doc like this one:
https://github.com/tkhq/qos/blob/main/docs/provision_yubikey.md

that says what the user should do with `ykman`

.github/workflows/release-plz.yml
```diff
@@ -39,6 +39,8 @@ jobs:
         with:
           toolchain: stable
           components: clippy,rustfmt
+      - name: install PC/SC headers (linked by the yubikey crate)
+        run: sudo apt-get update && sudo apt-get install -y libpcsclite-dev
```
daniilrrr, line 43: I think this step also needs to be in `.github/workflows/tvc-binaries-release.yml` to get CI to pass

tvc/src/config/turnkey/yubikey.rs
```diff
@@ -0,0 +1,322 @@
...
+    str::FromStr,
+};
+
+/// A YubiKey device serial number.
+///
+/// String parsing accepts bare hexadecimal input with surrounding whitespace.
+/// Display always emits the canonical form: lowercase hexadecimal, zero-padded
+/// to eight digits.
+#[derive(
+    Clone, Copy, Debug, Default, PartialEq, Eq, Hash, SerializeDisplay, DeserializeFromStr,
```
daniilrrr, line 18: "we'll take the lot" 😆

tvc/src/outcome.rs
```diff
@@ -54,6 +54,10 @@ pub enum Outcome {
     LiveDeploymentSet(app::set_live_deploy::LiveDeploymentSet),
     AppDeleted(app::delete::AppDeleted),
     OperatorKeyBackedUp(keys::backup_operator_key::OperatorKeyBackedUp),
+    // Spelled Yubikey, not YubiKey: the variant name IS the reason (see the
+    // module docs), and the wire strings are yubikey_provisioned/_deleted.
+    YubikeyProvisioned(keys::provision_yubikey::YubiKeyProvisioned),
+    YubikeyDeleted(keys::delete_yubikey::YubiKeyDeleted),
```
daniilrrr, line 60: I would prefer consistency in the code for this (use uppercase K everywhere) and explicitly rename like 

```suggestion
    #[serde(rename = "yubikey_provisioned")]
    YubikeyProvisioned(keys::provision_yubikey::YubiKeyProvisioned),
     #[serde(rename = "yubikey_deleted")]
    YubikeyDeleted(keys::delete_yubikey::YubiKeyDeleted),
```

tvc/src/config/turnkey/yubikey.rs
```diff
@@ -0,0 +1,322 @@
...
+pub enum Registration {
+    /// The serial was newly registered.
+    Added,
+    /// The serial was registered with a stale public key; the cache now
+    /// matches the device.
+    Updated,
+    /// The registry already matched the device.
+    #[default]
+    Unchanged,
+}
```
daniilrrr, line 161: naming: this was hard for me to grok. more explicit names would help at a glance imo

```suggestion
pub enum RegistryEntryChange {
    /// The serial was newly registered.
    Added,
    /// The serial was registered with a stale public key; the cache now
    /// matches the device.
    PublicKeyUpdated,
    /// The registry already matched the device.
    #[default]
    Unchanged,
}
```

then you have a bit nicer usage like 

```rust
if registry_change == YubiKeyRegistryChange::PublicKeyUpdated {
      config.save().await?;
  }
```

tvc/src/config/turnkey/mod.rs
```diff
@@ -99,10 +114,36 @@ mod disk {
...
-                Self::V1(config) => Ok(config),
+            let config = match self {
+                Self::V0(config) => migrate_v0(config)?,
+                Self::V1(config) => config,
+            };
+
+            // An organization operator may only reference a registered
+            // device; rejecting a dangling serial here keeps everything
+            // downstream of the parse free of that state.
+            let dangling = config.orgs.iter().find_map(|(alias, org)| {
```
daniilrrr, line 125: this is specific to Yubikeys right? I think lines 125-144 should be their own func 


```suggestion
            let unregistered_yubikeys = config.orgs.iter().find_map(|(alias, org)| {
```

tvc/src/config/turnkey/mod.rs
```diff
@@ -433,9 +476,15 @@ impl Config {
...
 
+    pub(crate) fn to_toml(&self) -> Result<String> {
+        disk::to_toml(self)
+    }
+
     /// Save config to disk
     pub async fn save(&self) -> Result<()> {
-        let path = config_file_path()?;
+        // Owns its whole failure surface: callers add what only they know
+        // (e.g. recovery guidance), never the file identity.
```
daniilrrr, line 486: nit: not sure this is a useful comment, could remove

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,547 @@
+//! YubiKey device management for TVC operator keys.
+//!
+//! Concrete hardware access has three shapes: discovering connected devices
+//! is an effect supplied by [`crate::output::Ctx`] (backed by
+//! [`connected_serials`]), opening a serial is the ambient [`open`] entry
+//! point, and per-device functionality hangs off the opened
+//! [`yubikey::YubiKey`] through the [`DeviceOps`] extension trait. Commands
```
daniilrrr, line 7: unnecessary prefix in my IDE


```suggestion
//! [`YubiKey`] through the [`DeviceOps`] extension trait. Commands
```

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,547 @@
...
+        "the {slot} slot holds a non-QuorumOS certificate ({subject}); refusing to touch this device"
+    )]
+    ForeignSlot { slot: QosSlot, subject: String },
+    #[error(
+        "the {slot} slot no longer holds a QuorumOS certificate; the device changed since it was inspected"
+    )]
+    ChangedSinceInspection { slot: QosSlot },
+    // qos_client's YubiKeyError implements neither Display nor Error, so it
+    // is rendered by value instead of chained as a source.
+    // TODO: derive those in qos_client and turn these fields into #[source].
```
daniilrrr, line 222: need ticket

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,547 @@
...
+    #[error(
+        "the {slot} slot holds a non-QuorumOS certificate ({subject}); refusing to touch this device"
+    )]
+    ForeignSlot { slot: QosSlot, subject: String },
+    #[error(
+        "the {slot} slot no longer holds a QuorumOS certificate; the device changed since it was inspected"
+    )]
+    ChangedSinceInspection { slot: QosSlot },
+    // qos_client's YubiKeyError implements neither Display nor Error, so it
+    // is rendered by value instead of chained as a source.
```
daniilrrr, line 221:
```suggestion
    /// qos_client's [`YubiKeyError`] implements neither Display nor Error, so it
    /// is rendered by value instead of chained as a source.
```

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,547 @@
...
+        slot: QosSlot,
+        #[source]
+        source: PivError,
+    },
+}
+
+/// Decision table mapping raw certificate and key-metadata outcomes to a
+/// slot classification.
+///
+/// `Certificate::read` reports an absent, unreadable, and malformed
```
daniilrrr, line 242:
```suggestion
/// [`Certificate::read`] reports an absent, unreadable, and malformed
```

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,547 @@
...
+/// certificate identically, so those outcomes alone never prove emptiness;
+/// the (lazily consulted) key metadata must also show no key before a slot
+/// counts as [`SlotStatus::Empty`]. Pure so the ambiguous cases stay
+/// regression-tested without hardware.
+fn classify_slot(
+    slot: QosSlot,
+    certificate: Result<String, PivError>,
+    key_presence: impl FnOnce() -> Result<bool, PivError>,
+) -> Result<SlotStatus, DeviceError> {
```
daniilrrr, line 251: i find this mix of responsibilities and code paths hard to read and reason about. had LLM spin up an alternative with small pure classifiers and a flatter logic structure that preserves lazy metadata access

LLM: 
> This separates three responsibilities cleanly:
>   1. classify_certificate interprets the certificate read.
>   2. slot_status decides whether metadata I/O is necessary.
>   3. classify_key_presence interprets the metadata result.

```rust
  /// Result of inspecting a slot's certificate.
  ///
  /// `None` means no readable certificate was found, so key metadata must be
  /// consulted before the slot can be classified.
  fn classify_certificate(
      slot: QosSlot,
      certificate: Result<String, PivError>,
  ) -> Result<Option<SlotStatus>, DeviceError> {
      match certificate {
          Ok(subject) if subject == QOS_CERTIFICATE_SUBJECT => {
              Ok(Some(SlotStatus::QosProvisioned))
          }
          Ok(subject) => Ok(Some(SlotStatus::Foreign { subject })),
          Err(PivError::InvalidObject | PivError::NotFound) => Ok(None),
          Err(source) => Err(DeviceError::ReadCertificate { slot, source }),
      }
  }

  /// Classify a slot that has no readable certificate using its key metadata.
  fn classify_key_presence(
      slot: QosSlot,
      key_presence: Result<bool, PivError>,
  ) -> Result<SlotStatus, DeviceError> {
      match key_presence {
          Ok(false) => Ok(SlotStatus::Empty),
          Ok(true) => Ok(SlotStatus::KeyWithoutCertificate),
          Err(source) => Err(DeviceError::ReadSlotMetadata { slot, source }),
      }
  }
```

so `slot_status()` becomes more explicit

```rust
  impl DeviceOps for YubiKey {
      fn slot_status(&mut self, slot: QosSlot) -> Result<SlotStatus, DeviceError> {
          let certificate =
              Certificate::read(self, slot.slot_id()).map(|certificate| certificate.subject());

          if let Some(status) = classify_certificate(slot, certificate)? {
              return Ok(status);
          }

          // A missing or unreadable certificate does not prove that the slot is
          // empty. Consult the PIV metadata to determine whether a key remains.
          let key_presence = match piv::metadata(self, slot.slot_id()) {
              Ok(_) => Ok(true),
              Err(PivError::NotFound) => Ok(false),
              Err(source) => Err(source),
          };

          classify_key_presence(slot, key_presence)
      }

      fn device_status(&mut self) -> Result<DeviceStatus, DeviceError> {
          Ok(DeviceStatus {
              signing: self.slot_status(QosSlot::Signing)?,
              key_agreement: self.slot_status(QosSlot::KeyAgreement)?,
          })
      }

      fn provision_slot(&mut self, slot: QosSlot, pin: &Pin) -> Result<(), DeviceError> {
          qos_client::yubikey::generate_signed_certificate(
              self,
              slot.slot_id(),
              pin.as_bytes(),
              MgmKey::default(),
              TouchPolicy::Always,
          )
          .map(drop)
          .map_err(|error| DeviceError::Provision { slot, error })
      }

      fn pair_public_key(&mut self) -> Result<QosOperatorPublicKey, DeviceError> {
          let bytes = qos_client::yubikey::pair_public_key(self)
              .map_err(|error| DeviceError::ReadPairPublicKey { error })?;

          QosOperatorPublicKey::try_from(bytes.as_slice())
              .map_err(DeviceError::MalformedPairPublicKey)
      }

      fn delete_qos_certificate(&mut self, slot: QosSlot) -> Result<(), DeviceError> {
          match self.slot_status(slot)? {
              SlotStatus::QosProvisioned => {}
              SlotStatus::Foreign { subject } => {
                  return Err(DeviceError::ForeignSlot { slot, subject });
              }
              SlotStatus::Empty | SlotStatus::KeyWithoutCertificate => {
                  return Err(DeviceError::ChangedSinceInspection { slot });
              }
          }

          self.authenticate(MgmKey::default())
              .map_err(DeviceError::Authenticate)?;

          Certificate::delete(self, slot.slot_id())
              .map_err(|source| DeviceError::DeleteCertificate { slot, source })
      }
  }
```

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,547 @@
...
+            key_agreement: self.slot_status(QosSlot::KeyAgreement)?,
+        })
+    }
+
+    fn provision_slot(&mut self, slot: QosSlot, pin: &Pin) -> Result<(), DeviceError> {
+        qos_client::yubikey::generate_signed_certificate(
+            self,
+            slot.slot_id(),
+            pin.as_bytes(),
+            MgmKey::default(),
```
daniilrrr, line 315: annoying ask - can we add a loud TODO here and anywhere else we assume `MgmKey::default(),` so that we dont forget to address this?

tvc/src/commands/keys/delete_yubikey.rs
daniilrrr, line 1: I think this and everything related should be renamed to `deregister_yubikey.rs` to more clearly say that we don't actually delete the key, we just deregister the material. To do true deletion you would need to go to `ykman` and I think it's valuable to make that extra clear and avoid internal confusion too

tvc/src/commands/keys/provision_yubikey.rs
```diff
@@ -0,0 +1,209 @@
...
+    #[arg(long, value_name = "SERIAL")]
+    serial: Option<YubiKeySerial>,
+}
+
+impl Run for Args {
+    type Outcome = YubiKeyProvisioned;
+
+    async fn run(self, ctx: &mut StdCtx, mut config: Config) -> Result<YubiKeyProvisioned> {
+        // No non-interactive fallback exists: the PIN is prompted and the
+        // QuorumOS touch policy demands physical touches during generation.
```
daniilrrr, line 33: redundant comment to `bail!` statement, remove

tvc/src/commands/keys/provision_yubikey.rs
```diff
@@ -0,0 +1,209 @@
...
+                    if let Some(first) = serials.next() {
+                        write!(list, "; connected: {first}")?;
+                    }
+
+                    serials.try_for_each(|next| write!(list, ", {next}"))?;
+
+                    list
+                };
+
+                bail!("YubiKey {serial} is not connected{connected_list}");
```
daniilrrr, line 61: typo
```suggestion
                bail!("YubiKey {serial} is not connected {connected_list}");
```

tvc/src/yubikey.rs
```diff
@@ -0,0 +1,547 @@
...
+
+    /// Full provision-inspect-delete cycle against real hardware.
+    ///
+    /// DESTRUCTIVE: generates keys in, and deletes the certificates from,
+    /// the PIV signing and key-agreement slots of the sole connected
+    /// YubiKey. Requires the factory-default PIN and management key, and
+    /// firmware 5.3+ (the metadata command proves slot emptiness). Run
+    /// manually: `cargo test -p tvc --lib -- --ignored hardware_`
+    #[test]
+    #[ignore = "requires a connected YubiKey with default PIN/management key; overwrites its QuorumOS slots"]
```
daniilrrr, line 505: this is sick lol. how do we run it in CI? or is the idea that this is a util?

daniilrrr (commented):
lgtm, various nits 

most urgent request is to clear up this [logic](https://github.com/tkhq/rust-sdk/pull/254#discussion_r3867809854)

tvc/src/yubikey.rs
richardpringle: I don't have an answer or this. I think the best thing we can do for now is use mocks/fakes and manually test with real hardware

emostov:
@richardpringle 

Maybe a safer thing here is just take the management key as an arg and if its not present try the default one. Management key allows you to reset pin so we really want to avoid foot gun of users thinking default managmenet key in production is ok

emostov (approved):

PR #257, opened 2026-08-25
feat(tvc): add YubiKey as a default operator backend

richardpringle:
Stacked on #254.

Adds `yubikey` as a default operator kind. Org records reference a registered device by serial only; resolution opens that exact serial, verifies the QuorumOS slots, and returns a `qos_client`-backed `Pair` — `sign` for manifest approval, `decrypt` for local-style share provisioning. Blocking PC/SC work runs off the async threads, one device at a time; the PIN goes through the existing secure prompt and is never stored.

Also adds `keys refresh-yubikey` to resync the registry's cached public key from the device.

Unit tests run against the fake device boundary and a fake `Pair`; real-hardware coverage stays in an ignored, opt-in suite.

Closes TVC-307.

### DIFF
--- a/tvc/src/config/turnkey/mod.rs
+++ b/tvc/src/config/turnkey/mod.rs
@@ -276,6 +276,7 @@ pub enum OperatorKind {
     #[default]
     Local,
     Hosted,
+    Yubikey,
 }
 
 impl Display for OperatorKind {
@@ -283,6 +284,7 @@ impl Display for OperatorKind {
         match self {
             Self::Local => "local".fmt(f),
             Self::Hosted => "hosted".fmt(f),
+            Self::Yubikey => "yubikey".fmt(f),
         }
     }
 }
@@ -734,6 +736,7 @@ future_entry = "keep"
 [orgs.default]
 id = "org-123"
 api_key_path = "/keys/api.json"
+default_operator_kind = "yubikey"
 
 [[orgs.default.operators]]
 name = "yubikey"
@@ -753,6 +756,7 @@ serial = "1C95C1F"
         assert_eq!(entry.public_key.to_string(), "07".repeat(130));
 
         let org = &config.orgs["default"];
+        assert_eq!(org.default_operator_kind, OperatorKind::Yubikey);
         let OperatorRecordKind::Yubikey(record) = &org.operators[0].kind else {
             panic!("operator must be a yubikey record");
         };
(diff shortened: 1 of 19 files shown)


tvc/src/config/turnkey/mod.rs
```diff
@@ -434,6 +446,28 @@ impl OrgConfig {
...
+            .operators
+            .iter()
+            .filter_map(|operator| match &operator.kind {
+                OperatorRecordKind::Yubikey(yubikey) => Some((operator, yubikey)),
+                OperatorRecordKind::Local(_) | OperatorRecordKind::Hosted(_) => None,
+            });
+
+        match (yubikeys.next(), yubikeys.next()) {
+            (Some(sole), None) => Ok(sole),
+            (None, _) => Err(SelectYubiKeyOperatorError::NoYubiKeyOperator),
```
richardpringle: Why is this invalid?

tvc/tests/common.rs
```diff
@@ -115,6 +116,63 @@ pub fn write_hosted_only_config(home: &Path, alias: &str, org_id: &str) {
     .unwrap();
 }
 
+/// Write a v1 `tvc.config.toml` under `home` whose sole, active organization
+/// defaults to the yubikey backend: one serial-only operator record plus the
+/// top-level registry entry caching a real generated composite key. Flows
+/// that read the cache need no device; device-touching flows fail at their
+/// first PC/SC step.
+pub fn write_yubikey_only_config(home: &Path, alias: &str, org_id: &str) {
```
annieke, line 124: gud case

tvc/src/yubikey.rs
```diff
@@ -324,6 +369,32 @@ impl DeviceOps for YubiKey {
             .map_err(DeviceError::MalformedPairPublicKey)
     }
 
+    fn sign(&mut self, pin: &Pin, message: &[u8]) -> Result<Vec<u8>, DeviceError> {
```
annieke, line 375: really bought in now to the `Pair` abstraction with `sign` with this change

tvc/src/yubikey/pair.rs
```diff
@@ -0,0 +1,444 @@
...
+}
+
+/// A [`PinAcquisition`] with the [`PinAcquisition::Unavailable`] case
+/// refused: holding one is proof that device I/O can start because a PIN
+/// can definitely follow.
+pub(crate) enum AvailablePin {
+    Prompt,
+    #[cfg(test)]
+    Fixed(Pin),
+}
```
annieke, line 44: i don't have the best idea of how pins should work but these 2 enums smell slightly overlapping

annieke: jk i actually take this back seeing how `AvailablePin` is used in `PairSource`

annieke (commented):
very cursory review and honestly look pretty good to me and extending `Pair` works great!

tvc/src/yubikey.rs
richardpringle: we can still alter it!

emostov (approved):
LGTM

PR #260, opened 2026-08-26
Add YubiKey operators to interactive organization and operator flows

richardpringle:
## Summary

YubiKey is now a first-class TVC operator backend while device administration remains outside TVC:

- `tvc login` and `tvc operator create --kind yubikey` can attach an already registered YubiKey to an organization. Several serials may coexist, with interactive selection or explicit `--serial` disambiguation.
- `tvc deploy approve` signs with the slot `9c` key, and `tvc keys re-encrypt-local-share` performs key agreement with slot `9d`.
- Operator selection matches configured public keys against the manifest set, including deriving the server-side operator identity from a fetched deployment.
- `tvc keys refresh-yubikey` reads installed certificates and caches the composite operator public key in local config.

## Device setup and lifecycle

- `tvc yubikey create-certs` reads the public keys and metadata from `9c` and `9d`, enforces P-256/generated/PIN Always/touch Always, and creates verified self-signed certificates without modifying the device or using a management key.
- Key generation and certificate import are explicit `ykman` operations. TVC never generates, imports, or deletes device keys or certificates and never authenticates with the PIV management key.
- `tvc yubikey unregister` removes only the local registry entry after checking that no local organization profile references it. It warns that the device and remote operator authorization remain valid.
- The old device-mutating provision and delete commands are removed.

## Documentation

[Rendered YubiKey setup and usage guide](https://github.com/tkhq/rust-sdk/blob/richard/tvc-308-add-yubikey-operators-to-interactive-organization-and/tvc/YUBIKEY.md)

Closes TVC-308.


tvc/src/commands/login.rs
```diff
@@ -283,14 +347,56 @@ fn build_login_plan_interactive(
...
+                                write!(f, "{} (serial {})", self.0.name, self.1.serial)
+                            }
+                        }
+
+                        let choices = org
+                            .yubikey_operators()
+                            .map(|(operator, yubikey)| Choice(operator, yubikey))
+                            .collect();
+                        let Choice(_, yubikey) =
+                            prompts::select("Select YubiKey operator", choices)?;
```
emostov, line 384: do we need to check if we are in interactive mode?

emostov (approved):

PR #263, opened 2026-08-28
Force use of modern file layout

richardpringle:
Call me old fashioned, but I actually didn't use AI on this one!

The real question is why? 
 - consistency (we want one way to define new file-based modules)
 - this file layout actually encourages splitting into submodules without blowing up the diff and focing files to move. You simple go from one `my_mod.rs` to `my_mod.rs` + `my_mod/my_sub_mod.rs` to add a submodule without having to move any files.


daniilrrr (approved):
chill, thx for adding lint. says commit is unverified so try a force push

PR #265, opened 2026-08-31
Re-enable tvc musl builds

richardpringle:
To ship yubikey support as quickly as possible, we disabbled the musl builds. This re-enables them. Using `Meson` is the official way to build `pcsc-lite`, see the repository workflow: https://github.com/LudovicRousseau/PCSC/blob/aa97f4b65448ede1a073db51195802cb00fc808b/.github/workflows/build_meson.yml

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,10 @@
 # Cargo.lock changes should trigger dependency reviewers
 Cargo.lock @tkhq/dependency-reviewers
 
+# TVC's musl build downloads and pins dependencies outside Cargo.lock
+/.github/cross/install-pcsc-lite.sh @tkhq/dependency-reviewers
+/.github/workflows/tvc-pcsc-release-check.yml @tkhq/dependency-reviewers
+
 # We also want this file (COWDEOWNERS file) to trigger dependency reviewers
 # Otherwise anyone can remove dependency reviews!
 /.github/CODEOWNERS @tkhq/dependency-reviewers
--- a/.github/cross/install-pcsc-lite.sh
+++ b/.github/cross/install-pcsc-lite.sh
@@ -87,6 +87,9 @@ tar -xJf "${archive}" -C /tmp
   echo "cpu = '${cpu}'"
   echo "endian = 'little'"
   echo
+  echo '[built-in options]'
+  echo "c_args = ['-D_FORTIFY_SOURCE=2', '-fPIC', '-fstack-protector-strong', '-Wformat', '-Wformat-security', '-Werror=format-security']"
+  echo
   echo '[properties]'
   echo 'needs_exe_wrapper = true'
 } > "${cross_file}"
--- a/.github/cross/install-pcsc-lite.sh
+++ b/.github/cross/install-pcsc-lite.sh
@@ -5,6 +5,8 @@ set -euo pipefail
 target="${1:?missing target triple}"
 version="2.5.1"
 sha512="027851359b38cf56c2ea97a969c0ae8c5eabbb977fc2a86c650c6c8e0e2caba43934f3de35573deb408bf6365e9deb2a9c443b3712b313692069421c77222057"
+meson_version="0.58.2"
+meson_sha256="7634ec32955d3f897d623b88e9d2988451035f43d73c17a29caf767387baedb7"
 
 case "${target}" in
   aarch64-unknown-linux-musl)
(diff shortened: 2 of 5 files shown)


Cross.toml
```diff
@@ -0,0 +1,19 @@
+[target.aarch64-unknown-linux-musl]
+pre-build = ".github/cross/install-pcsc-lite.sh"
+
+[target.aarch64-unknown-linux-musl.env]
+passthrough = [
+    "LIBPCSCLITE_STATIC=1",
```
daniilrrr, line 8: what is the purpose of this flag?

.github/cross/install-pcsc-lite.sh
```diff
@@ -0,0 +1,98 @@
...
+  ca-certificates \
+  curl \
+  flex \
+  ninja-build \
+  perl \
+  pkg-config \
+  python3-pip \
+  python3-setuptools \
+  xz-utils
+python3 -m pip install --no-cache-dir 'meson==0.58.2'
```
daniilrrr: for extra security and maintability it would be nice to list out what each dep does and why this needs them in comments

.github/cross/install-pcsc-lite.sh
```diff
@@ -0,0 +1,98 @@
...
+build_dir="/tmp/pcsc-lite-build"
+cross_file="/tmp/pcsc-lite-${target}.ini"
+prefix="/opt/pcsc/${target}"
+
+curl \
+  --fail \
+  --location \
+  --retry 5 \
+  --output "${archive}" \
+  "https://pcsclite.apdu.fr/files/pcsc-lite-${version}.tar.xz"
```
daniilrrr, line 101: a pretty random URL here, can you add a comment justifying the origin?

.github/cross/install-pcsc-lite.sh
```diff
@@ -0,0 +1,98 @@
...
+  --cross-file "${cross_file}" \
+  --libdir=lib \
+  --prefix="${prefix}" \
+  -Ddefault_library=static \
+  -Dlibsystemd=false \
+  -Dlibudev=false \
+  -Dlibusb=false \
+  -Dpolkit=false \
+  -Dserial=false \
+  -Dusb=false
```
daniilrrr, line 151: add comments justifying each flag

.github/workflows/tvc-binaries-release.yml
```diff
@@ -33,14 +33,14 @@ jobs:
...
-          #   build_tool: cross
-          #   rustflags: -C target-feature=+crt-static
-          # - target: aarch64-unknown-linux-musl
-          #   os: ubuntu-latest
-          #   build_tool: cross
-          #   rustflags: -C target-feature=+crt-static
+          - target: x86_64-unknown-linux-musl
+            os: ubuntu-latest
+            build_tool: cross
+            rustflags: -C target-feature=+crt-static
```
daniilrrr, line 39: what does `crt-static` do here?

daniilrrr (commented):
since this is a release script, max explicitness and justification for what each line does would be valuable for security and maintenance

.github/workflows/tvc-binaries-release.yml
richardpringle: The Rust compiler flag `-C target-feature=+crt-static` instructs the compiler to statically link the C runtime (CRT) into your final executable.

.github/cross/install-pcsc-lite.sh
richardpringle: It's the website for the macos and linux implementations of pcsc

.github/cross/install-pcsc-lite.sh
richardpringle: See comments above

Cross.toml
richardpringle: It's a build time env-var for `pkg-config` that tells it to statically link the c-library (default is to dynamically link it).

See https://docs.rs/pkg-config/latest/pkg_config/ for details

daniilrrr (commented):
this looks great, thanks for adding all the context in the files

you say `requires a supply-chain review` in a few places. Would it be worth adding these files to CODEOWNERS similar to how `Cargo.lock` changes trigger review? 
[link](https://github.com/tkhq/rust-sdk/blob/31384093bbb385362c82cfe640e7b8fdba12e08b/.github/CODEOWNERS#L2)

will approve once you decide

.github/CODEOWNERS
```diff
@@ -1,6 +1,10 @@
 # Cargo.lock changes should trigger dependency reviewers
 Cargo.lock @tkhq/dependency-reviewers
 
+# TVC's musl build downloads and pins dependencies outside Cargo.lock
```
daniilrrr, line 4: it's not exactly a dependency however `tvc-binaries-release.yaml` is now a sensitive file with its `Verify musl` binary hardening step, so maybe it should be included here to warrant extra review before edits? and we don't have a more applicable reviewer group afaict

daniilrrr (approved):

emostov (approved):

PR #267, opened 2026-08-31
Tighten YubiKey provisioning validation

richardpringle:
## Summary

- Validate device-generated P-256 metadata, PIN/touch policies, certificate key matching, and self-signatures before treating a YubiKey slot as provisioned.
- Preserve reader and metadata error sources while keeping device selection structured until the command boundary.
- Flatten slot-state handling to remove single-use helpers, an unnecessary clone, and a temporary allocation.

## Verification

- cargo test -p tvc --lib
- cargo clippy -p tvc --all-targets -- -D warnings
- cargo fmt --all -- --check

### DIFF
--- a/tvc/src/commands/yubikey/create_certs.rs
+++ b/tvc/src/commands/yubikey/create_certs.rs
@@ -6,7 +6,7 @@ use crate::{
     outcome::Outcome,
     output::StdCtx,
     prompts,
-    yubikey::{self, CertificateDeviceOps, ConnectedYubiKeys, Pin, QosSlot, YubiKeySelectionError},
+    yubikey::{self, CertificateDeviceOps, Pin, QosSlot, YubiKeySelectionError},
 };
 use anyhow::{Context, Result, bail};
 use clap::Args as ClapArgs;
@@ -37,7 +37,7 @@ impl Args {
             );
         }
 
-        let serial = match ConnectedYubiKeys::from(ctx.connected_yubikeys()?).choose(self.serial) {
+        let serial = match ctx.connected_yubikeys()?.choose(self.serial) {
             Ok(serial) => serial,
             Err(YubiKeySelectionError::NoneConnected) => bail!("no YubiKey is connected"),
             Err(YubiKeySelectionError::NotConnected {
--- a/tvc/src/commands/yubikey/create_certs.rs
+++ b/tvc/src/commands/yubikey/create_certs.rs
@@ -44,12 +44,6 @@ impl Args {
                 requested,
                 connected,
             }) => {
-                let connected = connected
-                    .into_iter()
-                    .map(|serial| serial.to_string())
-                    .collect::<Vec<_>>()
-                    .join(", ");
-
                 if connected.is_empty() {
                     bail!("YubiKey {requested} is not connected");
                 }
@@ -57,14 +51,8 @@ impl Args {
                 bail!("YubiKey {requested} is not connected; connected: {connected}")
             }
             Err(YubiKeySelectionError::Ambiguous { connected }) => {
-                let serials = connected
-                    .into_iter()
-                    .map(|serial| serial.to_string())
-                    .collect::<Vec<_>>()
-                    .join(", ");
-
                 bail!(
-                    "multiple YubiKeys are connected (serials {serials}); unplug all but the one \
+                    "multiple YubiKeys are connected (serials {connected}); unplug all but the one \
                      to use and try again, or pass --serial"
                 )
             }
--- a/tvc/src/commands/yubikey/create_certs.rs
+++ b/tvc/src/commands/yubikey/create_certs.rs
@@ -6,7 +6,7 @@ use crate::{
     outcome::Outcome,
     output::StdCtx,
     prompts,
-    yubikey::{self, CertificateDeviceOps, ConnectedYubiKeys, Pin, QosSlot},
+    yubikey::{self, CertificateDeviceOps, ConnectedYubiKeys, Pin, QosSlot, YubiKeySelectionError},
 };
 use anyhow::{Context, Result, bail};
 use clap::Args as ClapArgs;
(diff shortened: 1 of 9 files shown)


tvc/src/yubikey.rs
```diff
@@ -55,17 +55,31 @@ const QOS_CERTIFICATE_SUBJECT: &str = "CN=QuorumOS";
...
+            // a YubiKey with PIV disabled; both report a missing PIV applet.
+            Err(PivError::AppletNotFound { applet_name: "PIV" }) => {}
+            Err(source) => {
+                return Err(DeviceError::OpenReader {
+                    reader: reader.name().into_owned(),
+                    source,
+                });
+            }
+        }
+    }
```
richardpringle: You can still accomplish the same logic with an iterator and `try_collect`

tvc/src/commands/yubikey/create_certs.rs
```diff
@@ -37,7 +37,39 @@ impl Args {
...
+                    .map(|serial| serial.to_string())
+                    .collect::<Vec<_>>()
+                    .join(", ");
+
+                bail!(
+                    "multiple YubiKeys are connected (serials {serials}); unplug all but the one \
+                     to use and try again, or pass --serial"
+                )
+            }
+        };
```
richardpringle, line 59: `unplug all but one...` is contextual, but serializing all the connected serials into a string should be part of a custom `Display` implementation

tvc/src/yubikey.rs
```diff
@@ -568,32 +539,114 @@ pub(crate) trait DeviceOps {
...
+                    metadata_error,
+                })
+            }
+            (SlotStatus::Empty, _) => Err(DeviceError::EmptySlot {
+                slot: QosSlot::Signing,
+            }),
+            (_, SlotStatus::Empty) => Err(DeviceError::EmptySlot {
+                slot: QosSlot::KeyAgreement,
+            }),
+            (SlotStatus::QosProvisioned, SlotStatus::QosProvisioned) => self.pair_public_key(),
```
richardpringle: put this at the top

tvc/src/yubikey.rs
```diff
@@ -568,32 +539,114 @@ pub(crate) trait DeviceOps {
...
+            key_agreement,
+        } = self.device_status()?;
 
-        if let Some(slot) = status.slots_with(SlotStatus::Empty).into_iter().next() {
-            return Err(DeviceError::EmptySlot { slot });
+        match (signing, key_agreement) {
+            (SlotStatus::Foreign { subject }, _) => Err(DeviceError::ForeignSlot {
+                slot: QosSlot::Signing,
+                subject,
+            }),
```
richardpringle: hiding a potential slot error for the key-agreement will just lead to extra work for the user only to find out there was another error that they could have found out about sooner. Make an outer error enum that looks something like this:
```rust
enum SlotsError {
    Signing(DeviceError),
    KeyAgreement(DeviceError),
    SigningAndKeyAgreement {
        signing: DeviceError,
        key_agreement: DeviceError,
    }
}
```

tvc/src/yubikey.rs
```diff
@@ -55,20 +56,27 @@ const QOS_CERTIFICATE_SUBJECT: &str = "CN=QuorumOS";
...
 /// [`crate::output::Ctx`].
 pub(crate) fn connected_serials() -> Result<Vec<YubiKeySerial>, DeviceError> {
     let mut context = Context::open().map_err(DeviceError::Discovery)?;
     let readers = context.iter().map_err(DeviceError::Discovery)?;
 
-    Ok(readers
-        .filter_map(|reader| reader.open().ok())
-        .map(|yubikey| YubiKeySerial::from(yubikey.serial().0))
-        .collect())
+    Itertools::try_collect(readers.filter_map(|reader| match reader.open() {
```
richardpringle: I don't think this justifies adding a dependency (even if it's already in the workspace. I led you astray suggesting to use `try_collect`. What I actually meant for you to use was `collect::<Result<_, _>>()

tvc/src/yubikey.rs
```diff
@@ -81,37 +89,48 @@ impl ConnectedYubiKeys {
...
     }
+
+    pub(crate) fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+}
+
+impl Display for ConnectedYubiKeys {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0.iter().format(", "))
```
richardpringle: Don't need to use itertools here, can just do this the way you did it before

tvc/src/yubikey.rs
```diff
@@ -566,41 +565,104 @@ pub(crate) trait DeviceOps {
...
+            }
+            Ok(SlotStatus::UnknownWithoutCertificate { metadata_error }) => {
+                Err(DeviceError::UnknownWithoutCertificate {
+                    slot,
+                    metadata_error,
+                })
+            }
+            Ok(SlotStatus::Foreign { subject }) => Err(DeviceError::ForeignSlot { slot, subject }),
+            Err(error) => Err(error),
+        };
```
richardpringle: A closure that doesn't capture anything from its environment and is called within the same scope that it's defined is an anti-pattern. Make this a free function (private to this module)

tvc/src/yubikey/pair.rs
```diff
@@ -42,10 +42,14 @@ impl<D> Debug for YubiKeyPair<D> {
 impl<D: DeviceOps + Send + 'static> YubiKeyPair<D> {
     /// Run one blocking device operation on the blocking thread pool; the
     /// mutex serializes access to the device.
-    async fn device_op<T: Send + 'static>(
+    async fn device_op<T, E>(
```
richardpringle, line 45: Why did this change?

emostov (approved):

PR #269, opened 2026-09-02
ci: skill drift check — Valet-powered semantic ruling on command/skill parity

yourbuddyconner:
## What

Adds `.github/workflows/skill-drift-check.yml` — a new CI workflow that detects when CLI command source changes without a matching skill documentation update.

## Why

The existing `embedded_manifest_matches_skills_dir` cargo test enforces **file-list** parity between skills on disk and the embedded manifest. But it does not catch **semantic** drift — when a command's flags, output fields, error codes, or outcome reasons change and the skill prose is left stale. Per `tvc/AGENTS.md`:

> When changing the CLI surface or behavior — commands, flags, outcome `reason`s, error `code`s, output fields — update `skills/tvc-deployments/` in the same change.

This workflow automates that check.

## How it works

1. **Triggers** on PRs to `main` that touch `tvc/src/commands/**` or `tvc/skills/**`.
2. **Detects changes** — runs `git diff` to identify changed command source and skill files. Skips remaining steps if nothing relevant changed.
3. **Collects context** — gathers the command diff, skill diff, and full current skill content.
4. **Sends to Valet** — POSTs the diff + skill content to a Valet API session with a carefully crafted prompt that asks for a semantic ruling: does the skill still accurately describe the CLI after these changes?
5. **Posts a PR comment** — uses `github-script` to post (or update) a comment with the verdict (PASS/FAIL) and explanation.

## Credentials (placeholders)

All credentials are `secrets.*` references. Three secrets need configuring before this workflow runs:

| Secret | Purpose |
|---|---|
| `VALET_API_URL` | Base URL of the Valet API |
| `VALET_API_KEY` | Bearer token for authentication |
| `VALET_SESSION_ID` | Session ID for the conversation thread |

The Valet API call has `TODO` comments marking the request/response shape that needs adjusting once the Valet instance is available.

## Design decisions

- **Non-blocking.** The workflow is not a required check — it posts a comment but does not gate merge. Add to required checks once the team trusts the signal.
- **Self-contained.** No custom actions, no CLI installation — just `curl`, `jq`, and `github-script`.
- **Updates existing comment.** On re-push, the workflow finds and updates its previous comment rather than stacking duplicates.
- **No modifications to existing files.** This PR only adds the new workflow file.

## Related

- Discussion in #tvc-and-friends about skill staleness
- `tvc/AGENTS.md` — the rule this workflow enforces
- `tvc/src/commands/skills/save.rs` — the cargo test that enforces file-list parity

### DIFF
--- a/.github/workflows/skill-drift-check.yml
+++ b/.github/workflows/skill-drift-check.yml
@@ -1,0 +1,176 @@
+# Skill Drift Check — Valet-powered semantic ruling
+#
+# Detects when CLI command source changes without a matching skill update.
+# The cargo test `embedded_manifest_matches_skills_dir` enforces FILE-LIST
+# parity; this workflow checks SEMANTIC parity — whether the skill prose
+# still describes the CLI's actual behavior after a command changes.
+#
+# Required secrets (configure in Settings → Secrets and variables → Actions):
+#   VALET_API_URL      — Base URL of the Valet API (e.g. https://valet.example.com)
+#   VALET_API_KEY      — Bearer token for Valet API authentication
+#   VALET_SESSION_ID   — Session ID for the Valet conversation thread
+#
+# This workflow is intentionally non-blocking: it posts a comment but does
+# not gate merge. Add it to required checks only after the Valet integration
+# is stable and the team is comfortable with the signal.
+
+name: Skill Drift Check
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'tvc/src/commands/**'
+      - 'tvc/skills/**'
+
+jobs:
+  skill-drift-check:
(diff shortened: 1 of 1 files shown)


.github/workflows/skill-drift-check.yml
```diff
@@ -0,0 +1,176 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Detect changes
+        id: detect
+        run: |
+          CHANGED=$(git diff --name-only origin/main...HEAD -- tvc/src/commands/ tvc/skills/)
```
richardpringle: Isn't this already done by `paths` above?

PR #271, opened 2026-09-02
tvc: add instance size as deployment option

Turnalek:
Adds `instance_size_cpus` and `instance_size_ram` fields to the `deploy` config so users can request larger enclaves. We currently only support enumerated values of:
```
CPU: 2, RAM: 1Gi
CPU: 6, RAM: 24Gi
CPU: 14, RAM: 48Gi
```

Example deploy config:
```json
{
  "appId": "71606b3f-6bc0-43e4-b4c2-f831a1821b54",
  "qosVersion": "0.12.1",
  "pivotContainerImageUrl": "ghcr.io/anchorageoss/parser_app:latest@sha256:ff50383cc5bc5e80ef40430a6f58b80e511d2a50065b4b76699168b724eb3d9c",
  "pivotPath": "/parser_app",
  "pivotArgs": [],
  "expectedPivotDigest": "fb606155b55356a785a4b1350c1862516d07977b3a92512c9b1dffb96f395498",
  "dangerousDeployDebugMode": false,
  "healthCheckType": "TVC_HEALTH_CHECK_TYPE_HTTP",
  "healthCheckPort": 3000,
  "publicIngressPort": 3000,
  "instanceSizeCpus": 6,
  "instanceSizeRam": 24
}
```

Depends on https://github.com/tkhq/mono/pull/7966 but should be backwards compatible as the new intents values are all optional.

### DIFF
--- a/tvc/src/config/deploy.rs
+++ b/tvc/src/config/deploy.rs
@@ -48,6 +48,12 @@ pub struct DeployConfig {
     /// choice to the backend default.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub replicas: Option<u32>,
+    /// Enclave instance size CPU cound.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub instance_size_cpus: Option<u32>,
+    /// Enclave instance size RAM in Gi.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub instance_size_ram: Option<u32>,
 }
 
 /// Build a config seeded from an existing deployment, for use as a template.
@@ -92,6 +98,8 @@ impl TryFrom<TvcDeployment> for DeployConfig {
             health_check_type,
             health_check_port,
             public_ingress_port,
+            instance_size_cpus,
+            instance_size_ram,
         } = pivot_container
             .ok_or_else(|| anyhow!("deployment {id} has no pivot container spec"))?;
 
@@ -114,6 +122,8 @@ impl TryFrom<TvcDeployment> for DeployConfig {
             health_check_port,
             public_ingress_port,
             replicas: None,
+            instance_size_cpus: Some(instance_size_cpus),
+            instance_size_ram: Some(instance_size_ram),
         })
     }
 }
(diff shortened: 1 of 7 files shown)


tvc/src/config/deploy.rs
```diff
@@ -363,6 +374,8 @@ mod tests {
                 health_check_type: TvcHealthCheckType::Http,
                 health_check_port: 8080,
                 public_ingress_port: 9090,
+                instance_size_cpus: 2,
+                instance_size_ram: 8,
```
richardpringle, line 396: weren't we just going to do `small`, `medium`, `large`?

Turnalek: we are, this is just to not need to change the protocol in the future on the backend level, we will allow others later (or  basically any size)

Turnalek: the "enum" is enforced in code but it's always represented as cpus/ram size combo of 2 values with the exception of the observer column values  for front-end

Turnalek: Added validation on sdk side to ensure we don't try to send invalid combinations

richardpringle (dismissed):

tvc/src/config/deploy.rs
```diff
@@ -48,6 +48,12 @@ pub struct DeployConfig {
     /// choice to the backend default.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub replicas: Option<u32>,
+    /// Enclave instance size CPU cound.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub instance_size_cpus: Option<u32>,
+    /// Enclave instance size RAM in Gi.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub instance_size_ram: Option<u32>,
```
richardpringle, line 56: This is fine, but it's not the way I would do this. I would keep the actual api code the same, but I would do something like an instance type
```rust
#[derive(Default, Debug, Serialize, Deserialize)]
#[serde(rename_all = "kebab-case")]
enum InstanceType {
    #[default]
    Small,
    Medium,
    Large,
}
```
Then you don't need to do any validation client side and it only takes a cli change to open this up to more instance types (release the cli is not expensive and it's easy to get clients to upgrade - which we want anyway)

richardpringle (approved):

PR #272, opened 2026-09-03
fix(proofs): handle missing optional attestation fields

emostov:
Closes [ENG-4590](https://linear.app/turnkey/issue/ENG-4590)


daniilrrr (approved):
`.expect()` bad. As quick follow up could/should we add a clippy rule to ban `.expect()` throughout? cc @richardpringle

PR #273, opened 2026-09-05
feat: add unified tk CLI preview

figitaki:
# Add a unified tk CLI preview

Add a tk entry point alongside tvc, preserving existing TVC command behavior. Shared commands provide credential profiles and login/status/whoami, exact-body requests, activity inspection and recovery, user/policy/API-key management, wallets, and serialized signing. Local key generation writes an exclusive protected credential file and returns public material only.

This is an unreleased preview. TVC still uses its existing credential configuration; shared selectors are rejected there. Live lifecycle acceptance, browser onboarding, complete pagination, managed transaction delivery, import/export, and distribution remain follow-ups.

Validation: the implementation passed 460 CLI tests; the final mechanical helper move passed strict Clippy across all targets and a debug build/version smoke. The consolidated branch preserves that source exactly; only preview documentation was subsequently edited. No live organization acceptance was performed.

Review packet: https://reviews.figitaki.dev/r/ZG8muCcG

Integration status: main has advanced since this preview was validated. Merge conflicts in tvc/Cargo.toml and tvc/src/lib.rs must be resolved and checks rerun before this draft is ready to merge.

PR #274, opened 2026-09-09
feat: add encrypted Secrets transfers and resumable exports

figitaki:
Add encrypted Secrets import/export and metadata listing to the unified tk CLI. Import reads bytes from a file or stdin and encrypts them to an authenticated enclave target. Export writes to an explicit protected file and retains recipient recovery state across pending approvals, timeouts, and uncertain responses.

This PR is stacked on #273. It also extends the enclave encryption crate for arbitrary secret bytes with v1 signature and organization checks. Existing wallet/private-key bundle behavior is preserved.

The Secrets workflow does not print plaintext in ordinary command output. Policy-visible static properties are nonsecret metadata. Update/delete, provider-token rotation, and environment injection are outside this change. Import availability depends on the deployed API and organization.

Validation: 474 CLI tests and three additional process tests passed, with zero failures or ignored tests. Strict all-target Clippy passed. The crypto crate passed 29 unit tests and four doctests, including binary/empty payloads and adversarial bundles. Independent review approved the exact committed CLI head. Live organization acceptance remains pending; these results cover offline and mocked behavior.

Review packet: https://reviews.figitaki.dev/r/bVJIR66j (patchset 1).


tvc/src/shared_secrets.rs
```diff
@@ -0,0 +1,1414 @@
...
+    }
+}
+
+fn fingerprint(body: &str) -> String {
+    format!("sha256:{}", hex::encode(Sha256::digest(body.as_bytes())))
+}
+fn envelope(kind: &str, org: &str, timestamp: &str, parameters: Value) -> Value {
+    json!({"type":format!("ACTIVITY_TYPE_{kind}"),"timestampMs":timestamp,"organizationId":org,"parameters":parameters})
+}
+fn timestamp() -> Result<String> {
```
figitaki, line 295: surely not

figitaki:
Closing the combined CLI/crypto PR after splitting its responsibilities:

- CLI Secrets transfers and resumable exports: https://github.com/tkhq/tk/pull/26.
- Reusable SDK Secrets encryption helpers and tests: #279, extracted onto current main. The extraction passes 29 unit tests, four doctests, formatting, and strict Clippy.
- Updated CLI-backed Secrets skill: https://github.com/tkhq/turnkey-agent-skills/pull/24.

The replacement PRs remain under review. The tk port still needs to adopt the extracted SDK helpers once available, and live acceptance of that port remains separate from the original implementation's validation. Keeping this branch and review history for reference.

PR #278, opened 2026-09-09
Am/tvc 338/deploy approve clear not found error

am-hernandez:
### What

1. **`fix(tvc)`**: a typed `NotFoundInOrganization` error, built in the shared `fetch_tvc_deployment` helper when the backend returns 404. It keeps the client error as its source, so classification is still `not_found` with `httpStatus: 404` and the rendered chain keeps the server body. The message names the organization the request was scoped to, as `'alias' (id)` under a login profile or the bare ID under env auth, and the human-mode hint names the remedy for that auth source. `AuthenticatedClient` now records the login alias to make this possible, and the helper's redundant `organization_id` parameter is dropped. `deploy approve` and `deploy provision` route through the helper; `deploy get-status` and `deploy init --from-deployment` already did.
2. **`refactor(tvc)`**: `deploy status` uses the same helper, so the identical failure renders identically across subcommands. Tests updated for the new 404 wording.
3. **`docs(tvc-deployments)`**: the embedded skill documents the new message and remedies, and notes that `tvc app status --app-id` can recover the IDs of live deployments.

### Limitation and follow-ups

- The backend sends an empty `turnkeyErrorCode` on this 404, so the mapping keys off HTTP status alone. A route 404 from a bad `TVC_API_BASE_URL` gets the same message, and grpc-gateway uses the same `code: 5` body for both, so a body check would not help. The hint is worded conditionally for that reason. A specific backend code, as the version gate has with `TVC_CLIENT_VERSION_TOO_OLD`, would let the CLI match precisely.
- App and provisioning-details fetches still surface a bare 404 under wrong-org credentials. Same mapping applies; left for a scoped follow-up.
- [TVC-348](https://linear.app/turnkey/issue/TVC-348/get-app-status-returns-500-for-apps-with-no-live-deployment-instead-of): `get_app_status` returns 500 for apps with no live deployment. Found while verifying this change; root cause documented on the ticket.
- `fetch_tvc_deployment`'s signature changed. Nothing in the workspace depends on `tvc` as a library, but the function is technically public.

### Testing

- Unit tests cover classification, rendering with and without an alias, the hint per auth source, and that a version rejection riding a 404 still hints and classifies as a version rejection.
- Process-level tests run against a local server: the exact JSON object for a 404 and the `error:` plus `hint:` lines in human mode for `deploy approve`; the new wording for `deploy status`, with the human error line still matching the JSON message verbatim.
- `cargo clippy -p tvc --all-targets` clean, `cargo test -p tvc` green.
- Manually verified against prod with a login profile: the cross-org output above, and an unchanged happy path on an own-org deployment.

### Ticket 

Closes [TVC-338](https://linear.app/turnkey/issue/TVC-338/tvc-deploy-approve-clear-error-when-the-deployment-isnt-in-the-current)


tvc/src/errors.rs
```diff
@@ -37,6 +38,86 @@ impl MissingResource {
...
+                 switch with `{} login --org <alias-or-id>`",
+                binary_name()
+            ),
+            None => format!(
+                "check TVC_ORG_ID; if the {resource} belongs to a different organization, \
+                 set it to that organization's ID"
+            ),
+        }
+    }
+}
```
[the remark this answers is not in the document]
am-hernandez: CI is green and the crate compiles as written (cargo check -p tvc also passes locally), so this isn't a compile error. `Some(_)` is a wildcard pattern and doesn't bind the inner String, so nothing is moved out of the borrowed field. The crate compiles as written (`cargo check -p tvc` passes on this branch). Borrowing would only be required if the arm were `Some(alias)`. Leaving as is.

tvc/src/client.rs
```diff
@@ -29,6 +29,9 @@ pub struct AuthenticatedClient {
     pub client: TurnkeyClient<TurnkeyP256ApiKey>,
     /// The organization ID for API calls.
     pub org_id: String,
+    /// Alias of the saved login profile the credentials came from, for
+    /// user-facing messages. `None` when they came from environment variables.
+    pub org_alias: Option<String>,
```
richardpringle: I actually don't want to do this. I have a PR open that changes the way aliases work that will render this irrelevant

tvc/src/client.rs
```diff
@@ -79,20 +90,37 @@ pub async fn fetch_tvc_app(auth: &AuthenticatedClient, app_id: &str) -> Result<T
         .ok_or_else(|| MissingResource::new("app", app_id).into())
 }
 
+/// Fetch a deployment in the client's organization.
+///
+/// A 404 becomes a [`NotFoundInOrganization`] naming that organization: the
+/// usual cause is credentials scoped to a different organization than the one
+/// owning the deployment, and the bare status would not say so.
```
richardpringle, line 86: Just to be clear (and I'll see from the code in a second) we don't differentiate between the deployment not existing and it not being found in the org, right? 

We don't ever want to leak that a deployment might exist in another org (because we can't tell if the user is actually a part of that org or not). 

(this note is as much for myself as it is for you, I'll probably get my answer below)

tvc/src/client.rs
```diff
@@ -79,20 +90,37 @@ pub async fn fetch_tvc_app(auth: &AuthenticatedClient, app_id: &str) -> Result<T
...
-    let response = auth
-        .client
-        .get_tvc_deployment(GetTvcDeploymentRequest {
-            organization_id,
-            deployment_id: deployment_id.clone(),
-        })
-        .await
-        .with_context(|| format!("failed to fetch deployment {deployment_id}"))?;
+    let request = GetTvcDeploymentRequest {
+        organization_id: auth.org_id.clone(),
```
richardpringle, line 93: Nice improvement 🥇

tvc/src/client.rs
```diff
@@ -79,20 +90,37 @@ pub async fn fetch_tvc_app(auth: &AuthenticatedClient, app_id: &str) -> Result<T
...
+        deployment_id: deployment_id.clone(),
+    };
+
+    let response = match auth.client.get_tvc_deployment(request).await {
+        Ok(response) => response,
+        Err(source @ TurnkeyClientError::UnexpectedHttpStatus(404, _)) => {
+            return Err(NotFoundInOrganization::new(
+                "deployment",
+                deployment_id,
+                OrganizationLabel::new(auth.org_alias.clone(), auth.org_id.clone()),
```
richardpringle: This is a good idea, but will work a completely different way once I get my code landed. Do you mind backing this part out?

tvc/src/errors.rs
```diff
@@ -37,6 +38,86 @@ impl MissingResource {
...
+    id: String,
+    organization: OrganizationLabel,
+    #[source]
+    source: TurnkeyClientError,
+}
+
+impl NotFoundInOrganization {
+    pub fn new(
+        resource: &'static str,
+        id: impl Into<String>,
```
richardpringle, line 60: nice

tvc/src/errors.rs
```diff
@@ -339,6 +431,75 @@ mod tests {
...
+    }
+
+    fn not_found_in_organization_with_body(alias: Option<&str>, body: &str) -> anyhow::Error {
+        anyhow::Error::new(NotFoundInOrganization::new(
+            "deployment",
+            "abc-123",
+            OrganizationLabel::new(alias.map(str::to_string), "org-1".to_string()),
+            TurnkeyClientError::UnexpectedHttpStatus(404, body.to_string()),
+        ))
+    }
```
richardpringle: Haha, just an FYI, this is a bit of an anti pattern in Rust that the LLMs love pulling from JS/TS (where there are no ownership implications at function boundaries). It's fine here because it's in a test... just thought I would let you know

richardpringle (changes requested):
Just requesting that we back out the org-alias component and only because I'm making a big change to how that works.

tvc/src/client.rs
am-hernandez: > Just to be clear (and I'll see from the code in a second) we don't differentiate between the deployment not existing and it not being found in the org, right?

Exactly, we only know the backend answered 404 for that deployment ID in the logged-in org. We have no way to tell whether it exists under another org, and the CLI never does a second lookup to find out. The wrapper only adds the org ID the caller is already authenticated to, so nothing about other orgs show up in the message.

tvc/src/client.rs
am-hernandez: reversed this

tvc/src/client.rs
am-hernandez: done! Resolved here: 5dbca2b2bffc6e6697c1dde2f08e62075c6da974

tvc/src/client.rs
am-hernandez: resolved here: 5dbca2b2bffc6e6697c1dde2f08e62075c6da974

richardpringle (approved):

